PR #6, opened 2020-11-26
CLOUDP-77635: Atlas Project CRD

antonlisovenko:
This PR adds the fields to the `AtlasProject` struct.
Code and CRDs generation methods were called - so the CRDs were generated. 

* `make generate` - generated `zz_generated.deepcopy.go`
* `make manifests` - generated CRD files (`config/crd/bases`)


pkg/api/v1/atlasproject_types.go
```diff
@@ -20,27 +20,47 @@ import (
...
+
+	// ConnectionSecret is the name of the Kubernetes Secret which contains the information about the way to connect to
+	// Atlas (organization ID, API keys). The default Operator connection configuration will be used if not provided.
+	// +optional
+	ConnectionSecret *SecretRef `json:"connectionSecretRef,omitempty"`
+
+	// ProjectIPAccessList allows to enable the API Access List for the Project. See more information at
+	// https://docs.atlas.mongodb.com/configure-api-access/
+	// +optional
+	ProjectIPAccessList []ProjectIPAccessList `json:"ipAccessList,omitempty"`
```
gssbzn, line 50: [nit/q] for me `ProjectIPAccessList` has to do with https://docs.atlas.mongodb.com/reference/api/access-lists/ but your doc for this field makes me think this is for
https://docs.atlas.mongodb.com/reference/api/api-access-list/get-all-api-access-entries/
can you please clarify? if the field is for the project ip access list let's fix the link, if this is for API access list then could we rename the field to `APIAccessList`?

gssbzn: reading the the details of `ProjectIPAccessList` I think this is a case of fixing the link unless you really meant API access and then implemented project access

pkg/api/v1/atlasproject_types.go
```diff
@@ -51,14 +71,27 @@ type AtlasProject struct {
...
 // AtlasProjectList contains a list of AtlasProject
 type AtlasProjectList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []AtlasProject `json:"items"`
 }
 
-func init() {
-	SchemeBuilder.Register(&AtlasProject{}, &AtlasProjectList{})
+type ProjectIPAccessList struct {
```
gssbzn, line 81: see my other comment, this is ok if project access, not so if API access

gssbzn (commented):
I have a question to clarify `ProjectIPAccessList` as right now I'm not sure if it's really API access or project IP access

pkg/api/v1/atlasproject_types.go
antonlisovenko: yes, you are right, thought that we could omit `project` prefix considering that this configuration is for the `AtlasProject` entity but I agree that this introduces confusion. Will change 👍

gssbzn (approved):
LGTM thanks for improving the docs

leo-ri (approved):

PR #8, opened 2020-12-02
CLOUDP-77636: basic support for AtlasProject

antonlisovenko:
Very basic support for Atlas Project (will create the project in case it's not created).

Note, that due to Atlas library limitations it's now not possible to understand if the creation failed because the project already exists. This will be addressed in https://jira.mongodb.org/browse/CLOUDP-78002

So far this reconciliation will succeed only once - the second one will fail as the project already exists - this is expected and will be addressed after atlas library if fixed -see above

The logic is simplified to avoid getting a big PR, not tests so far


pkg/controller/atlas/connection.go
```diff
@@ -0,0 +1,55 @@
+package atlas
+
+import (
```
gssbzn, line 3: [q] any plans to `goimports` files?

pkg/controller/atlas/connection.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	if _, ok := secretData["orgId"]; !ok {
+		return Connection{}, fmt.Errorf("Missing field with key 'orgId' in the secret %v", secretRef)
+	}
+	if _, ok := secretData["publicApiKey"]; !ok {
+		return Connection{}, fmt.Errorf("Missing field with key 'publicApiKey' in the secret %v", secretRef)
+	}
+	if _, ok := secretData["privateApiKey"]; !ok {
+		return Connection{}, fmt.Errorf("Missing field with key 'privateApiKey' in the secret %v", secretRef)
+	}
```
gssbzn, line 49: [q] given you are wrapping `secretRef` could this error leak secrets? ie orgID is not set but api keys are and we end up logging this error somewhere?

gssbzn: also optional/future improvement, I find tools that error on a missing field just to error again on another one, it would be nice to check all required fields and returning once with all the missing info instead of me having to run this again to find a new piece of missing information

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -18,28 +18,58 @@ package atlasproject
...
+
+	connection, err := atlas.ReadConnection(r.Client, "TODO!", project.ConnectionSecretObjectKey(), log)
+	if err != nil {
+		log.Error(err, "Failed to read Atlas Connection details")
+		return reconcile.Result{RequeueAfter: time.Second * 10}, nil
+	}
+
+	if err := ensureProjectExists(connection, project, log); err != nil {
+		log.Error(err, "Failed to read the AtlasProject")
+		return reconcile.Result{RequeueAfter: time.Second * 10}, nil
```
gssbzn, line 69: [nit] I usually find to define a const like `const refreshRate = 10 * time.Second` and then using that const around to be more readable/maintainable, also given how this is used multiple time in this file could make a refactor of the time easier later on

pkg/controller/atlasproject/project.go
```diff
@@ -0,0 +1,44 @@
...
+
+func ensureProjectExists(connection atlas.Connection, project *mdbv1.AtlasProject, log *zap.SugaredLogger) error {
+	// TODO transport + http client + configurable atlas backend!
+	t := digest.NewTransport(connection.PublicKey, connection.PrivateKey)
+	tc, err := t.Client()
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+
+	client, err := mongodbatlas.New(tc, mongodbatlas.SetBaseURL("https://cloud-qa.mongodb.com/api/atlas/v1.0/"))
```
gssbzn, line 20: [q] `// TODO change?` could this be set via env right now in case someone wants to test with dev or prod?

pkg/controller/atlas/connection.go
```diff
@@ -0,0 +1,55 @@
...
+// former is not specified
+func ReadConnection(kubeClient client.Client, operatorName string, projectOverrideSecretRef *client.ObjectKey, log *zap.SugaredLogger) (Connection, error) {
+	if projectOverrideSecretRef != nil {
+		// TODO is it possible that part of connection (like orgID is still in the Operator level secret and needs to get merged?)
+		log.Infof("Reading Atlas API credentials from the AtlasProject Secret %s", projectOverrideSecretRef)
+		return readAtlasConnectionFromSecret(kubeClient, *projectOverrideSecretRef)
+	}
+	// TODO check the default "Operator level" Secret
+	// return readAtlasConnectionFromSecret(operatorName + "-connection")
+	return Connection{}, errors.New("The API keys are not configured")
```
gssbzn, line 29: [not blocking] in general go doesn't recommend to start errors with CAPS as they may be logged and the capitalisation ends up being weird in the middle of a sentence, just flagging this out as I'm seeing it in a lot of places

gssbzn (commented):
I have a question about the potential to leak secrets, the rest is totally optional

pkg/controller/atlas/connection.go
antonlisovenko: yep, we have a ticket to create a gitcommit hook, but will configure Idea to do this as well, thanks for pointing

pkg/controller/atlas/connection.go
antonlisovenko: re 1: don't think we'll leak anything as only the secret name is printed, not its content
re 2: I agree with you, there is some duplication here as well, can be done by using an array and throwing the error once 👍

pkg/controller/atlasproject/atlasproject_controller.go
antonlisovenko: totally agree! Planned to address this in the next PRs, this one has many TODOs about consolidating functionality. Here, for example some other things should happen: update the status of the resource, log, return - this should (and will) be handled the more generic way 👍

pkg/controller/atlasproject/project.go
antonlisovenko: it will be! Just wanted to make this running and avoid a huge PR. 
Operator configuration will be done a proper way later

pkg/controller/atlas/connection.go
antonlisovenko: yes, always forget about that!

gssbzn (approved):
If we don't leak then LGTM we can address the rest later

leo-ri (approved):

PR #9, opened 2020-12-03
Some quickstart instructions and small fixes

antonlisovenko:
Played with the ways to deploy operator locally and found the combination of Kind + Docker local registry the best option.
K3s requires Linux so the VM needs to be installed to make it work on MacOS (I assume Windows will also have problems).

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,7 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: localhost:5000/mongodb-atlas-kubernetes-operator
-  newTag: latest
-- name: mongodb-atlas-kubernetes-operator
-  newName: localhost:5000/mongodb-atlas-kubernetes-operator
   newTag: main-311bf0e
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,2 +1,11 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: localhost:5000/mongodb-atlas-kubernetes-operator
+  newTag: latest
+- name: mongodb-atlas-kubernetes-operator
+  newName: localhost:5000/mongodb-atlas-kubernetes-operator
+  newTag: main-311bf0e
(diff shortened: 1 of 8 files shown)


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
```
antonlisovenko, line 3: this was generated when I called `make deploy`, tbh so far not sure I understand what to do with this...

docs/dev/start-guide.md
```diff
@@ -0,0 +1,14 @@
...
+
+1. Install Go (1.15)
+1. Install act (`brew install act` on Mac or check [instructions](https://github.com/nektos/act#installation))
+1. Install Kind ([instructions](https://kind.sigs.k8s.io/docs/user/quick-start/#installation))
+1. Clone the project to your workspace (note, that this doesn't need to be `GOPATH` as the project uses Go Modules)
+1. Copy the default Github Actions settings for local run: `cp .actrc.local.sample .actrc`
+1. Build and push Operator image to the local registry: `act -j build-push`
+1. Ensure local docker registry started and create Kind cluster (with support for local registry): use the following [script](https://kind.sigs.k8s.io/docs/user/local-registry/)
+1. Install CRDs into the K8s cluster: `make install`
+1. Deploy Operator into the K8s cluster: `make deploy IMG=localhost:5000/mongodb-atlas-kubernetes-operator:main-311bf0e` (TODO we need to use 'latest' for local development)
```
antonlisovenko, line 11: fyi @leo-ri for local development `latest` tag would be the most convenient. Seems like some new config (gh secret?) may be created with `local=true` or whatever?

antonlisovenko: yes, this will work, agree 👍

config/manager/kustomization.yaml
leo-ri: should  we commit it? it will be generated every time

leo-ri (approved):

config/manager/kustomization.yaml
antonlisovenko: we should definitely address this and not commit...
Let's so far just overwrite it each time instead of wiping before each commit and then some time later make it work better. 
Kustomize is still a bit of learning debt to me :(

PR #10, opened 2020-12-04
Fix api groups

antonlisovenko:
I messed up a bit with generation and provided the wrong `group` so the wrong API groups were generated.
Also some more annotations were required to fix generation using `make manifests`


pkg/api/v1/atlasproject_types.go
```diff
@@ -63,6 +63,8 @@ type AtlasProjectStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:printcolumn:name="Name",type=string,JSONPath=`.spec.name`
 // +kubebuilder:subresource:status
+// +groupName:=atlas.mongodb.com
```
antonlisovenko, line 66: this one is the critical annotation to generate proper `spec.group` in the CRDs and also the CRD files with correct names, otherwise `config/crd/bases/mongodb.com.mongodb.com_atlasprojects.yaml` were generated...

vasilevp: q: so without this one, does it just assume "mongodb.com" because it's some sort of a top-level group name?

vasilevp (approved):

pkg/api/v1/atlasproject_types.go
antonlisovenko: that's what I have no clue about... magically this annotation is not needed if the layout of the file is the same that was provided by OperatorSDK (e.g. root `apis` and `controllers` instead of grouped in `pkg` folder) and the files and `spec.group` are generated fine in this case.

PR #11, opened 2020-12-07
Common Atlas client

antonlisovenko:
This PR moves the code to initialize Atlas Client connection to a common place that allows reuse.
The logging functionality is added on top that allows to debug HTTP requests.

This is the output:

```
2020-12-07T18:50:38.651Z	INFO	controllers.AtlasProject	-> Starting AtlasProject reconciliation	{"atlasproject": "mongodb-atlas-kubernetes-system/my-project", "spec": {"name":"Test Atlas Operator Project","connectionSecretRef":{"name":"my-atlas-key"}}}
2020-12-07T18:50:38.651Z	INFO	controllers.AtlasProject	Reading Atlas API credentials from the AtlasProject Secret mongodb-atlas-kubernetes-system/my-atlas-key	{"atlasproject": "mongodb-atlas-kubernetes-system/my-project"}
2020-12-07T18:50:43.698Z	DEBUG	controllers.AtlasProject	HTTP Request (POST) https://cloud-qa.mongodb.com/api/atlas/v1.0/groups [time (ms): 5046, status: 201]	{"atlasproject": "mongodb-atlas-kubernetes-system/my-project"}
2020-12-07T18:50:43.698Z	DEBUG	controller	Successfully Reconciled	{"reconcilerGroup": "atlas.mongodb.com", "reconcilerKind": "AtlasProject", "controller": "atlasproject", "name": "my-project", "namespace": "mongodb-atlas-kubernetes-system"}
```


pkg/util/httputil/loggedclient.go
```diff
@@ -0,0 +1,42 @@
...
+		CheckRedirect: client.CheckRedirect,
+		Jar:           client.Jar,
+		Timeout:       client.Timeout,
+	}
+}
+
+type loggedRoundTripper struct {
+	rt  http.RoundTripper
+	log *zap.SugaredLogger
+}
```
gssbzn, line 23: [nit] I think this may be a bit wasteful, `digest` already gave you a client and you're now copying it over to a new client, I would probably embed `digest.Transport` here
like
```go
type LoggingTransport struct {
	digest.Transport
	log *zap.SugaredLogger
}

func NewLoggingTransport(username, password string, log *zap.SugaredLogger) *LoggingTransport {
	return &LoggingTransport{
		Transport:     digest.Transport{
			Username: username,
			Password: password,
			Transport: http.DefaultTransport,
		},
		log: log,
	}
}

func (t *LoggingTransport) RoundTrip(request *http.Request) (*http.Response, error) {
	startTime := time.Now()
	response, err := t.Transport.RoundTrip(request)
	duration := time.Since(startTime)
	t.logResponse(request, response, err, duration)
	return response, err
}

// to be used like

t := httputil.NewLoggingTransport(connection.PublicKey, connection.PrivateKey, log)
tc, err := t.Client()
if err != nil {
	return nil, err
}
mongodbatlas.New(tc, mongodbatlas.SetBaseURL("https://cloud-qa.mongodb.com/api/atlas/v1.0/"))
```

I would even consider making `NewLoggingTransport(username, password string, log *zap.SugaredLogger)` receive options in a functional way but this can de done later. My point here is that it feels like a lot of passing around a client (copy way) when just one should be enough 

(some options in my example could be tuned based on pointers or not, I don't have a strong opinion on the details of my example just the point of reusing transport instead of creating a new client)

Edit: reusing `t.Client()` may not work

pkg/controller/atlas/client.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/mongodb-forks/digest"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/httputil"
+	"go.mongodb.org/atlas/mongodbatlas"
+	"go.uber.org/zap"
+)
+
+// AtlasClient is the central place to create a client for Atlas using specified API keys and a server URL.
+// Note, that the default HTTP transport is reused globally by Go so all caching, keep-alive etc will be in action.
+func AtlasClient(connection Connection, log *zap.SugaredLogger) (*mongodbatlas.Client, error) {
```
gssbzn, line 12: [nit] given the name of the package is `atlas`, `atlas.AtlasClient` is a bit redundant, please consider something like `atlas.NewClient`

antonlisovenko: Yes!

pkg/util/httputil/loggedclient.go
antonlisovenko: thanks for the extended review!
I see your point and I think it has good reasons and explanation.
The reasons why I like docorating pattern more here is that this LoggingClient can optionally "wrap" another HTTP client and not couple with other functionality (digest in this case). Initially I even wanted to make "digest client" a decorator over the HTTP client as well but `digest` library is built a bit different way so I decided to make it a "starting point" of creating an HTTP client (this is ok as the default go Http client keeps the global cache etc state that I wasn't aware of). So now the digest client is initialized in the `atlas/client.go` and not in this `util` package which contains only the logging wrapper.
As to copying the structs - I believe it's a very [cheap operation](https://go101.org/article/value-copy-cost.html#:~:text=The%20costs%20of%20copying%20small,are%20all%20small%2Dsize%20types.&text=The%20costs%20of%20copying%20values,base%20types%20are%20very%20small.) and afaik is even cheaper than working with pointers

gssbzn: on the subject of pointers vs value, it's cheap for basic types, for advanced structs it depends of the struct as each property adds up to the cost so big structs can fall in the not cheap side. This can also be seeing in your link where they benchmark copy vs pointers, copy is twice as slow as pointers for a struct*
```
Benchmark_Range_OneIterVar-4    500000   3203 ns/op
Benchmark_Range_TwoIterVars-4   200000   6616 ns/op
```
* they mention this and I agree, it will always depend on multiple things

Also understanding the flow of digest.Client to LoggedClient requires some mental exercise here because you are still decorating transport even if you don't want to tightly couple it with `digest.Transport`, `RoundTrip` is for Transport and not client.

I agree with you if you don't want to tightly couple this with `digest` but to avoid this another option can be 

```go
type LoggingTransport struct {
	http.RoundTripper
	log *zap.SugaredLogger
}

func NewLoggingClient(t http.RoundTripper, log *zap.SugaredLogger) *http.Client {
	return &http.Client{
		Transport: &LoggingTransport{RoundTripper: t, log: log},
	}
}

func (t *LoggingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	startTime := time.Now()
	res, err := t.RoundTripper.RoundTrip(req)
	duration := time.Since(startTime)
	t.logResponse(request, response, err, duration)
	return res, err
}
```

Again the point here is to avoid so many http.Client going around (no strong opinion here if you want `RoundTripper` embedded or as a prop)

antonlisovenko: As discussed p2p the functional approach could be an alternative to flexible configuration of the client.

go.sum
```diff
@@ -307,6 +307,8 @@ go.etcd.io/bbolt v1.3.3/go.mod h1:IbVyRI1SCnLcuJnV2u8VeU0CEYM7e686BmAb1XKL+uU=
 go.etcd.io/etcd v0.0.0-20191023171146-3cf2f69b5738/go.mod h1:dnLIgRNXwCJa5e+c6mIZCrds/GIG4ncV9HhK5PX7jPg=
 go.mongodb.org/atlas v0.5.0 h1:WoeXdXSLCyquhSqSqTa0PwpjxWZbm70E1Gkbx+w3Sco=
 go.mongodb.org/atlas v0.5.0/go.mod h1:CIaBeO8GLHhtYLw7xSSXsw7N90Z4MFY87Oy9qcPyuEs=
+go.mongodb.org/atlas v0.5.1-0.20201208094933-0e2a93147ccd h1:522hIBk7NjCNro3wxslqPjnMUlg/A1JZVHTWudcOC50=
+go.mongodb.org/atlas v0.5.1-0.20201208094933-0e2a93147ccd/go.mod h1:CIaBeO8GLHhtYLw7xSSXsw7N90Z4MFY87Oy9qcPyuEs=
```
gssbzn, line 311: you may want to do `go mod tidy` to clean duplicates and leave just the ones you're using

pkg/util/httputil/decoratedclient_test.go
```diff
@@ -0,0 +1,24 @@
...
+func Test_DecorateClient(t *testing.T) {
+	httpClient := &http.Client{Transport: http.DefaultTransport}
+	withDigest := Digest("publiApi", "privateApi")
+	withLogging := LoggingTransport(zap.S())
+
+	decorated, err := DecorateClient(&http.Client{Transport: http.DefaultTransport}, withDigest, withLogging)
+	assert.NoError(t, err)
+	assert.Equal(t, httpClient.Timeout, decorated.Timeout)
+	assert.Equal(t, httpClient.Jar, decorated.Jar)
+	assert.NotNil(t, decorated.Transport)
```
gssbzn, line 20: [nit] when asserting multiple times `testify` recommends to use the following pattern 

```suggestion
	a := assert.New(t) // using "a" as example but the name is not important 
	a.NoError(err)
	a.Equal(httpClient.Timeout, decorated.Timeout)
	a.Equal(httpClient.Jar, decorated.Jar)
	a.NotNil(decorated.Transport)
```

gssbzn (approved):
Thanks for all the changes, I left a couple of recommendations but nothing blocking this here

vasilevp (approved):

leo-ri (approved):

PR #15, opened 2020-12-14
CLOUDP-78895: basic support for status

antonlisovenko:
This is part 1 of the basic support for statuses.
This adds the basic status (with `phase`) shared by all the entities. Also adds the `id` fields into the `AtlasProject` CR.
Also adds the functionality to perform a patch of the status.

Some thoughts behind this design: as discussed with the team we'll keep to the [recommendations](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) by K8s to use `conditions` to represent detailed information about the state of the entity but we'll also use a "shortcut" `phase` showing the general state of the object to the user. It's the same for the `Pod` that has both.

Conditions will be addressed later.

This is not glued into the controller code to avoid the PR growing bigger - will be addressed in the next PR


pkg/controller/status/patch.go
```diff
@@ -0,0 +1,39 @@
...
+		Value: statusValue,
+	}}
+
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return err
+	}
+
+	patch := client.RawPatch(types.JSONPatchType, data)
+	return kubeClient.Status().Patch(context.Background(), resource, patch)
```
vasilevp, line 38: If this context will be passed in the future, we should either expose it right away or use `context.TODO()`

pkg/api/v1/atlascustomresource.go
```diff
@@ -0,0 +1,13 @@
+package v1
+
+import (
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/status"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+//+k8s:deepcopy-gen=false
+// AtlasCustomResource is the interface common for all Atlas entities
+type AtlasCustomResource interface {
```
vasilevp, line 10: Should we add static interface checks to make sure our types implement this interface?

pkg/api/v1/atlasproject_types.go
```diff
@@ -54,12 +55,14 @@ type AtlasProjectSpec struct {
 
 // AtlasProjectStatus defines the observed state of AtlasProject
 type AtlasProjectStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	status.Common `json:",inline"`
```
gssbzn, line 58: [nit] `,inline` is not "officially" supported in json tags and there are still discussions going around it

pkg/controller/status/patch.go
```diff
@@ -0,0 +1,39 @@
...
+
+type patchValue struct {
+	Op    string      `json:"op"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value"`
+}
+
+// PatchUpdateStatus performs the JSONPatch patch update to the Atlas Custom Resource.
+// The "jsonPatch" merge allows to update only status field so is more
+func PatchUpdateStatus(kubeClient client.Client, resource mdbv1.AtlasCustomResource) error {
```
gssbzn, line 21: [non-blocking] I recently found that defining interfaces in the same package in which they are used (even in the same file where they may be needed) and only with the methods I need to be a great way for SOLID, this may lead to have a lot of smaller interfaces around but I'm ok with the tradeoff than having a single big interface

pkg/controller/status/patch.go
antonlisovenko: I thought about it and I wasn't sure if we need proper context support.
To my mind the context shines when there is a customer-facing system and we need to put time thresholds, responsiveness adjusting technics (so have possibility to cancel contexts and the whole pipe of requests). In this case the reconciliation is done in background fashion so it's ok if the request to Atlas has only HTTP timeouts and no extra orchestration.

Happy to discuss more

pkg/api/v1/atlascustomresource.go
antonlisovenko: do you mean smth like 
```
var _ AtlasCustomResource = AtlasProject
```

pkg/api/v1/atlasproject_types.go
antonlisovenko: this is a requirement of one of the code generating libs we use - it requires all the elements to have a `json` annotation

pkg/controller/status/patch.go
antonlisovenko: sure, do you see anything that could be optimized here? Is it the `AtlasCustomResource`?

pkg/api/v1/atlasproject_types.go
gssbzn: Oh, interesting then nothing to do here

pkg/controller/status/patch.go
gssbzn: yes, I'm wondering if it would make sense to move it from `pkg/api/v1/atlascustomresource.go` to this same file but up to you, also if you follow vasilevp comment on static checking it may be better to leave it where it is to avoid a circular dependency

pkg/controller/status/patch.go
gssbzn: I'm with anton in this one that if there's no need to cascade request and cancellation background is a good option

gssbzn (approved):
LGTM

pkg/controller/status/patch.go
antonlisovenko: yes, this is the central interface residing in the `apis` package which doesn't depend on anything else. But `controller` package depends on `apis` so moving it elsewhere will turn into circular dependency.

But thanks for the suggestion, I agree that having "local" interfaces looks like a common Go pattern 👍

pkg/api/v1/atlascustomresource.go
vasilevp: Yes, pretty much

pkg/controller/status/patch.go
vasilevp: My only concern here is hanging requests due to lack of timeouts. Not an issue if the client being used has reasonable default timeouts

vasilevp (approved):

leo-ri (approved):

PR #16, opened 2020-12-16
CLOUDP-78984: workflow support

antonlisovenko:
This PR enables the generic abstractions to support proper status update for any Atlas Custom Resource.

As discussed with the team we'll stop on using `status.conditions` as the only way to track the state of the CR (as recommended by https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) and not using `status.phase`. This makes handling the workflow a bit harder as we need to keep both the "current reconciliation status" updating it on the run and also stop reconciliations due to error/waiting for resources.

For the first thing I've introduced `workflow.Context` which is supposed to be passed to each method.
For the second purpose - `workflow.Result` which is the way to affect the reconciliation processing and signal that it needs to be retried. 

I've shown this to the team, so @gssbzn - your review is optional and I don't want to distract you much! Can explain more individually if necessary.

No tests have been added to avoid the PR growing big.
Some outputs for different exceptional situations:

(the secret with Atlas credentials hasn't been provided)
```
status:
  conditions:
  - lastTransitionTime: "2020-12-16T15:52:42Z"
    message: Secret "my-atlas-key" not found
    reason: AtlasCredentialsNotProvided
    status: "False"
    type: ProjectReady
```

(the project already exists - this will be fixed soon PR)
```
  conditions:
  - lastTransitionTime: "2020-12-16T16:02:54Z"
    message: 'POST https://cloud-qa.mongodb.com/api/atlas/v1.0/groups: 409 (request
      "GROUP_ALREADY_EXISTS") A group with name "Test AO project (Anton)" already
      exists.'
    reason: ProjectNotCreatedInAtlas
    status: "False"
    type: ProjectReady
```

Operator logs:
```
2020-12-16T16:06:53.348Z	INFO	controllers.AtlasProject	-> Starting AtlasProject reconciliation	{"atlasproject": "default/my-project", "spec": {"name":"Test AO project (Anton)","connectionSecretRef":{"name":"my-atlas-key"}}}
2020-12-16T16:06:53.348Z	INFO	controllers.AtlasProject	Reading Atlas API credentials from the AtlasProject Secret default/my-atlas-key	{"atlasproject": "default/my-project"}
2020-12-16T16:06:53.664Z	DEBUG	controllers.AtlasProject	HTTP Request (POST) https://cloud-qa.mongodb.com/api/atlas/v1.0/groups [time (ms): 315, status: 409]	{"atlasproject": "default/my-project"}
2020-12-16T16:06:53.664Z	INFO	controllers.AtlasProject	Status update	{"atlasproject": "default/my-project", "lastCondition": {"type":"ProjectReady","status":"False","lastTransitionTime":null,"reason":"ProjectNotCreatedInAtlas","message":"POST https://cloud-qa.mongodb.com/api/atlas/v1.0/groups: 409 (request \"GROUP_ALREADY_EXISTS\") A group with name \"Test AO project (Anton)\" already exists."}}
```


pkg/controller/atlasproject/project.go
```diff
@@ -31,7 +33,9 @@ func ensureProjectExists(connection atlas.Connection, project *mdbv1.AtlasProjec
 				}
 
 			}*/
-		return err
+		return "", workflow.Terminate(workflow.ProjectNotCreatedInAtlas, err.Error())
 	}
-	return nil
+	ctx.EnsureStatusOption(status.NewIDOption(p.ID))
```
antonlisovenko, line 38: this is an example of how the "internal" method can update the "ongoing" status. Here the method is responsible for ensuring the project ID is added to the status in the end

gssbzn (approved):
This is probably to k8s specific and out of my go experience so far. LGTM if the rest of the team agrees

leo-ri (approved):

pkg/api/v1/status/option.go
```diff
@@ -0,0 +1,32 @@
...
+import "reflect"
+
+// Option is the generic container for any information that needs to be put into the status of the Custom Resource.
+// This is the way to handle some random data that need to be written to status
+type Option interface {
+	Value() interface{}
+}
+
+// GetOption finds the option by its type
+func GetOption(statusOptions []Option, targetOption Option) (Option, bool) {
```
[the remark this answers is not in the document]
antonlisovenko: yep, I like the idea of using functions instead of data containers 👍 
I think I liked the idea that the "target" (`AtlasProject` etc) has control over what parts of status are updated in the `updateStatus` instead of delegating to some caller but on the other hand there will still be restriction as only the callbacks working with `atlasProject.Status` will be called and not the others.

Don't see the "cons" you've mentioned - we still can move the "specific" options to subpackages, I think generation doesn't work only if the CRD structs are moved to subpackages, it's ok to move some "helper" code there.

pkg/controller/workflow/status.go
```diff
@@ -0,0 +1,43 @@
...
+			return
+		}
+	}
+	// Condition not found - appending
+	s.conditions = append(s.conditions, condition)
+}
+
+func (s *Status) EnsureOption(option status.Option) {
+	for i, c := range s.options {
+		if reflect.TypeOf(c) == reflect.TypeOf(option) {
```
vasilevp, line 36: This will not work correctly with the new approach - it will be true for all `AtlasProjectStatusOption`s.

I suggest we scrap this check and just append every time - the last option will take precedence anyway.

vasilevp (changes requested):
`EnsureOption` needs to take into account the new functor types

pkg/controller/workflow/status.go
antonlisovenko: good catch.
Yes, so far it looks quite safe as the last option will override what was done by all previous ones (unless there's some conditional logic inside the function so the changes are done to some different parts).

If we get to this problem we may turn the model to maps instead or make Option a struct with name and func fields (so we'll be able to map by name)

vasilevp (approved):

PR #17, opened 2020-12-16
CLOUDP-77591: Add Cluster Spec definitions

vasilevp:
This PR implements the full Atlas Cluster Spec with the help of Atlas docs & `mongodbatlas` client library.


pkg/api/v1/atlascluster_types.go
```diff
@@ -17,23 +17,231 @@ limitations under the License.
...
+	// +optional
+	ClusterType string `json:"clusterType,omitempty"`
+
+	// Capacity, in gigabytes, of the host's root volume.
+	// Increase this number to add capacity, up to a maximum possible value of 4096 (i.e., 4 TB).
+	// This value must be a positive integer.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=4096
+	// +optional
+	DiskSizeGB *int `json:"diskSizeGB,omitempty"` // TODO: why on earth is this *float64 in mongodbatlas?
```
gssbzn, line 66: See: https://github.com/mongodb/go-client-mongodb-atlas/issues/140
Also can we please change this TODO to link to that issue

pkg/api/v1/atlascluster_types.go
```diff
@@ -17,23 +17,231 @@ limitations under the License.
...
+	// Atlas (organization ID, API keys). The default Operator connection configuration will be used if not provided.
+	// +optional
+	ConnectionSecret *SecretRef `json:"connectionSecretRef,omitempty"`
+
+	// Collection of settings that configures auto-scaling information for the cluster.
+	// If you specify the autoScaling object, you must also specify the providerSettings.autoScaling object.
+	// +optional
+	AutoScaling *AutoScalingSpec `json:"autoScaling,omitempty"`
+
+	// Deprecated: do not use.
```
gssbzn, line 44: FYI, if you try to use it your changes will be rejected, it's kept as users who had it enabled can still use it

leo-ri (approved):

gssbzn (approved):
LGTM just one FYI

pkg/api/v1/atlascluster_types.go
vasilevp: Forgot to delete this field - fixing

pkg/api/v1/atlascluster_types.go
```diff
@@ -17,23 +17,231 @@ limitations under the License.
...
+	// Number of read-only nodes for Atlas to deploy to the region.
+	// Read-only nodes can never become the primary, but can facilitate local-reads.
+	// +optional
+	ReadOnlyNodes *int64 `json:"readOnlyNodes,omitempty"`
+}
+
+// Cluster converts the Spec to native Atlas client format.
+func (spec *AtlasClusterSpec) Cluster() *mongodbatlas.Cluster {
+	result := &mongodbatlas.Cluster{}
+	err := copier.Copy(result, spec)
```
antonlisovenko, line 230: [q] why not json serialize -> deserialize?
I'm afraid that that library will just silently omit the non-matching fields though I believe our contract is that all `AtlasCluster` fields must match the ones from the atlas-go client?

pkg/api/v1/atlascluster_types.go
```diff
@@ -17,23 +17,231 @@ limitations under the License.
...
+	if err != nil {
+		panic(err)
```
antonlisovenko, line 232: we need to handle these kinds of exceptions

antonlisovenko (changes requested):
Please remove the `connectionSecret` - it's provided by the project.

Apart of that only some cosmetic comments

pkg/api/v1/atlascluster_types.go
```diff
@@ -17,23 +17,231 @@ limitations under the License.
...
+	// +optional
+	EncryptionAtRestProvider string `json:"encryptionAtRestProvider,omitempty"`
+
+	// Collection of key-value pairs that tag and categorize the cluster.
+	// Each key and value has a maximum length of 255 characters.
+	// +optional
+	Labels []LabelSpec `json:"labels,omitempty"`
+
+	// Version of the cluster to deploy.
+	// +kubebuilder:validation:Enum="3.6";"4.0";"4.2";"4.4"
```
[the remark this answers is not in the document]
gssbzn: *4.9, and then 5.0

pkg/api/v1/atlascluster_types.go
vasilevp: this panic would mean we have incompatibilities in our types, which is not something the user can fix without altering the code - so how should we handle this?

vasilevp: besides, from my observations, we don't handle similar panics in other PRs/existing code - should we?

pkg/api/v1/atlascluster_types.go
```diff
@@ -17,23 +17,231 @@ limitations under the License.
...
+}
 
 // AtlasClusterSpec defines the desired state of AtlasCluster
 type AtlasClusterSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// ConnectionSecret is the name of the Kubernetes Secret which contains the information about the way to connect to
+	// Atlas (organization ID, API keys). The default Operator connection configuration will be used if not provided.
+	// +optional
+	ConnectionSecret *SecretRef `json:"connectionSecretRef,omitempty"`
```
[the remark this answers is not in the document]
vasilevp: If that's the case - what's the best approach to get that secret from `Project` in controller?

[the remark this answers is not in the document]
antonlisovenko: this should be the link to the project: https://docs.google.com/document/d/1k2JVqRMz5hsfYvScXn8GY0cO7Ux9f_7FoKHFzSlQiVA/edit#bookmark=id.vrkw2z6hazov
(the same format as for `SecretRef`)

pkg/api/v1/atlascluster_types.go
vasilevp: You're probably right, but using JSON still won't protect us from silent omissions if our tags are different for some reason - which should not happen barring a typing mistake.

We might actually have to write a unit test specifically for this - that will also give us a helpful future-proof tool to catch new fields being added to the Atlas client.

pkg/api/v1/atlascluster_types.go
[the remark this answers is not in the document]
vasilevp: can we reuse it as `ResourceRef`, for example?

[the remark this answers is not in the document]
antonlisovenko: you know... I think you are right. I know in Enterprise we used `SecretRef` that has an optional field `key` to reference a key in the secret. Though not the current case so `ResourceRef` looks good 👍

pkg/api/v1/atlascluster_types.go
antonlisovenko: I agree, having the unit test is the best. I think it could work in both directions: finding the fields from K8s that are not mapped to the Atlas-lib and the other direction (with some explicit exclusions for the fields we don't expose)
could you file the ticket please?

pkg/api/v1/atlascluster_types.go
antonlisovenko: One thing: in case of panic the whole operator falls and will fall on retry as well (as the event will come again) so it will get stuck. 
Catching this bug will isolate the problem for one single resource and will result in a proper status message that the customer will read and maybe will manage to fix the problem by removing some fields.
I agree that it's possible to panic on some internal development logic that doesn't depend on user input - so the tests would catch it. But here the `cluster` is the user-provided spec and they may change it to fix the problem

vasilevp: I agree we should be able to gracefully handle incorrect input. This logic, however, does _not_ handle any user input, as it only converts between two different statically-typed `struct`s, and it is our job to make sure they are correct and convertible.

Even after I change this to JSON S/D, any error during (de)serialization will mean something is broken in our code beyond repair, and I think we should fail as loud as possible whenever that's the case - hence the panic.

Am I missing something here?

pkg/api/v1/atlascluster_types.go
vasilevp: Created [CLOUDP-79276](https://jira.mongodb.org/browse/CLOUDP-79276)

pkg/api/v1/atlascluster_types.go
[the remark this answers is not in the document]
vasilevp: Created [CLOUDP-79277](https://jira.mongodb.org/browse/CLOUDP-79277)

pkg/api/v1/atlascluster_types.go
antonlisovenko: > This logic, however, does not handle any user input
Not sure I agree with this as the logic deals with the AtlasCluster `spec` which is provided by the user. The Spec is taken and transformed into the atlas-client struct

pkg/api/v1/common.go
```diff
@@ -8,8 +8,14 @@ not use this file except in compliance with the License. You may obtain
 a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
 */
 
-// SecretRef is a reference to a Kubernetes Secret
-type SecretRef struct {
+// ResourceRef is a reference to a Kubernetes Secret
```
antonlisovenko, line 11: [nit] comment

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.15
 require (
 	github.com/go-logr/logr v0.1.0
 	github.com/go-logr/zapr v0.1.0
+	github.com/jinzhu/copier v0.1.0
```
antonlisovenko, line 8: I think this library is not needed so far?

antonlisovenko (approved):
Please run `make manifests` as it seems the CRD was not updated.
LGTM after `panic` handling is replaced by errors 👍

pkg/api/v1/atlascluster_types.go
vasilevp: As discussed 1:1, the panics were removed

PR #18, opened 2020-12-16
Add golangci-lint settings

vasilevp:
This PR adds a set of linter rules almost as strict as was used in `atlas-osb`; some rules were relaxed due to the differences in architecture (use of `init()` functions, global variables, etc.)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,88 @@
+linters:
+  enable:
+    - bodyclose
+    - deadcode
+    - depguard
+    - dogsled
+    - dupl
+    - errcheck
+    - errorlint
+    - exportloopref
+    # - gochecknoinits
+    # - gochecknoglobals
+    - gocritic
+    - gocyclo
+    - gofumpt
+    - goimports
+    - goprintffuncname
+    - gosec
+    - gosimple
+    - govet
+    - ineffassign
+    - interfacer
+    - misspell
+    # - nlreturn
+    - nolintlint
+    - noctx
+    - prealloc
+    - rowserrcheck
+    - scopelint
+    - staticcheck
+    - structcheck
+    - stylecheck
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+    - whitespace
+    # - wrapcheck
+issues:
+  exclude-use-default: false
+  exclude:
+    # EXC0001 errcheck: Almost all programs ignore errors on these functions and in most cases it's ok
+    - Error return value of .((os\.)?std(out|err)\..*|.*Close|.*Flush|os\.Remove(All)?|.*print(f|ln)?|os\.(Un)?Setenv). is not checked
+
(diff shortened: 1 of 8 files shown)


.golangci.yml
```diff
@@ -0,0 +1,88 @@
+linters:
```
gssbzn, line 1: [nit] since you're enabling `misspell` let's configure it to american english
```suggestion
linters-settings:
  misspell:
    locale: US
linters:
```

.golangci.yml
```diff
@@ -0,0 +1,88 @@
...
+  enable:
```
gssbzn, line 2: [not-blocking] the golangci team sometimes recommends to disable all (`disable-all: true`) to prevent new versions of the linter introducing unexpected new linters but up to you if you want to disable all and manually enable what you need (which you are kind of already doing)

gssbzn (commented):
Left a couple of recommendations but I guess the main thing is to fix the flagged lint errors

.golangci.yml
vasilevp: Good catch, I'll add that.

pkg/api/v1/atlascustomresource.go
```diff
@@ -6,6 +6,7 @@ import (
 )
 
 //+k8s:deepcopy-gen=false
+
```
antonlisovenko, line 9: [q] does linter require a new line between annotations and normal comments?

.golangci.yml
```diff
@@ -0,0 +1,88 @@
...
+    - varcheck
+    - whitespace
+
+linters-settings:
+  misspell:
+    locale: US
+
+issues:
+  exclude-use-default: false
+  exclude:
```
antonlisovenko, line 42: [q] have you found these exclusions somewhere or added based on previous experience?

antonlisovenko (approved):
This looks great! Just a couple of questions.

I think we need to add the pre-commit hook for this, but not required for the current PR

gssbzn (approved):
LGTM

pkg/api/v1/atlascustomresource.go
vasilevp: It looks like kubebuilder still recognizes these comments, while godoc doesn't, so our docs aren't polluted with tags.

I'm actually considering changing this in Cluster as well, although it will bloat the file somewhat.

.golangci.yml
vasilevp: These are the default exclusions. I just disabled some of them to make the linters stricter.

pkg/api/v1/atlascustomresource.go
vasilevp: Answering your question: yes, the linter enforces a specific godoc comment format - it has to start with the name of whatever you're commenting, followed by its description. In this case the first thing was this k8s comment which did not pass the check.

PR #19, opened 2020-12-17
CLOUDP-78878: check if project already exists

antonlisovenko:
This PR handles the situation when the project already exists.
Initially I wanted to check the errors on creation (so `GroupAlreadyExists` or whatever) but this has some drawbacks (e.g. if the `status.ID` is wiped by the user somehow - it will never be fixed and all the next reconciliation logic will not work properly

Querying the projects before trying to create allows to find the ID and "fix" the status if the worst happened and somehow it disappeared.

Unfortunately, atlas-client API doesn't support the `name` filter so just scrolling through pages and looking for the project is done. In practice, this shouldn't be critical as the number of projects in the organization shouldn't be huge.


pkg/controller/atlasproject/project.go
```diff
@@ -16,26 +15,50 @@ func ensureProjectExists(ctx *workflow.Context, connection atlas.Connection, pro
 	if err != nil {
 		return "", workflow.Terminate(workflow.Internal, err.Error())
 	}
+	// Try to find the project
+	projectID, err := findProject(connection, project, client)
```
vasilevp, line 19: Please consider using [mongodbatlas.GetOneProjectByName()](https://github.com/mongodb/go-client-mongodb-atlas/blob/85a3c4de47becac8fa7887bea382190833c4ee0f/mongodbatlas/projects.go#L135) instead

antonlisovenko: This is a very bad endpoint and we had issues with it in Enterprise Operator. I don't remember details though
The good alternative is to use the "getProjects(name=<foo>)" but this is not supported by atlas client (though is exposed by the API)

antonlisovenko: After discussing with @gssbzn it seems like this endpoint should work ok for organization programmatic API keys - will adress

vasilevp (approved):

gssbzn (approved):
LGTM

PR #23, opened 2020-12-30
CLOUDP-79537: catch the API error if the project already exists

antonlisovenko:
There was a bug that the code didn't handle the "get project by name" properly. This returns 401 and it seems handling to error code is the best instead of HTTP header

### DIFF
--- a/pkg/controller/atlasproject/project.go
+++ b/pkg/controller/atlasproject/project.go
@@ -3,10 +3,10 @@ package atlasproject
 import (
 	"context"
 	"errors"
+
 	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/atlas"
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/workflow"
-	"github.com/prometheus/common/log"
 	"go.mongodb.org/atlas/mongodbatlas"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/controller/atlasproject/project.go
```diff
@@ -18,23 +19,24 @@ func ensureProjectExists(ctx *workflow.Context, connection atlas.Connection, pro
...
+			}
+			ctx.Log.Infow("Created Atlas Project", "name", project.Spec.Name, "id", p.ID)
+		} else {
+			return "", workflow.Terminate(workflow.ProjectNotCreatedInAtlas, err.Error())
+		}
 	}
-
-	if p, _, err = client.Projects.Create(context.Background(), p); err != nil {
-		return "", workflow.Terminate(workflow.ProjectNotCreatedInAtlas, err.Error())
+	if p == nil || p.ID == "" {
```
vasilevp, line 37: [nit] can we add a line break here as well?

pkg/controller/atlasproject/project.go
```diff
@@ -2,10 +2,11 @@ package atlasproject
 
 import (
 	"context"
-
+	"errors"
```
vasilevp, line 5: The linter most likely wants a newline here (which I humbly support)

pkg/controller/atlasproject/project.go
```diff
@@ -18,23 +19,24 @@ func ensureProjectExists(ctx *workflow.Context, connection atlas.Connection, pro
...
+			ctx.Log.Infow("Created Atlas Project", "name", project.Spec.Name, "id", p.ID)
+		} else {
+			return "", workflow.Terminate(workflow.ProjectNotCreatedInAtlas, err.Error())
+		}
 	}
-
-	if p, _, err = client.Projects.Create(context.Background(), p); err != nil {
-		return "", workflow.Terminate(workflow.ProjectNotCreatedInAtlas, err.Error())
+	if p == nil || p.ID == "" {
+		log.Error("Project or its project ID are empty")
```
vasilevp, line 38: Please change to `ctx.Log` - or maybe we could use this error msg as a `Reason`?

vasilevp (approved):
LGTM with a few small changes

pkg/controller/atlasproject/project.go
antonlisovenko: ah, that was an auto-completion so the `"github.com/prometheus/common/log"` is used here.. 🤦 

good catch

leo-ri (approved):

PR #29, opened 2021-01-18
CLOUDP-80550: support for observed generation

antonlisovenko:
This PR adds support for `status.observedGeneration` field that allows to understand when the Atlas Operator kicks into reconciling of the resource. The field is copied from `metadata.generation` on each status update. The first status update happens right after the resource is read at the beginning of reconciliation. This time the "main" condition "Ready" is set to `false`

The tests are updated to check the sequence of conditions now instead of "does the conditions contain this condition"


pkg/api/v1/status/status.go
```diff
@@ -14,8 +14,30 @@ type Writer interface {
 	UpdateStatus(conditions []Condition, option ...Option)
 }
 
+// +k8s:deepcopy-gen=false
+
+// Status is a generic status for any Custom Resource managed by Atlas Operator
+type Status interface {
```
antonlisovenko, line 20: the interface allows the generic access to the fields of any AO custom resource

pkg/controller/atlascluster/atlascluster_controller.go
```diff
@@ -44,22 +46,20 @@ func (r *AtlasClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
 		return result.ReconcileResult(), nil
 	}
 
-	log = log.With("clusterName", cluster.Spec.Name)
+	log.Infow("-> Starting AtlasCluster reconciliation", "spec", cluster.Spec)
+	defer statushandler.Update(wctx, r, cluster)
 
 	project := &mdbv1.AtlasProject{}
-	if result := customresource.GetResource(r.Client, req, project, log); !result.IsOk() {
+	if result := readProjectResource(r, cluster, project); !result.IsOk() {
```
antonlisovenko, line 58: @vasilevp  this is the fix for reading the project - now `PrepareResource` doesn't fit any more as we just simply need to read the Custom resource

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -78,7 +78,7 @@ func (r *AtlasProjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	statushandler.Update(ctx.SetConditionFalse(status.IPAccessListReadyType), r.Client, project)
 
 	// TODO projectAccessList
-
+	ctx.SetConditionTrue(status.ReadyType)
```
antonlisovenko, line 81: each reconciliation should set the "ready" to `true`

pkg/controller/customresource/customresource.go
```diff
@@ -3,30 +3,43 @@ package customresource
...
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-// GetResource queries the Custom Resource 'request.NamespacedName' and populates the 'resource' pointer.
-// Note the logic: any reconcile result different from nil should be considered as "terminal" and will stop reconciliation
-// right away (the pointer will be empty). Otherwise the pointer 'resource' will always reference the existing resource
-func GetResource(client client.Client, request reconcile.Request, resource runtime.Object, log *zap.SugaredLogger) workflow.Result {
+// PrepareResource queries the Custom Resource 'request.NamespacedName' and populates the 'resource' pointer.
+// It also performs some updates to the 'status' field to indicate the start of reconciliation.
+func PrepareResource(client client.Client, request reconcile.Request, resource mdbv1.AtlasCustomResource, log *zap.SugaredLogger) (workflow.Result, *workflow.Context) {
```
antonlisovenko, line 18: Now this function also returns the `workflow.Context` as it's initialized with the existing conditions from the CR (would be bad to have the conditions slice jumpting from  full to empty and back)

antonlisovenko: as discussed - rename back to `getResource` and call `indicateStartOfReconciliation` explicitely

vasilevp (approved):
My only concern is the hidden call to `indicateStartOfReconciliation`. Please make it explicit before merging. Otherwise LGTM!

PR #64, opened 2021-01-27
CLOUDP-81282: support for Atlas Domain configuration

crew-helper:
Adds support for configuring atlas domain. This is done using standard `flags` library as the (very nice) https://github.com/alexflint/go-arg is quite strict and fails if the unknown flags are passed (this happens when running tests)

New Kustomize overlay "prod" was added so now
`kustomize build config/prod`

returns the deployment (with `--atlas-domain` set to production):

```
template:
    metadata:
      labels:
        control-plane: controller-manager
    spec:
      containers:
      - args:
        - --secure-listen-address=0.0.0.0:8443
        - --upstream=http://127.0.0.1:8080/
        - --logtostderr=true
        - --v=10
        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
        name: kube-rbac-proxy
        ports:
        - containerPort: 8443
          name: https
      - args:
        - --atlas-domain=https://cloud.mongodb.com
        - --log-level=DEBUG
        - --metrics-addr=127.0.0.1:8080
        - --enable-leader-election
        command:
        - /manager
        image: localhost:5000/mongodb-atlas-kubernetes-operator:fix-api-groups-05114b9
        imagePullPolicy: Always
        name: manager
        resources:
          limits:
            cpu: 100m
            memory: 30Mi
          requests:
            cpu: 100m
            memory: 20Mi
      securityContext:
        runAsNonRoot: true
        runAsUser: 2000
      terminationGracePeriodSeconds: 10
```


.github/actions/gen-install-scripts/entrypoint.sh
```diff
@@ -6,6 +6,6 @@ mkdir "${target_dir}"
 # Generate configuration and save it to `all-in-one`
 controller-gen crd:crdVersions=v1 rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 cd config/manager && kustomize edit set image controller="${INPUT_IMAGE_URL}"
-cd - && kustomize build config/default > "${target_dir}"/all-in-one.yaml
+cd - && kustomize build config/prod > "${target_dir}"/all-in-one.yaml
```
crew-helper, line 9: this one is expected to produce the write `Deployment` config with `cloud.com` domain

config/manager/manager.yaml
```diff
@@ -22,6 +22,9 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      securityContext:
```
crew-helper, line 25: this one makes sure the pod is run as non-root

pkg/controller/config/operator_config.go
```diff
@@ -0,0 +1,25 @@
+package config
```
vasilevp, line 1: Shouldn't this be in `main`? CLI stuff doesn't seem to fit `controller`...

I think it's OK to pass Atlas URI to our reconcilers during construction; everything else is already only used from `main`.

pkg/controller/config/operator_config.go
```diff
@@ -0,0 +1,25 @@
...
+type Config struct {
+	AtlasDomain          string
+	EnableLeaderElection bool
+	MetricsAddr          string
+}
+
+// ParseConfiguration fills the 'OperatorConfig' from the flags passed to the program
+func ParseConfiguration() {
+	OperatorConfig = Config{}
+	flag.StringVar(&OperatorConfig.AtlasDomain, "atlas-domain", "https://cloud-qa.mongodb.com", "the Atlas URL domain name (no slash in the end).")
```
leo-ri, line 18: idk maybe cloud.mongodb.com should be the default?

crew-helper: yes, was a bit lazy but I think you are right

leo-ri (approved):

vasilevp (approved):

PR #71, opened 2021-01-29
CLOUDP-81194: fix for cluster update flakiness

antonlisovenko:
This is the fix for cluster update flakiness. See descriptions inlined

### DIFF
--- a/test/int/cluster_test.go
+++ b/test/int/cluster_test.go
@@ -126,6 +126,8 @@ var _ = Describe("AtlasCluster", func() {
 			Eventually(testutil.WaitFor(k8sClient, createdCluster, status.TrueCondition(status.ReadyType), validatePending),
 				1200, interval).Should(BeTrue())
 
+			fmt.Printf("Cluster updated!! %+v\n", createdCluster.Status)
+
 			Expect(createdCluster.Status.ConnectionStrings).NotTo(BeNil())
 			Expect(createdCluster.Status.ConnectionStrings.Standard).NotTo(BeNil())
 			Expect(createdCluster.Status.ConnectionStrings.StandardSrv).NotTo(BeNil())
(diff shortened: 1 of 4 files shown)


test/int/cluster_test.go
```diff
@@ -175,6 +173,25 @@ func clusterPendingFunc(expectedState, expectedMessage string, reason workflow.C
 		}
 	}
 }
+func validateClusterUpdatingFunc() func(a mdbv1.AtlasCustomResource) {
+	isIdle := true
+	return func(a mdbv1.AtlasCustomResource) {
+		c := a.(*mdbv1.AtlasCluster)
+		// It's ok if the first invocations see IDLE
+		if c.Status.StateName != "IDLE" {
```
antonlisovenko, line 181: as soon as the update is done the project may see "idle" state for some short time until the AO makes the API call for cluster update. 
This is not allowed in general though - only in the beginning, hence the "one side flag" (cannot be switched back)

test/int/integration_suite_test.go
```diff
@@ -110,7 +110,10 @@ var _ = BeforeSuite(func(done Done) {
 		Expect(err).ToNot(HaveOccurred())
 	}()
 
-	k8sClient = k8sManager.GetClient()
+	// It's recommended to construct the client directly for tests
+	// see https://github.com/kubernetes-sigs/controller-runtime/issues/343#issuecomment-469435686
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
```
antonlisovenko, line 115: this is the main treatment against caches - the client from `k8sManager` establishes caches (and shared informers?) for GET requests so for tests it's necessary to build the simple direct client

leo-ri (approved):

vasilevp (approved):

PR #72, opened 2021-01-29
Update mongodbatlas library & corresponding specs

vasilevp:


pkg/api/v1/status/atlascluster.go
```diff
@@ -72,7 +99,28 @@ func AtlasClusterMongoDBVersionOption(mongoDBVersion string) AtlasClusterStatusO
...
 	return func(s *AtlasClusterStatus) {
-		cs := ConnectionStrings(*connectionStrings)
+		pe := make([]PrivateEndpoint, 0, len(connectionStrings.PrivateEndpoint))
+		for _, v := range connectionStrings.PrivateEndpoint {
+			endpoints := make([]Endpoint, 0, len(v.Endpoints))
+			for _, e := range v.Endpoints {
+				endpoints = append(endpoints, Endpoint(e))
+			}
+
+			pe = append(pe, PrivateEndpoint{
```
antonlisovenko, line 109: [q] why cant' we build the `PrivateEndpoint` directly from the Atlas object?

pkg/api/v1/status/atlascluster.go
```diff
@@ -72,7 +99,28 @@ func AtlasClusterMongoDBVersionOption(mongoDBVersion string) AtlasClusterStatusO
...
+				ConnectionString:    v.ConnectionString,
+				Endpoints:           endpoints,
+				SRVConnectionString: v.SRVConnectionString,
+				Type:                v.Type,
+			})
+		}
+
+		cs := ConnectionStrings{
```
antonlisovenko, line 117: am I right that we do field-by-field building because of the custom `PrivateEndpoint` built above?

antonlisovenko (commented):
just questions about the implementation (would be great to describe this in the PR description/comments on PR creation)

Can we unit test the `AtlasClusterConnectionStringsOption` method as it got some building logic?

pkg/api/v1/status/atlascluster.go
vasilevp: Because it contains `[]mongodbatlas.Endpoint`, while our struct contains `[]status.Endpoint`. Even though the two `Endpoints` are directly convertible, their parent structs are not.

pkg/api/v1/status/atlascluster.go
vasilevp: Also because we're now missing the deprecated `AwsPrivateEndpoint...` fields

pkg/api/v1/status/atlascluster.go
antonlisovenko: what if the atlas-client and we add some field but forget to add the instruction here? 
Maybe we can serialize the `mongodbatlas.ConnectionStrings` into bytes and then deserialize into our object?

vasilevp: You're probably right - I'll use JSON the same way we do in controller

pkg/util/compat/jsoncopy.go
```diff
@@ -0,0 +1,18 @@
+package compat
```
antonlisovenko, line 1: [q] what does `compat` mean? :)

pkg/util/compat/jsoncopy.go
```diff
@@ -0,0 +1,18 @@
...
+
+import "encoding/json"
+
+// JSONCopy will copy src to dst via JSON serialization/deserialization.
+func JSONCopy(dst, src interface{}) error {
```
antonlisovenko, line 6: Nice to have this in library! Can we add the unit test for this? In this case the more high level unit tests won't be necessary. Just ideas for checking - what happens if the dst has some fields not known to src and vice versa.

antonlisovenko (commented):
looks great and much better maintaineable, can we add the unit test for `JSONCopy` please?

pkg/util/compat/jsoncopy.go
vasilevp: compatibility, reason being "stuff we need to be compatible with `mongodbatlas`"

pkg/util/compat/jsoncopy.go
vasilevp: This is actually well documented:
> By default, object keys which don't have a corresponding struct field are ignored (https://golang.org/pkg/encoding/json/#Unmarshal)

Thus, any mismatched fields/JSON keys will be silently omitted and left unchanged. (See https://goplay.space/#iPsozaeZY2W for an example)

That said, we're left with the following scenarios where this might fail:
1. Unconvertible types in dst/src with same name (i.e. `DiskSizeGB *int` <-> `DiskSizeGB *float64`)
2. Non-marshalable types in dst/src (i.e. `func()`)
3. Cyclic structs
4. Memory runs out

I find the 1st one to be at least somewhat relatable to our use cases, but we cannot test for it here.

vasilevp: That said, we can add a sanity-check unit test for this function if the goal is to bump coverage to 100% :wink:

antonlisovenko: thanks for the investigation above 👍 
I think it's good to have utility functions unit tested - this gives the confidence that no changes to them may break the main code. 
What do you think?

antonlisovenko (approved):
👍

PR #73, opened 2021-01-29
CLOUDP-77775: precommit hook

antonlisovenko:
This adds support for the nice https://pre-commit.com/ 

The golang-ci hook is used from https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml

I also moved the lint jobs to a separate workflow (and aggregated two jobs into one)

Note: you need to follow the new steps in the `start-guide.md` to enable precommit hook!


leo-ri (changes requested):
I'm not sure we should separate workflow:
-  I use `act -j shellcheck/golangci` separately...  (depends what was changed)
-  it will be faster like it was - in parallel. why do we need 1 job? if we need all check on push trigger we can run `act push`
- it is easier to watch... could change the name though

leo-ri (approved):

PR #77, opened 2021-02-01
CLOUDP-81076: Add more Cluster tests (happy path)

vasilevp:


test/int/cluster_test.go
```diff
@@ -68,72 +69,86 @@ var _ = Describe("AtlasCluster", func() {
...
+
+		doCommonChecks()
+	}
+
 	Describe("Create/Update the cluster", func() {
 		It("Should Succeed", func() {
 			expectedCluster := testAtlasCluster(namespace.Name, "test-cluster", createdProject.Name)
 
-			By(fmt.Sprintf("Creating the Cluster %s", kube.ObjectKeyFromObject(expectedCluster)))
+			By(fmt.Sprintf("Creating the Cluster %s", kube.ObjectKeyFromObject(expectedCluster)), func() {
```
antonlisovenko, line 99: [q] what is the advantage of having the logic in callbacks - is it some visual separation or anything else?

test/int/cluster_test.go
```diff
@@ -68,72 +69,86 @@ var _ = Describe("AtlasCluster", func() {
...
-			Expect(err).ToNot(HaveOccurred())
-
-			Expect(atlasCluster.Name).To(Equal(createdAtlasCluster.Name))
-			Expect(atlasCluster.Labels).To(Equal(createdAtlasCluster.Labels))
-			Expect(atlasCluster.ProviderSettings.InstanceSizeName).To(Equal(createdAtlasCluster.ProviderSettings.InstanceSizeName))
-			Expect(atlasCluster.ProviderSettings.ProviderName).To(Equal(createdAtlasCluster.ProviderSettings.ProviderName))
-			Expect(atlasCluster.ProviderSettings.RegionName).To(Equal(createdAtlasCluster.ProviderSettings.RegionName))
+			By("Updating the Cluster labels", func() {
+				createdCluster.Spec.Labels = []mdbv1.LabelSpec{{Key: "int-test", Value: "true"}}
+				performUpdate()
```
antonlisovenko, line 123: I think to make the test more reliable we need to add checks for AtlasCluster state in Atlas after each update. Thus we'll make sure we have changed the new field and haven't changed the others. We can move the common checks (providerSettings etc) to another function like you did for "doCommonChecks" and call that.

Maybe for making this more concise even create a function "checkInAtlas" and pass the callback to assert the object in Atlas leaving out boilerplate code of "GET" requests.

WDYT?

test/int/cluster_test.go
```diff
@@ -68,72 +69,86 @@ var _ = Describe("AtlasCluster", func() {
...
+		Expect(createdCluster.Status.ConnectionStrings.Standard).NotTo(BeNil())
+		Expect(createdCluster.Status.ConnectionStrings.StandardSrv).NotTo(BeNil())
+		Expect(createdCluster.Status.MongoDBVersion).NotTo(BeNil())
+		Expect(createdCluster.Status.MongoURIUpdated).NotTo(BeNil())
+		Expect(createdCluster.Status.StateName).To(Equal("IDLE"))
+		Expect(createdCluster.Status.Conditions).To(ConsistOf(testutil.MatchConditions(
+			status.TrueCondition(status.ClusterReadyType),
+			status.TrueCondition(status.ReadyType),
+		)))
+		Expect(createdCluster.Status.ObservedGeneration).To(Equal(createdCluster.Generation))
```
antonlisovenko, line 83: can we add one more check - to verify that both fields have increased after the update? This will make sure the updates were not missed by the Operator for any reasons. Some global counter may work out?

antonlisovenko (changes requested):
This looks great 👍 
One thing suggested is to add more checks after each update operation - not in the end. 
Tell me if you want to discuss this!

test/int/cluster_test.go
vasilevp: Yes, that way we have different "test cases" instead of different parts of one "test case"

test/int/cluster_test.go
```diff
@@ -68,72 +72,94 @@ var _ = Describe("AtlasCluster", func() {
...
-			// Atlas
-			atlasCluster, _, err = atlasClient.Clusters.Get(context.Background(), createdProject.Status.ID, createdCluster.Spec.Name)
-			Expect(err).ToNot(HaveOccurred())
-
 			createdAtlasCluster, err := createdCluster.Spec.Cluster()
 			Expect(err).ToNot(HaveOccurred())
 
 			Expect(atlasCluster.Name).To(Equal(createdAtlasCluster.Name))
-			Expect(atlasCluster.Labels).To(Equal(createdAtlasCluster.Labels))
+			Expect(atlasCluster.Labels).To(ConsistOf(createdAtlasCluster.Labels))
```
antonlisovenko, line 102: 👍  have also noticed this - quite convenient for slices

test/int/cluster_test.go
```diff
@@ -68,72 +72,94 @@ var _ = Describe("AtlasCluster", func() {
...
+			})
+
+			By("Decreasing the Cluster disk size", func() {
+				createdCluster.Spec.DiskSizeGB = intptr(10)
+				performUpdate()
+				checkAtlasState(func(c *mongodbatlas.Cluster) {
+					Expect(*c.DiskSizeGB).To(BeEquivalentTo(*createdCluster.Spec.DiskSizeGB))
+
+					// check whether https://github.com/mongodb/go-client-mongodb-atlas/issues/140 is fixed
+					Expect(c.DiskSizeGB).To(BeAssignableToTypeOf(float64ptr(0)), "DiskSizeGB is no longer a *float64, please check the spec!")
```
antonlisovenko, line 160: 👍

antonlisovenko (approved):
Great work!

PR #78, opened 2021-02-01
CLOUDP-80515: resource handler for watching the other resources

antonlisovenko:
This PR adds support for watching the resources that the Atlas resource depends on. E.g. AtlasCluster -> connection secret. So if any changes are done to the connection secret - the request for reconciliation of all dependent resources (e.g. two AtlasProjects referencing the same Secret) will be triggered for both AtlasProjects

The resource handler is implemented, it uses the map `map[WatchedObject][]types.NamespacedName{}` to check if the reconciliation needs to be triggered. The logics to ensure the state of the map is not implemented in this PR to avoid huge size - will be done in the following PR.


pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -84,9 +88,23 @@ func (r *AtlasProjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+
 func (r *AtlasProjectReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&mdbv1.AtlasProject{}).
+		Watches(
+			&source.Kind{Type: &mdbv1.AtlasCluster{}},
+			&watch.DeleteEventHandler{Controller: r},
+			builder.WithPredicates(watch.DeleteOnly()),
+		).
+		Watches(
```
antonlisovenko, line 104: we register the new watch for Secrets - so any events on Secrets will call the resource handler

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -24,15 +24,19 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // AtlasProjectReconciler reconciles a AtlasProject object
 type AtlasProjectReconciler struct {
 	client.Client
+	watch.ResourceWatcher
```
vasilevp, line 39: Is there a reason this can't be demoted to a field? I find regular fields to be way safer and more obvious than embedding in most cases.

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -59,7 +63,7 @@ func (r *AtlasProjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return reconcile.Result{}, nil
 	}
 	// This update will make sure the status is always updated in case of any errors or successful result
-	defer statushandler.Update(ctx, r, project)
+	defer statushandler.Update(ctx, r.Client, project)
```
vasilevp, line 66: May I suggest we make `r.Client` a regular field as well? I bet there will be more method naming conflicts if we don't

pkg/controller/watch/watched_resource_handler.go
```diff
@@ -0,0 +1,73 @@
...
+		ResourceKind: kind,
+		Resource:     types.NamespacedName{Name: name, Namespace: namespace},
+	}
+	for _, v := range c.TrackedResources[watchedResource] {
+		zap.S().Infof("%s has been modified -> triggering reconciliation for the %s", watchedResource, v)
+		q.Add(reconcile.Request{NamespacedName: v})
+	}
+}
+
+// Delete (Seems we don't need to react on watched resources removal..)
```
vasilevp, line 70: [q] Does this mean we already handle `Delete` as part of `Update` or do we just not want to react on `Delete`?

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -84,9 +88,23 @@ func (r *AtlasProjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+func (r *AtlasProjectReconciler) Delete(obj runtime.Object) error {
+	// TODO CLOUDP-80607
+	return nil
+}
+
 func (r *AtlasProjectReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&mdbv1.AtlasProject{}).
+		Watches(
+			&source.Kind{Type: &mdbv1.AtlasCluster{}},
```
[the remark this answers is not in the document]
antonlisovenko: bad copy-paste 🤦

pkg/controller/atlasproject/atlasproject_controller.go
antonlisovenko: I agree with you in general - especially for the `client.Client`!
Here though I intended to conform to the "is-a" model more - to point out that the reconciler "is a" resource watcher itself and pretends to contain that watched map internally...

WDYT?

antonlisovenko: I agree with you in general - especially for the `client.Client`!
Here though I intended to conform to the "is-a" model more - to point out that the reconciler "is a" resource watcher itself and pretends to contain that watched map internally...

WDYT?

pkg/controller/atlasproject/atlasproject_controller.go
antonlisovenko: sure, I thought you've fixed this in your other PR?
I had to do this because for some reasons this stopped compiling - though I'm ok if your fix supersedes this

pkg/controller/watch/watched_resource_handler.go
antonlisovenko: we just don't react. Maybe I'm wrong here but the result will be that the atlasProject may get to permanent failed state because of this if there is no global connection secret and will just constantly pollute logs. I thought that if the user recreates the connection secret than reacing on create will be enough..

pkg/controller/atlasproject/atlasproject_controller.go
vasilevp: My fix was for Cluster only to limit the scope of changes, which was large enough as it was.

We could, however, do this later as part of [CLOUDP-80607](https://jira.mongodb.org/browse/CLOUDP-80607) for the same reason.

antonlisovenko: ok, I'll do this then!

pkg/controller/atlasproject/atlasproject_controller.go
vasilevp: Not something I'm willing to die for - go ahead :smiley:

test/int/project_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var _ = Describe("AtlasProject", func() {
+var _ = FDescribe("AtlasProject", func() {
```
vasilevp, line 22: Is that intended?

vasilevp (approved):

test/int/project_test.go
antonlisovenko: no 🤦

PR #83, opened 2021-02-03
Organize imports & improve Makefile

vasilevp:
- To be discussed w/team:
  - Fix imports order: local imports (those beginning with `github.com/mongodb/mongodb-atlas-kubernetes/...`) are separated from third-party imports and are always last in the import list:
```go
import (
	// stdlib
	"io"
	... 

	// third-party packages
	"gopkg.in/yaml.v2"
	...

	// our packages
	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/something"
	...
)
```
- Add `make golangci-lint` & `make lint` to automatically install and run `golangci-lint` locally, similar to other tools we use
- Fix an annoying bug in `Makefile`: it will no longer try to interact with `kind` unless `run-kind` target is specified


.golangci.yml
```diff
@@ -35,6 +35,8 @@ linters:
 linters-settings:
   misspell:
     locale: US
+  goimports:
+    local-prefixes: github.com/mongodb/mongodb-atlas-kubernetes
```
vasilevp, line 39: Corresponding change in IDEs:
- VS Code:
```yaml
"gopls": {
    "formatting.local": "github.com/mongodb/mongodb-atlas-kubernetes/"
}
```
or
```yaml
"go.formatFlags": [
    "-local=github.com/mongodb/mongodb-atlas-kubernetes"
]
```
- GoLand: <details>
	  <summary>(Screenshot)</summary>
	  image
	  </details>

Makefile
```diff
@@ -129,11 +133,11 @@ bundle-build: ## Build the bundle image.
 
 .PHONY: run-kind
 run-kind: ## Create a local kind cluster
-ifeq ($(shell kind get clusters),kind)
-	@echo "create kind cluster: nothing to do"
-else
-	@bash ./scripts/create_kind_cluster.sh
-endif
+	@if [ "$(kind get clusters)" = "kind" ]; then \
```
vasilevp, line 136: Top-level `ifeq` statements are always executed, so `make` will spew `kind`-related errors even when this target is not requested. Also, this is just faster.

Makefile
```diff
@@ -71,6 +71,10 @@ fmt: ## Run go fmt against code
 vet: ## Run go vet against code
 	go vet ./...
 
+.PHONY: lint
+lint: golangci-lint ## Run golangci-lint against code
```
antonlisovenko, line 75: [q] do we really need a separate lint goal considering that we have a precommit hook that will fix the code on each commit?

As an alternative - I'd suggest to [call](https://pre-commit.com/#usage) `pre-commit run --all-files` instead that will run all hooks on all the files. So if we add more hooks - all of them will be run here automatically

Makefile
```diff
@@ -142,3 +146,12 @@ stop-kind: ## Stop the local kind cluster
 .PHONY: log
 log: ## View manager logs
 	kubectl logs deploy/mongodb-atlas-kubernetes-controller-manager manager -n mongodb-atlas-kubernetes-system -f
+
+.PHONY: golangci-lint
+golangci-lint: ## Find golangci-lint or download it if necessary
```
antonlisovenko, line 151: I think this one will be unnecessary if the comment above is addressed. The golint will be downloaded automatically then.

.golangci.yml
antonlisovenko: just fyi: have checked and the pre-commit hook linter handles this config perfectly, so IDE integration is optional - it will be anyway formated on commit.

antonlisovenko (commented):
Great change!
Just a thought that instead of calling the golang-lint directly we can call `pre-commit` tool that will do everything automatically

.golangci.yml
vasilevp: I actually did not set this up on my WSL machine yet, as pre-commit is failing with a python error...

vasilevp: @antonlisovenko Also - I thought `pre-commit` will only diagnose code, but not change it?

antonlisovenko: no, it calls the `golangci-lint`: https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml so performs the updates.

Makefile
vasilevp: I agree this is kind of redundant - will remove.

antonlisovenko (approved):
👍 
(there seem to be conflicts)

PR #84, opened 2021-02-03
Cloudp-80268 e2e initial

leo-ri:
part 1: base test
flaky parts: 
a) project creating (probably fixed)
b) sometimes 35min is not enough for updating cluster (at least in qa environment, increased to 50 min(?))

what is not ready:
configurations table


.github/workflows/test.yml
```diff
@@ -43,3 +44,95 @@ jobs:
           ATLAS_PUBLIC_KEY: ${{ secrets.ATLAS_PUBLIC_KEY }}
           ATLAS_PRIVATE_KEY: ${{ secrets.ATLAS_PRIVATE_KEY }}
 
+  prepare-e2e:
+    name: Prepare E2E configuration and image
+    needs: [unit-test, int-test]
```
antonlisovenko, line 49: can we please remove these and just run the test in parallel with the others?

.github/workflows/test.yml
```diff
@@ -43,3 +44,95 @@ jobs:
...
+    steps:
+
+    - name: Check out code
+      uses: actions/checkout@v2.3.1
+
+    - name: Prepare tag
+      id: prepare
+      uses: ./.github/actions/set-tag
+
+    - name: Push Atlas Operator to Registry
```
antonlisovenko, line 60: It seems, that if we push the dev image here - there's no need in the "Publish image to Registry" workflow?

.github/workflows/test.yml
```diff
@@ -43,3 +44,95 @@ jobs:
...
+      run: |
+        version=$(echo ${{ matrix.k8s }} | awk -F "-" '{print $1}')
+        platform=$(echo ${{ matrix.k8s }} | awk -F "-" '{print $2}')
+        echo "::set-output name=k8s_version::$version"
+        echo "::set-output name=k8s_platform::$platform"
+
+    # run if platform = kind #TODO
+    - name: Create k8s Kind Cluster
+      if: ${{ steps.properties.outputs.k8s_platform == 'kind' && !env.ACT }}
+      uses: helm/kind-action@v1.1.0
```
antonlisovenko, line 105: good to know there is an action for Kind already!

.github/workflows/test.yml
```diff
@@ -43,3 +44,95 @@ jobs:
...
+    - name: Set properties
+      id: properties
+      run: |
+        version=$(echo ${{ matrix.k8s }} | awk -F "-" '{print $1}')
+        platform=$(echo ${{ matrix.k8s }} | awk -F "-" '{print $2}')
+        echo "::set-output name=k8s_version::$version"
+        echo "::set-output name=k8s_platform::$platform"
+
+    # run if platform = kind #TODO
+    - name: Create k8s Kind Cluster
```
antonlisovenko, line 103: (for discussion)
If we setup the environment in the workflow itself - this will mean that we'll have to duplicate this workflow code for each separate Ginkgo test case that we run.
While if we move the "configure cluster" code into the ginkgo test itself (to the "prepare suite") - this would allow us to have one single workflow job and have all the parallelization inside the ginkgo instead

(let's discuss this with the team)

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,63 @@
...
+	Platform   = "kind"
+	K8sVersion = "v1.17.17"
+)
+
+func TestE2e(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "E2e Suite")
+}
+
+var _ = SynchronizedBeforeSuite(func() []byte {
```
antonlisovenko, line 33: [q] why do we need the `SynchronizedBeforeSuite` instead of the `BeforeSuite`?

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
+package e2e_test
```
antonlisovenko, line 1: [nit] can we please rename the `deploy_test.go` to something more meaningful, maybe `basic_test.go` or whatever?

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	cli "github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/utils"
+	. "github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("Deploy simple cluster", func() {
+
+	It("Release sample all-in-one.yaml should work", func() {
+		By("Prepare namespaces")
```
antonlisovenko, line 18: [nit] I found the design proposed by @vasilevp with `By` with code in callbacks looking very nice, WDYT about using it here?

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+		namespaceUserResources := "e2e-" + uuid.NewRandom().String() // TODO
+		namespaceOperator := "mongodb-atlas-kubernetes-system"
+		session := cli.Execute("kubectl", "create", "namespace", namespaceUserResources)
+		Expect(session).ShouldNot(Say("created"))
```
antonlisovenko, line 22: [q] why `ShouldNot` and not `Should`?

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+		session = cli.Execute("kubectl", "create", "namespace", namespaceOperator)
```
antonlisovenko, line 23: my understanding is that we don't need to create this namespace manually - the K8s scripts already have this command

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+		By("Create Sample Project\n")
+		session = cli.Execute("kubectl", "apply", "-f", ProjectSample, "-n", namespaceUserResources)
+		Eventually(session.Wait()).Should(Say("atlasproject.atlas.mongodb.com/my-project"))
+
+		By("Sample Cluster\n")
+		session = cli.Execute("kubectl", "apply", "-f", ClusterSample, "-n", namespaceUserResources)
+		Eventually(session.Wait()).Should(Say("atlascluster-sample created"))
+
+		By("Wait creating and check that it was created")
+		Eventually(cli.GetGeneration(namespaceUserResources)).Should(Equal("1"))
```
antonlisovenko, line 61: I think we need to wait for the "Ready": true condition in the `AtlasProject` resource. This is the main indication that the Project resource has reconciled

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+		Eventually(
```
antonlisovenko, line 62: after the Project is ready - then the Atlas is expected to have (not eventually) the project created

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+			cli.IsProjectExist(userProjectConfig.Spec.Name),
+			"7m", "10s",
+		).Should(BeTrue())
+
+		projectID := cli.GetProjectID(userProjectConfig.Spec.Name)
```
antonlisovenko, line 67: I'd recommend to get the project id from the `AtlasProject.status.id` instead of the mongocli. This will make sure that the Operator writes the correct id to the status

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+		GinkgoWriter.Write([]byte("projectID = " + projectID))
+
+		Eventually(
+			cli.IsClusterExist(projectID, userClusterConfig.Spec.Name),
```
antonlisovenko, line 71: the same as above: I'd recommend to check for the "status.conditions.Ready=true" instead of checking Atlas. This will make sure the status is updated correctly.

Basically the logic should be: wait for the AtlasCluster to get Ready:true, then make the request to Atlas to check if the cluster created there matches the one that we have just created

test/e2e/utils/cli.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// GenKubeVersion TODO move
+func GenKubeVersion(fullVersion string) string {
+	version := strings.Split(fullVersion, ".")
+	return fmt.Sprintf("Major:\"%s\", Minor:\"%s\"", version[0], version[1])
+}
+
+// GetPodStatus TODO move
+func GetPodStatus(ns string) func() string {
```
antonlisovenko, line 109: What do you think about splitting the helper functions for mongocli and kubectl to separate packages? Then we'd import them with the appropriate names and it would be clear from the calling code what is happening under the hood

test/e2e/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+			cluster.ProviderSettings.RegionName,
+		).Should(Equal(userClusterConfig.Spec.ProviderSettings.RegionName))
+
+		By("Update cluster\n")
+		session = cli.Execute("kubectl", "apply", "-f", "data/updated_atlascluster_basic.yaml", "-n", namespaceUserResources) // TODO param
+		Eventually(session.Wait()).Should(Say("atlascluster-sample configured"))
+
+		By("Wait creation")
+		userClusterConfig = cli.LoadUserClusterConfig("data/updated_atlascluster_basic.yaml")
+		Eventually(cli.GetGeneration(namespaceUserResources)).Should(Equal("2"))
```
antonlisovenko, line 98: The same as for the create logic the update the code should wait for two things:
1. the observed generation is equal to the generation field (it's ok to check for "2" directly as well)
2. the "status.conditions.Ready : true"

antonlisovenko (changes requested):
Great work! 🥇  The main suggestion is around the test logic (check for the resource status.Ready) field.

We can discuss the cluster configuration part with the team later!

test/e2e/e2e_suite_test.go
leo-ri: this check run once instead of each time

test/e2e/deploy_test.go
leo-ri: I use exec command inside , eventually here means : this command eventually finish with exit 0 (there is a wait() inside) . probably need to discuss

test/e2e/utils/cli.go
leo-ri: that s why I added "TODO move" )

.github/workflows/test.yml
theburi: agree with int-test as they take 20 min. I think it makes sense to wait for unit-test as it take a min, but will avoid an unnecessary run of e2e in case of error.

.github/workflows/test.yml
leo-ri: I removed push trigger from build image workflow, but I would like to leave this workflow for a while until I am sure that e2e are stable

test/e2e/deploy_test.go
leo-ri: `configuration_test.go` is ok? it will be contain at least 4

test/e2e/deploy_test.go
leo-ri: ofc it is "should"

test/e2e/deploy_test.go
leo-ri: hm... it probably could fix this:
a) project creating

.github/workflows/test.yml
antonlisovenko: forget that - I think I forgot that we wanted to move the "e2e suite name" to the matrix params. This solves the duplication problem

test/e2e/configuration_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/pborman/uuid"
+
+	. "github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+
+	cli "github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/utils"
+)
+
+var _ = Describe("Deploy simple cluster", func() {
+
```
antonlisovenko, line 18: we need to remove the cluster in `AfterEach` to ensure we don't leave rubbish, see int tests

antonlisovenko (commented):
Reviewed p2p

test/e2e/deploy_test.go
leo-ri: deleted

antonlisovenko (approved):
Just a couple of nits - mostly around namings and code formatting. I believe after committing there will be no need in waiting for the tests.

Great work! 🥇

test/e2e/configuration_test.go
```diff
@@ -0,0 +1,156 @@
...
+	var ID string
+
+	var _ = AfterEach(func() {
+		GinkgoWriter.Write([]byte(ID))
+		Eventually(cli.DeleteNamespace("ns-" + ID)).Should(Say("deleted"))
+		// cli.DeleteCluster(ID, "cluster45") // TODO struct
+	})
+
+	It("Release sample all-in-one.yaml should work", func() {
+		By("Prepare namespaces and project configuration") // TODO clusters/keys will be a bit later
```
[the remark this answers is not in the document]
leo-ri: the thing is idk which steps will be final, this test - is the initial version, steps will be reformatted later for all tests need

PR #85, opened 2021-02-03
CLOUDP-80515: watch secret (part 2)

antonlisovenko:
This PR adds a proper support for watcher and covers that with unit and int tests.

The comments in the PR and code should give more insight into the things.


pkg/controller/atlascluster/atlascluster_controller.go
```diff
@@ -108,7 +108,7 @@ func (r *AtlasClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		WithEventFilter(watch.CommonPredicates()).
 		Watches(
 			&source.Kind{Type: &mdbv1.AtlasCluster{}},
-			&watch.DeleteEventHandler{Controller: r},
+			&watch.AtlasResourceEventHandler{Controller: r},
```
antonlisovenko, line 111: have changed the name as this handler inherits the generic "create/update" handler as well (and is used for projects below)

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -54,6 +54,10 @@ func (r *AtlasProjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	if !result.IsOk() {
 		return result.ReconcileResult(), nil
 	}
+	if project.ConnectionSecretObjectKey() != nil {
+		r.EnsureResourcesAreWatched(req.NamespacedName, "Secret", log, *project.ConnectionSecretObjectKey())
```
antonlisovenko, line 58: this is where the connection secret watch registration happens. If the connection secret reference has changed - this will handle removal of the previous secret and will register the new one

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -94,18 +98,21 @@ func (r *AtlasProjectReconciler) Delete(obj runtime.Object) error {
...
-			&watch.DeleteEventHandler{Controller: r},
-			builder.WithPredicates(watch.DeleteOnly()),
-		).
-		Watches(
-			&source.Kind{Type: &corev1.Secret{}},
-			&watch.ResourcesHandler{TrackedResources: r.WatchedResources},
-		).
-		WithEventFilter(watch.CommonPredicates()).
-		Complete(r)
+	c, err := controller.New("AtlasProject", mgr, controller.Options{Reconciler: r})
```
antonlisovenko, line 101: the builder is awful, I fought with it but it just resulted in duplicated handlings....
So I did everything directly
the cluster manager so far doesn't need this though

pkg/controller/watch/resource_watcher.go
```diff
@@ -1,17 +1,72 @@
...
 	return ResourceWatcher{
-		WatchedResources: map[WatchedObject][]types.NamespacedName{},
+		WatchedResources: map[WatchedObject][]client.ObjectKey{},
 	}
 }
 
+// ResourceWatcher is the object containing the map of watched_resource -> []dependant_resource.
 type ResourceWatcher struct {
-	WatchedResources map[WatchedObject][]types.NamespacedName
+	WatchedResources map[WatchedObject][]client.ObjectKey
```
vasilevp, line 17: What do you think about converting this to a set? Logically, that's what it currently looks like anyway with `pos()` and `remove()` and the whole keeping values unique business...

pkg/controller/watch/resource_watcher.go
```diff
@@ -1,17 +1,72 @@
...
+	found := false
+	for _, v := range r.WatchedResources[key] {
+		if v == dependentResourceNsName {
+			found = true
+		}
+	}
+	if !found {
+		r.WatchedResources[key] = append(r.WatchedResources[key], dependentResourceNsName)
+		log.Debugf("Watching %s to trigger reconciliation for %s", key, dependentResourceNsName)
+	}
```
vasilevp, line 46: If `r.WatchedResources[key]` were to be a set, this would actually be reduced to just `r.WatchedResources[key][dependentResourceNsName] = {}` ...

pkg/controller/watch/resource_watcher.go
```diff
@@ -1,17 +1,72 @@
...
+		if k == key {
+			return i
+		}
+	}
+	return -1
+}
+
+func remove(slice []client.ObjectKey, s int) []client.ObjectKey {
+	return append(slice[:s], slice[s+1:]...)
+}
```
vasilevp, line 72: ... and this would be gone altogether

pkg/controller/watch/resource_watcher.go
```diff
@@ -1,17 +1,72 @@
...
+
+func (r ResourceWatcher) cleanNonWatchedResources(dependant client.ObjectKey, resourceKind string, watchedKeys []client.ObjectKey) {
+	for k, v := range r.WatchedResources {
+		if pos(watchedKeys, k.Resource) < 0 || k.ResourceKind != resourceKind {
+			var dependantPos int
+			if dependantPos = pos(v, dependant); dependantPos >= 0 {
+				// we found the old dependency (not watched any more) so we need to remove it
+				r.WatchedResources[k] = remove(r.WatchedResources[k], dependantPos)
+			}
```
vasilevp, line 56: ... this would become `delete(v, dependant)` ...

vasilevp (changes requested):

pkg/controller/watch/resource_watcher.go
antonlisovenko: I think you mean the `map[client.ObjectKey]bool`?
yes, this will work out, will make the code more concise 👍

vasilevp (approved):

PR #86, opened 2021-02-04
CLOUDP-79276: unit test cluster types compatibility

antonlisovenko:
Test for checking that all cluster fields including embedded are used unless explicitely ignored.


pkg/api/v1/atlascluster_types_test.go
```diff
@@ -0,0 +1,117 @@
...
+			ft = ft.Elem()
+		}
+		if _, ok := excludedFields[theTag]; !ok && ft.Kind() == reflect.Struct {
+			res = append(res, theTag)
+		}
+	}
+	return res
+}
+
+func getAllFieldsSorted(theStruct interface{}, excluded map[string]bool) []string {
```
antonlisovenko, line 94: all the fields in the struct in sorted order with explicit exclusion

pkg/api/v1/atlascluster_types_test.go
```diff
@@ -0,0 +1,117 @@
...
+	excludedClusterFieldsTheirs["mongoURIUpdated"] = true
+	excludedClusterFieldsTheirs["mongoURIWithOptions"] = true
+	excludedClusterFieldsTheirs["connectionStrings"] = true
+	excludedClusterFieldsTheirs["srvAddress"] = true
+	excludedClusterFieldsTheirs["stateName"] = true
+
+	// CLOUDP-80765
+	excludedClusterFieldsTheirs["paused"] = true
+
+	// TODO why not included?
```
vasilevp, line 39: This was implemented [11 days ago](https://github.com/mongodb/go-client-mongodb-atlas/commit/41343eb93a4848048a0e5a6f6d98dc421cecc3c2#diff-cdec3370b78f74b0e08291a906d41ff1be66a9a9c0f7a5d3cfbbf9fe7ad23e2eR33) - our spec is out of date already 🙂

pkg/api/v1/atlascluster_types_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"reflect"
+	"sort"
+	"testing"
+
+	"github.com/fatih/structtag"
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/atlas/mongodbatlas"
+)
+
+var excludedClusterFieldsOurs = map[string]bool{}
```
vasilevp, line 13: FYI: you can actually use empty `struct{}` as value type to avoid allocations (but please don't bother).

vasilevp (approved):
LGTM!

pkg/api/v1/atlascluster_types_test.go
antonlisovenko: ah, ok, great to know!

pkg/api/v1/atlascluster_types_test.go
antonlisovenko: wow, nice! Seems like the test is working :) 
I'll add the field in this PR

PR #88, opened 2021-02-05
CLOUDP-80607: Implement project delete logic

vasilevp:


pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -104,8 +139,14 @@ func (r *AtlasProjectReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
-	// Watch for changes to primary resource MongoDbReplicaSet
-	err = c.Watch(&source.Kind{Type: &mdbv1.AtlasProject{}}, &watch.AtlasResourceEventHandler{Controller: r}, watch.CommonPredicates())
+	// Watch for changes to primary resource AtlasProject
+	err = c.Watch(&source.Kind{Type: &mdbv1.AtlasProject{}}, &handler.EnqueueRequestForObject{}, watch.CommonPredicates())
```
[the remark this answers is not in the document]
antonlisovenko: [q] is there any advantage in two `Watch` calls instead of one?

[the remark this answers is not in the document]
vasilevp: IMO the advantage is that we can better separate the "standard" watcher from our custom one, and it's more obvious that way that those are meant to do different things

test/int/cluster_test.go
```diff
@@ -60,9 +58,14 @@ var _ = Describe("AtlasCluster", func() {
 			if createdCluster != nil {
 				By("Removing Atlas Cluster " + createdCluster.Name)
 				Expect(k8sClient.Delete(context.Background(), createdCluster)).To(Succeed())
-
 				Eventually(checkAtlasClusterRemoved(createdProject.Status.ID, createdCluster.Name), 600, interval).Should(BeTrue())
 			}
+
+			// TODO: CLOUDP-82115
```
antonlisovenko, line 64: can we add TODO but leave the code here? Otherwise our projects won't be removed and we'll hit the Atlas limits (250 projects per organization)

test/int/project_test.go
```diff
@@ -47,8 +48,8 @@ var _ = Describe("AtlasProject", func() {
 	AfterEach(func() {
 		if createdProject != nil && createdProject.Status.ID != "" {
 			By("Removing Atlas Project " + createdProject.Status.ID)
-			_, err := atlasClient.Projects.Delete(context.Background(), createdProject.Status.ID)
-			Expect(err).ToNot(HaveOccurred())
+			Expect(k8sClient.Delete(context.Background(), createdProject)).To(Succeed())
+			Eventually(checkAtlasProjectRemoved(createdProject.Status.ID), 600, interval).Should(BeTrue())
```
antonlisovenko, line 52: [nit] would suggest to use the smaller timeout - 10 seconds?

pkg/controller/watch/event_handler.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"go.uber.org/zap"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/kube"
+)
+
+type ResourceEventHandler struct {
```
antonlisovenko, line 12: as discussed p2p - it should be ok to use the `AtlasResourceEventHandler` here and "override" the delete handle

antonlisovenko (changes requested):
Really good - the only thing is the RequestHandler as discussed p2p

test/int/cluster_test.go
antonlisovenko: ah sorry dismiss this - I see the project is removed below 👍

pkg/controller/watch/event_handler.go
```diff
@@ -0,0 +1,29 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/kube"
+)
+
+// EventHandlerWithDelete is an extension of EnqueueRequestForObject that will _not_ trigger a reconciliation for a Delete event.
+// Instead, it will call an external controller's Delete() method and pass the event argument unchanged.
+type EventHandlerWithDelete struct {
```
antonlisovenko, line 14: 👍

test/int/project_test.go
antonlisovenko: up - in my understanding in this case when there are no clusters in the project the removal time is expected to be very short or am I missing something?

antonlisovenko (approved):
Nice work! Thanks for addressing the comments 👍

test/int/project_test.go
vasilevp: right - that's a copy-paste issue, will fix

PR #89, opened 2021-02-05
CLOUDP-80765: Handle Cluster pause/unpause

vasilevp:


test/int/cluster_test.go
```diff
@@ -160,6 +160,22 @@ var _ = Describe("AtlasCluster", func() {
 					Expect(c.DiskSizeGB).To(BeAssignableToTypeOf(float64ptr(0)), "DiskSizeGB is no longer a *float64, please check the spec!")
 				})
 			})
+
+			By("Pausing the cluster", func() {
+				createdCluster.Spec.Paused = boolptr(true)
+				performUpdate()
+				checkAtlasState(func(c *mongodbatlas.Cluster) {
+					Expect(c.Paused).To(Equal(createdCluster.Spec.Paused))
+				})
```
antonlisovenko, line 169: can we check that after the cluster paused/unpaused - the following updates to the other fields can be done? 
Also I guess we need to make sure that those updates don't change the cluster in Atlas (paused/unpaused) back to `nil`

pkg/controller/atlascluster/cluster.go
```diff
@@ -48,15 +48,28 @@ func (r *AtlasClusterReconciler) ensureClusterState(log *zap.SugaredLogger, conn
...
+		if cluster.Spec.Paused != nil {
+			if c.Paused == nil || *c.Paused != *cluster.Spec.Paused {
+				// paused is different from Atlas
+				// we need to first send a special (un)pause request before reconciling everything else
+				spec = &mongodbatlas.Cluster{
+					Paused: cluster.Spec.Paused,
+				}
+			} else {
+				// otherwise, don't send the paused field
+				spec.Paused = nil
```
antonlisovenko, line 71: interesting! Will this update the `paused` in Atlas to nil as well? How does Atlas interpret paused as nil - may it start the cluster back?

antonlisovenko (changes requested):
Interesting logic with nullifying 'paused' and performing a one-time change request.
Looks good, the only suggestion is to add update requests after pause/unpause to make sure that the explicit nullyfing the field hasn't affected anything in atlas

pkg/controller/atlascluster/cluster.go
vasilevp: `omitempty` in JSON spec means that `nil` is not included in the payload

test/int/cluster_test.go
vasilevp: [Atlas does not support updates on paused clusters:](https://docs.atlas.mongodb.com/reference/api/clusters-modify-one/#request-body-parameters)
> You cannot modify a paused cluster other than to resume the cluster. Nor can you pause a cluster with pending changes.

antonlisovenko: ok, thanks for this clarification.
Then I guess the following behavior is expected?
- the user updates "paused: true"
- then the user changes other fields in the spec, controller sends this to Atlas but no changes happen there as the cluster is paused? The AtlasCluster resource is in "ready: true" state though?
- then the user updates "pause: false" and the changes done in the previous step reach Atlas now?

WDYT about testing the above scenario or maybe you'll find a better scenario to check

pkg/controller/atlascluster/cluster.go
antonlisovenko: thanks for the information, I think this should be safe 👍

test/int/cluster_test.go
vasilevp: > then the user changes other fields in the spec, controller sends this to Atlas but no changes happen there as the cluster is paused? The AtlasCluster resource is in "ready: true" state though?

The cluster will be in a non-ready state because such updates will fail on Atlas' side. I'm not sure whether we want to make it "ready" in this scenario though...

antonlisovenko: as discussed p2p having the cluster as "not ready" is a correct outcome 👍  This is something we can confirm with the test

test/int/cluster_test.go
```diff
@@ -160,6 +160,42 @@ var _ = Describe("AtlasCluster", func() {
...
+			By("Unpausing the cluster", func() {
+				createdCluster.Spec.Paused = boolptr(false)
+				performUpdate()
+				checkAtlasState(func(c *mongodbatlas.Cluster) {
+					Expect(c.Paused).To(Equal(createdCluster.Spec.Paused))
+				})
+			})
+
+			By("Checking that modifications were applied after unpausing", func() {
+				checkAtlasState(func(c *mongodbatlas.Cluster) {
```
antonlisovenko, line 195: can we also check that the AtlasCluster has now all status conditions set to true?

antonlisovenko (approved):
Thanks for the test, just an additional check in the test and should be ready to go!

PR #90, opened 2021-02-05
CLOUDP-78181: project ip access list (part 1)

antonlisovenko:
This PR adds support for project IP access list. Anything in Atlas that doesn't match the spec is removed from Atlas, the rest is added. This ensures that if any ip access list entry is removed from the spec - it will be removed in Atlas as well.

The basic int test is added. 

More unit and int tests will be added in the next PR


pkg/controller/atlasproject/ipaccess_list.go
```diff
@@ -0,0 +1,129 @@
...
+	return nil
+}
+
+func filterActiveIPAccessLists(project *mdbv1.AtlasProject) ([]mdbv1.ProjectIPAccessList, []mdbv1.ProjectIPAccessList) {
+	active := make([]mdbv1.ProjectIPAccessList, 0)
+	expired := make([]mdbv1.ProjectIPAccessList, 0)
+	for _, list := range project.Spec.ProjectIPAccessList {
+		if list.DeleteAfterDate != "" {
+			// We are ignoring the error as it will never happen due to validation check before
+			iso8601, _ := timeutil.ParseISO8601(list.DeleteAfterDate)
```
vasilevp, line 115: This will error out if time format is different, which can happen in theory since ISO 8601 can range in precision and the presence of TZ info (see below)

pkg/util/timeutil/timeutil.go
```diff
@@ -0,0 +1,7 @@
+package timeutil
+
+import "time"
+
+func ParseISO8601(dateTime string) (time.Time, error) {
+	return time.Parse(dateTime, "2006-01-02T15:04:05-0700")
```
vasilevp, line 6: Two issues:
1. The arguments are in the wrong order
2. Atlas can specify the time in UTC format (this is always true for access lists), which will cause an error: `2021-02-08T19:42:36Z`

pkg/controller/workflow/reason.go
```diff
@@ -12,7 +12,9 @@ const (
 
 // Atlas Project reasons
 const (
-	ProjectNotCreatedInAtlas ConditionReason = "ProjectNotCreatedInAtlas"
+	ProjectNotCreatedInAtlas   ConditionReason = "ProjectNotCreatedInAtlas"
+	ProjectIPAccessInvalid     ConditionReason = "ProjectIpAccessListInvalid"
```
vasilevp, line 16: [q] Why not keep the Go-style casing in reasons?

vasilevp (changes requested):

pkg/util/timeutil/timeutil.go
antonlisovenko: yes, you are right - this is fixed in the next PR :)

pkg/controller/atlasproject/ipaccess_list.go
antonlisovenko: yes, you are right - this is fixed in the next PR!

pkg/controller/workflow/reason.go
antonlisovenko: ah, you mean `ProjectIPAccessListInvalid` instead of `ProjectIpAccessListInvalid`? 
yes, makes sense 👍

pkg/controller/atlasproject/ipaccess_list.go
```diff
@@ -0,0 +1,129 @@
...
+
+func validateIPAccessLists(ipAccessList []mdbv1.ProjectIPAccessList) error {
+	for _, list := range ipAccessList {
+		if list.DeleteAfterDate != "" {
+			_, err := timeutil.ParseISO8601(list.DeleteAfterDate)
+			if err != nil {
+				return err
+			}
+		}
+		// Go doesn't support XOR, but uses '!=' instead: https://stackoverflow.com/a/23025720/614239
```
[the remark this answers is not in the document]
antonlisovenko: yes! :)

pkg/controller/atlasproject/ipaccess_list.go
```diff
@@ -0,0 +1,129 @@
...
+		onlyOneSpecified := isNotEmpty(list.AwsSecurityGroup) != isNotEmpty(list.CIDRBlock) != isNotEmpty(list.IPAddress)
```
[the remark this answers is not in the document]
antonlisovenko: *Very* good catch, addressed in the next PR: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/91/commits/916d061a7202f0e6bc0e18785cbd947ba939c200

vasilevp (approved):

pkg/util/timeutil/timeutil.go
antonlisovenko: sorry haven't replied to the second point:
MMS has the following formatting for the date: `yyyy-MM-dd'T'HH:mm:ssX` `X` in java is quite flexible - it can be the timestamp of any format, starting from `Z` and ending with `+0800`.

PR #92, opened 2021-02-08
CLOUDP-78181: ip access list (part 3)

antonlisovenko:
This is part 3 of ip access list support. Should be reviewed after https://github.com/mongodb/mongodb-atlas-kubernetes/pull/91

Adds more tests and the functionality to handle expired IP Access Lists

Expired IP Access lists are not sent to Atlas but instead are written to the status


pkg/api/v1/status/atlasproject.go
```diff
@@ -18,4 +23,29 @@ type AtlasProjectStatus struct {
...
 	ID string `json:"id,omitempty"`
+
+	// The list of IP Access List entries that are expired due to 'deleteAfterDate' being less than the current date.
+	// Note, that this field is updated by the Atlas Operator only after specification changes
+	ExpiredIPAccessList []ProjectIPAccessList `json:"expiredIpAccessList,omitempty"`
+}
+
+// Copy of mdbv1.ProjectIPAccessList
+// TODO solve circular dependency (move ProjectIPAccessList to subpackage?)
+
```
vasilevp, line 34: Let's change these lines to avoid TODO appearing in generated files:
```suggestion
// TODO solve circular dependency (move ProjectIPAccessList to subpackage?)

// ProjectIPAccessList is a copy of mdbv1.ProjectIPAccessList
```

pkg/controller/atlasproject/ipaccess_list.go
```diff
@@ -20,14 +22,20 @@ import (
 type atlasProjectIPAccessList mongodbatlas.ProjectIPAccessList
 
 func (i atlasProjectIPAccessList) Identifier() interface{} {
+	// hack: Atlas adds the CIDRBlock in case IPAddress is specified in the response.
+	// This breaks the contract (one field per List) and doesn't allow to "merge" lists.
+	// So we ignore the CIDRblock in this case.
+	if i.CIDRBlock != "" && i.IPAddress != "" && strings.HasPrefix(i.CIDRBlock, i.IPAddress) {
```
vasilevp, line 28: Are you sure this `strings.HasPrefix(i.CIDRBlock, i.IPAddress)` check is correct?

Say we get the following (technically valid) output:
```yaml
"cidrBlock": "192.0.2.0/24",
"ipAddress": "192.0.2.15",
```

The check would fail in this case.

pkg/controller/atlasproject/ipaccess_list.go
```diff
@@ -20,14 +22,20 @@ import (
 type atlasProjectIPAccessList mongodbatlas.ProjectIPAccessList
 
 func (i atlasProjectIPAccessList) Identifier() interface{} {
+	// hack: Atlas adds the CIDRBlock in case IPAddress is specified in the response.
+	// This breaks the contract (one field per List) and doesn't allow to "merge" lists.
```
vasilevp, line 26: It doesn't actually break the contract, since it's not mentioned anywhere that the API **_response_** may only have one of the three 🙂

test/int/project_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/testutil"
 )
 
-var _ = Describe("AtlasProject", func() {
+var _ = FDescribe("AtlasProject", func() {
```
vasilevp:
```suggestion
var _ = Describe("AtlasProject", func() {
```

vasilevp (changes requested):
Nice work, but I have some concerns over the CIDR check

pkg/controller/atlasproject/ipaccess_list.go
antonlisovenko: yes, this will be correct.
The problem is that when we create the IP Access list with some IP - Atlas sets the CIDRBlock with the mask :32:

we:
```
"ipAddress": "192.0.2.15"
```
atlas response:
```
"ipAddress": "192.0.2.15"
"cidrBlock": "192.0.2.15/32",
```

considering the rule that only one field can be set in the requests I believe that if both fields are set and IPAddress is prefix of cidrBlock - this means that the cidrBlock was generated... 
(unless I'm missing something)

pkg/controller/atlasproject/ipaccess_list.go
antonlisovenko: yeah.. but this breaks our "joining" logic which looks ad "IDs" of joined entities. In this case we have to "denormalize" .
But I'll change the comment, you are right

pkg/controller/atlasproject/ipaccess_list.go
vasilevp: Maybe just ignore the CIDR block altogether if IP is set?

vasilevp: I see... I was confused by this example from the docs that made it seem like it's possible to have a /24 mask: https://docs.atlas.mongodb.com/reference/api/ip-access-list/add-entries-to-access-list/#example-body

antonlisovenko: ah! I see now! That's a good catch - I think the example is buggy, I'll create a DOCS ticket to fix that (and will also ask someone from the Atlas team to review that)

vasilevp: Awesome!

vasilevp (approved):
LGTM!

pkg/controller/atlasproject/ipaccess_list.go
antonlisovenko: the final thought - looking at the examples I think I'll be on a more safe side and will use the logic suggested (no prefix check) as I could be missing something and the CIDR may be calculated magically and can have a mask different from the :32

Thanks for point out 👍

PR #96, opened 2021-02-09
CLOUDP-80516: global connection secret

antonlisovenko:
This PR adds the support for the global connection secret (the one that the Operator defaults to if the project-level one is not specified). The Secret name must follow the pattern "<operator-pod-name>-api-key".


config/manager/manager.yaml
```diff
@@ -40,4 +40,13 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
+        env:
```
antonlisovenko, line 43: this is the "downward API" in Kubernetes - that allows to pass some Kuberentes information (pod name, namespace, annotations etc) to the container

vasilevp (approved):

PR #97, opened 2021-02-10
CLOUDP-80774: support enums in the spec

antonlisovenko:
This PR transforms some of the fields to Enums. Unit tests are added to ensure behavior is not broken.
Note, that only two fields are added: `ProviderName` and `ClusterType`. The suggestion is to do this gradually and also for those enums that are quite stable and "protected" by CRD enum validation.


pkg/api/v1/atlascluster_types.go
```diff
@@ -30,6 +30,20 @@ func init() {
...
+type ProviderName string
+type ClusterType string
+
+const (
+	ProviderAWS    ProviderName = "AWS"
+	ProviderGCP    ProviderName = "GCP"
+	ProviderAzure  ProviderName = "AZURE"
+	ProviderTenant ProviderName = "TENANT"
+
+	TypeReplicaSet ClusterType = "REPLICASET"
```
vasilevp, line 42: [nit] It's good practice to separate different types into different `const` blocks. This also allows the Go documentation generator to parse comments for such blocks as relevant documentation.

Unofficial example: https://pkg.go.dev/github.com/fluhus/godoc-tricks#Mock_enums

vasilevp (approved):
LGTM, just a small style suggestion

leo-ri (approved):

PR #98, opened 2021-02-11
CLOUDP-81143: Add failure scenarios to Cluster int tests

vasilevp:


config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
```diff
@@ -124,8 +124,7 @@ spec:
                   this field is updated by the Atlas Operator only after specification
                   changes
                 items:
-                  description: TODO solve circular dependency (move ProjectIPAccessList
-                    to subpackage?) Copy of mdbv1.ProjectIPAccessList
+                  description: ProjectIPAccessList is a copy of mdbv1.ProjectIPAccessList
```
antonlisovenko, line 127: thanks!

test/int/cluster_test.go
```diff
@@ -122,6 +122,43 @@ var _ = Describe("AtlasCluster", func() {
...
+				createdCluster.ObjectMeta = expectedCluster.ObjectMeta
+				Expect(k8sClient.Create(context.Background(), expectedCluster)).ToNot(HaveOccurred())
+
+				Eventually(
+					testutil.WaitFor(
+						k8sClient,
+						createdCluster,
+						status.
+							FalseCondition(status.ClusterReadyType).
+							WithReason(string(workflow.Internal)).
```
antonlisovenko, line 139: [q]  I would expect this to be `ClusterNotCreatedInAtlas` instead of `Internal` or am I missing something?

antonlisovenko (commented):
just one question

test/int/cluster_test.go
antonlisovenko: Ok, I know the answer. It's not the `create` request that fails, but the `client.Clusters.Get(ctx, project.Status.ID, cluster.Spec.Name)` and this results in the internal error.

Can we please add the comment here to explain this as logically it seems that the creation should fail, not get

antonlisovenko (approved):
Nice! 🥳 
would be great to add the comment in the test about why it's 'internal'

PR #106, opened 2021-02-15
Move IPAccessList to a separate file

antonlisovenko:
The refactoring of `IPAccessList` struct - moving it to the subpackage to avoid circular dependency as it's used from the the `status` subpackage


test/int/cluster_test.go
```diff
@@ -151,7 +151,7 @@ var _ = Describe("AtlasCluster", func() {
 
 				Expect(k8sClient.Update(context.Background(), createdCluster)).To(Succeed())
 
-				Eventually(testutil.WaitFor(k8sClient, createdCluster, status.TrueCondition(status.ReadyType), validateClusterCreatingFunc()),
+				Eventually(testutil.WaitFor(k8sClient, createdCluster, status.TrueCondition(status.ReadyType)),
```
antonlisovenko, line 154: @vasilevp  this fixes the test flakiness as the `validateClusterCreatingFunc` expects that the cluster to be in good shape and here it isn't

vasilevp (approved):

PR #107, opened 2021-02-15
CLOUDP-82585: Database User CRD

antonlisovenko:
This is a Database User types and CRD, no controller has been added yet

Note, that the int test failing is flaky - fixed in the parallel PR: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/106


pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -0,0 +1,135 @@
...
+	// Labels is an array containing key-value pairs that tag and categorize the database user.
+	// Each key and value has a maximum length of 255 characters.
+	Labels []LabelSpec `json:"labels,omitempty"`
+
+	// Roles is an array of this user's roles and the databases / collections on which the roles apply. A role allows
+	// the user to perform particular actions on the specified database.
+	Roles []RoleSpec `json:"roles"`
+
+	// Scopes is an array of clusters and Atlas Data Lakes that this user has access to.
+	Scopes []ScopeSpec `json:"scopes"`
```
vasilevp, line 64: Scopes should be `omitempty` to support project-wide users

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+type ScopeType string
+
+const (
+	ClusterScopeType  ScopeType = "CLUSTER"
+	DataLakeScopeType ScopeType = "DATA_LAKE"
+)
+
+// AtlasDatabaseUserSpec defines the desired state of Database User in Atlas
+type AtlasDatabaseUserSpec struct {
```
vasilevp, line 43: [q] Am I right that x509, LDAP, etc. are coming in at a later stage?

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -0,0 +1,135 @@
...
+// AtlasDatabaseUserList contains a list of AtlasDatabaseUser
+type AtlasDatabaseUserList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []AtlasDatabaseUser `json:"items"`
+}
+
+// RoleSpec allows the user to perform particular actions on the specified database.
+// A role on the admin database can include privileges that apply to the other databases as well.
+type RoleSpec struct {
```
vasilevp, line 99: It looks like all fields in a `Role` are required, so we need to remove `omitempty`

pkg/api/v1/atlasdatabaseuser_types.go
antonlisovenko: I think in the scope document there were suggestions to make this mandatory (I believe from Jason Mimick): https://docs.google.com/document/d/1hZrcQy18Calv5hwkxvcgHO_AHKusy6-O2by7uaETdi8/edit#bookmark=id.4q1yljo2rfvh

I'm not sure what is the best approach here, I think we need the second opinion on this, @jamesbroadhead @theburi what do you think?

pkg/api/v1/atlasdatabaseuser_types.go
antonlisovenko: yes, it's out of scope for trial

pkg/api/v1/atlasdatabaseuser_types.go
antonlisovenko: good catch!

pkg/api/v1/atlasdatabaseuser_types.go
jamesbroadhead: I'll defer to @theburi. It seems to me that we should support project-wide users

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// AtlasDatabaseUserSpec defines the desired state of Database User in Atlas
+type AtlasDatabaseUserSpec struct {
+	// Project is a reference to AtlasProject resource the user belongs to
+	Project ResourceRef `json:"projectRef"`
+
+	// DatabaseName is a Database against which Atlas authenticates the user. Must be equal to 'admin'
+	// as only SCRAM authentication is supported so far.
+	DatabaseName string `json:"databaseName"`
```
jamesbroadhead, line 49: can this be hardcoded & not appear in the CRD until we support more auth forms?

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -0,0 +1,135 @@
...
+	// The specified date must be in the future and within one week of the time you make the API request.
+	DeleteAfterDate string `json:"deleteAfterDate,omitempty"`
+
+	// Labels is an array containing key-value pairs that tag and categorize the database user.
+	// Each key and value has a maximum length of 255 characters.
+	Labels []LabelSpec `json:"labels,omitempty"`
+
+	// Roles is an array of this user's roles and the databases / collections on which the roles apply. A role allows
+	// the user to perform particular actions on the specified database.
+	Roles []RoleSpec `json:"roles"`
```
jamesbroadhead, line 61: What's the default behaviour if this list is left empty?

pkg/api/v1/atlasdatabaseuser_types.go
jamesbroadhead: ^ and that not requiring people to fill this field would be good for demo use-cases

antonlisovenko: Have just checked - so having the `scopes` a mandatory still should allow to get project-wide users by just specifying `scopes:[]`

So this will be an intentional configuration done by a user, the positive sign I see is that they won't dismiss this just because they are not aware of such field.
WDYT?

jamesbroadhead: probably nice if they could leave it out & for the operator to use '[]'

pkg/api/v1/atlasdatabaseuser_types.go
antonlisovenko: let me check if I understand you correctly: do you want to remove the comment from CRD? Because as soon as we support other auth forms - the comment won't be actual anymore, it's only for the Trial version of the Operator.

pkg/api/v1/atlasdatabaseuser_types.go
antonlisovenko: CRD validation will pass, now I'm thinking that we need to add validation for the list shouldn't be empty.

vasilevp (approved):

pkg/api/v1/atlasdatabaseuser_types.go
jamesbroadhead: I mean Atlas-side -- what will Atlas do if this list is left empty?

pkg/api/v1/atlasdatabaseuser_types.go
jamesbroadhead: I would drop this field from the CRD entirely -- and then add it when there are values that the user should specify

I don't think it's good design to have a field where the user must type the string "admin" -- it makes work for them , without any benefit.

jamesbroadhead:
@antonlisovenko  remember that I asked you to include Ent Op engineers on CRD reviews -- they have lots of experience with CRDs, so might have helpful input

pkg/api/v1/atlasdatabaseuser_types.go
vasilevp: Alternatively, we could add a default value via `kubebuilder`, like `// +kubebuilder:default=admin`

antonlisovenko: @jamesbroadhead I see your point. Though I have concerns about omitting this field now because when we make this mandatory in the future when we start supporting more auth modes - this will break the existing CRs for the users and they will have to modify the specifications to fix tjos. I'd rather made this field mandatory from the beginning or used the default value suggested by Pavel

chatton: defaulting to `admin` sounds like a good idea here, I don't think it's a good user experience to need to specify the same value no matter what.

pkg/api/v1/atlasdatabaseuser_types.go
chatton: I'm also interested in this, I know the agent does not handle a null value for users. I believe it will correctly. create a user with no roles but it would be good to be able to confirm this!

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -0,0 +1,136 @@
...
+type AtlasDatabaseUserSpec struct {
+	// Project is a reference to AtlasProject resource the user belongs to
+	Project ResourceRef `json:"projectRef"`
+
+	// DatabaseName is a Database against which Atlas authenticates the user. Must be equal to 'admin'
+	// as only SCRAM authentication is supported so far.
+	DatabaseName string `json:"databaseName"`
+
+	// DeleteAfterDate is a timestamp in ISO 8601 date and time format in UTC after which Atlas deletes the user.
+	// The specified date must be in the future and within one week of the time you make the API request.
```
chatton, line 52: I think this comment has a little too much implementation detail leaked in, I think the only information necessary for the user is that `The specified date must be in the future and within one week`, the API request is the operator's job to handle!

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -0,0 +1,136 @@
...
+	DeleteAfterDate string `json:"deleteAfterDate,omitempty"`
+
+	// Labels is an array containing key-value pairs that tag and categorize the database user.
+	// Each key and value has a maximum length of 255 characters.
```
chatton, line 56: We should enforce this in both fields with validations` +kubebuilder:validation:MaxLength:=255`

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// DatabaseName is a database on which the user has the specified role. A role on the admin database can include
+	// privileges that apply to the other databases.
+	DatabaseName string `json:"databaseName"`
+
+	// CollectionName is a collection for which the role applies.
+	CollectionName string `json:"collectionName"`
+}
+
+// ScopeSpec if presents a database user only have access to the indicated resource (Cluster or Atlas Data Lake)
```
chatton, line 112:
```suggestion
// ScopeSpec if present a database user only have access to the indicated resource (Cluster or Atlas Data Lake)
```

pkg/api/v1/status/atlasdatabaseuser.go
```diff
@@ -0,0 +1,11 @@
+package status
+
+// +k8s:deepcopy-gen=false
+
+// AtlasDatabaseUserStatusOption is the option that is applied to Atlas Project Status
+type AtlasDatabaseUserStatusOption func(s *AtlasDatabaseUserStatus)
+
+// AtlasDatabaseUserStatus defines the observed state of AtlasProject
+type AtlasDatabaseUserStatus struct {
+	Common `json:",inline"`
```
chatton, line 10: [q] is Common used elsewhere? Seems like we can just put the fields in here directly instead?

chatton (approved):
LGTM left a few comments/concerns,

I think we should definitely add the validations on the LabelSpec.

pkg/api/v1/atlasdatabaseuser_types.go
antonlisovenko: yes, good catch!

pkg/api/v1/status/atlasdatabaseuser.go
antonlisovenko: yes, it's used in two other statuses (for cluster and project)

pkg/api/v1/atlasdatabaseuser_types.go
antonlisovenko: The Atlas returns an error if Roles is empty:
```
{"detail":"The required attribute roles was not specified.","error":400,"errorCode":"MISSING_ATTRIBUTE","parameters":["roles"],"reason":"Bad Request"}
```

jamesbroadhead: ack thanks

jamesbroadhead (approved):
thanks for addressing

PR #119, opened 2021-02-17
Refactoring client and connection to Context

antonlisovenko:
As discussed on the sync-up: moving the `client` and `connection` to `Context` to enable better reuse


pkg/controller/atlas/connection.go
```diff
@@ -27,7 +26,7 @@ type Connection struct {
 
 // ReadConnection reads Atlas API connection parameters from AtlasProject Secret or from the default Operator one if the
 // former is not specified
-func ReadConnection(log *zap.SugaredLogger, kubeClient client.Client, operatorPodObjectKey client.ObjectKey, projectOverrideSecretRef *client.ObjectKey) (Connection, workflow.Result) {
+func ReadConnection(log *zap.SugaredLogger, kubeClient client.Client, operatorPodObjectKey client.ObjectKey, projectOverrideSecretRef *client.ObjectKey) (Connection, error) {
```
antonlisovenko, line 29: had to change this to return `error` otherwise there was a circular dependency with `workflow` package (as we added `Connection` to the `Context`)

vasilevp (approved):

PR #120, opened 2021-02-17
CLOUDP-82586: basic database user controller

antonlisovenko:
This PR adds support for the database user controller. The simple create test was added.
The implementation follows the patterns used for the other controllers. Note, that no "wait for clusters to propagate" is implemented yet - this is due to be done in the following PRs


pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -106,7 +113,7 @@ type RoleSpec struct {
 	DatabaseName string `json:"databaseName"`
 
 	// CollectionName is a collection for which the role applies.
-	CollectionName string `json:"collectionName"`
+	CollectionName string `json:"collectionName,omitempty"`
```
antonlisovenko, line 116: while the API docs mark this parameter as required - in fact it's possible to omit it for some of the roles

test/e2e/cli/mongocli/mongocli.go
```diff
@@ -95,5 +95,5 @@ func GetClusterStateName(projectID string, clusterName string) string {
 
 func GetVersionOutput() {
 	session := cli.Execute("mongocli", "--version")
-	session.Wait()
+	session.Wait(10)
```
leo-ri, line 98: default wait() timeout for this suite is equal to 1min. Why?

scripts/e2e_local.sh
```diff
@@ -1,10 +1,14 @@
...
+set -eou pipefail
+
 act -j build-push
 
 public_key=$(grep "ATLAS_PUBLIC_KEY" .actrc | cut -d "=" -f 2)
 private_key=$(grep "ATLAS_PRIVATE_KEY" .actrc | cut -d "=" -f 2)
 org_id=$(grep "ATLAS_ORG_ID" .actrc | cut -d "=" -f 2)
-image=$(grep "DOCKER_REPO" .actrc | cut -d "=" -f 2):$(git rev-parse --abbrev-ref HEAD)-$(git rev-parse --short HEAD)
+# this is the format how it's pushed by act -j build-push
+image=$(grep "DOCKER_REGISTRY" .env | cut -d "=" -f 2)/mongodb-atlas-kubernetes-operator:$(git rev-parse --abbrev-ref HEAD)-$(git rev-parse --short HEAD)
```
leo-ri, line 11: am I missing something? why do we remove DOCKER_REPO if it is still a .env property and it is used in other workflows? for local testing I usually use my own name and for debug workflow it is also useful since I can use my repo name and test full run

leo-ri (changes requested):

test/e2e/cli/mongocli/mongocli.go
antonlisovenko: discussed p2p: this is just to avoid waiting for 1 min as this command is expected to return fast

scripts/e2e_local.sh
antonlisovenko: Discussed and clarified together, tell me Svetlana if you have more questions!

leo-ri (approved):

vasilevp (approved):

PR #121, opened 2021-02-18
Using builders for types

antonlisovenko:
This PR introduces builders for main types used. This will make writing tests simpler and easier.
Following the Enterprise Operator experience `apis` package looks like the best place as both unit and int tests will be able to reuse it. One option could be to have a `builders` package and put builders there but this would result in circular dependency. So I thought about avoidiing `builder` pattern and just add the "fluent" methods to types themselves.


pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -168,3 +168,48 @@ func (p AtlasDatabaseUser) ToAtlas(kubeClient client.Client) (*mongodbatlas.Data
...
+			Roles:          []RoleSpec{},
+		},
+	}
+}
+
+func (p *AtlasDatabaseUser) WithName(name string) *AtlasDatabaseUser {
+	p.Name = name
+	return p
+}
+func (p *AtlasDatabaseUser) WithDBUserName(name string) *AtlasDatabaseUser {
```
vasilevp, line 193: [nit] There are some inconsistencies: in Cluster & Project it's `WithAtlasName`, but here we have `WithDBUserName`. Maybe we could change this to `WithAtlasName` as well?

vasilevp (approved):

PR #123, opened 2021-02-19
CLOUDP-82257: simple test app for checking user data

leo-ri:
simple app for the testing needs


test/app/k8s/deployment.json
```diff
@@ -0,0 +1,7 @@
+[
```
antonlisovenko, line 1: [q] what is this file for?

test/app/main.go
```diff
@@ -0,0 +1,136 @@
...
+	r.HandleFunc("/mongo/", deleteKeyValue).Methods("DELETE")
+	return r
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprintf(w, "It is working")
+}
+
+func getKeyValue(w http.ResponseWriter, r *http.Request) {
+	ship := getShipFromRequest(r)
```
antonlisovenko, line 50: what do you think if we pass the `id` parameter instead of passing the whole `Ship` object? It seems that we only need to find the object in database by id?

test/app/main.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	response, _ := json.Marshal(res)
+	fmt.Fprintf(w, string(response))
+}
+
+func deleteKeyValue(w http.ResponseWriter, r *http.Request) {
+	ship := getShipFromRequest(r)
```
antonlisovenko, line 81: the same as above - I'd suggest passing the id here instead of the whole object, what do you think?

antonlisovenko (approved):
Looks good for the start! 👍 
Just a couple of questions/comments but not-blocking

test/app/k8s/deployment.json
leo-ri: I was thinking to patch the secret name/image
```
 valueFrom:
              secretKeyRef:
                name:
```

test/app/main.go
leo-ri: I was thinking maybe we could add UI for manual test/demonstration purposes later. Or we don't need that ?

antonlisovenko (approved):
👍

PR #137, opened 2021-03-01
CLOUDP-82255-ddt

leo-ri:


test/e2e/configuration_test.go
```diff
@@ -2,118 +2,138 @@ package e2e_test
...
+		project := utils.NewProject().
+			ProjectName(userSpec.projectName).
+			SecretRef(userSpec.keyName).
+			CompleteK8sConfig(userSpec.k8sProjectName)
+		utils.SaveToFile(FilePathTo(userSpec.projectName), project)
+
+		userSpec.clusters = append(userSpec.clusters, utils.LoadUserClusterConfig(clusterConfigurationFile))
+		userSpec.clusters[0].Spec.Project.Name = userSpec.k8sProjectName
+		userSpec.clusters[0].Spec.ProviderSettings.InstanceSizeName = "M10"
+		utils.SaveToFile(
```
antonlisovenko, line 69: am I right that the thing we are doing here: load yaml file for cluster, change a couple of fields and save back?
Is it possible to specify these 2 fields directly in the file to avoid overwriting it?

test/e2e/configuration_test.go
```diff
@@ -2,118 +2,138 @@ package e2e_test
...
+		},
+		Entry(newData("Simple Test with default cluster", "data/atlascluster_basic.yaml")),
+		Entry(newData("Simple Backup Configuration and 2 users", "data/atlascluster_backup.yaml")),
+		// Entry(newData("Multiregion, Backup and 2 users", "data/atlascluster_multiregion.yaml")), // TODO CLOUDP-83419
+	)
 })
+
+func mainCycle(clusterConfigurationFile string, userSpec userInputs) {
+	By("Prepare namespaces and project configuration", func() {
+		kube.CreateNamespace(userSpec.namespace)
```
antonlisovenko, line 58: [nit] I think this shouldn't be necessary as kustomize will do it later

test/e2e/configuration_test.go
```diff
@@ -2,118 +2,138 @@ package e2e_test
...
+		userSpec.clusters = append(userSpec.clusters, utils.LoadUserClusterConfig(clusterConfigurationFile))
+		userSpec.clusters[0].Spec.Project.Name = userSpec.k8sProjectName
+		userSpec.clusters[0].Spec.ProviderSettings.InstanceSizeName = "M10"
+		utils.SaveToFile(
+			userSpec.clusters[0].ClusterFileName(),
+			utils.JSONToYAMLConvert(userSpec.clusters[0]),
+		)
+	})
+
+	By("Apply namespaced operator configuration\n", func() {
```
antonlisovenko, line 75: maybe "Create namespaced Operator"?

test/e2e/configuration_test.go
```diff
@@ -2,118 +2,138 @@ package e2e_test
...
+		utils.CreateCopyKustomizeNamespace(userSpec.namespace)
+		kube.Apply("-k", "data/"+userSpec.namespace)
+		Eventually(
+			kube.GetPodStatus(userSpec.namespace),
+			"5m", "3s",
+		).Should(Equal("Running"), "The operator should successfully run")
+	})
+
+	By("Create users resources", func() {
+		kube.CreateKeySecret(userSpec.keyName, userSpec.namespace)
```
antonlisovenko, line 85: [nit] maybe "CreateApiKeySecret"?

test/e2e/configuration_test.go
```diff
@@ -2,118 +2,138 @@ package e2e_test
...
+	}
+
+	DescribeTable("Namespaced operators working only with its own namespace with different configuration",
+		func(test testDataProvider) {
+			data = test
+			mainCycle(test.confPath, test.userSpec)
+		},
+		Entry(newData("Simple Test with default cluster", "data/atlascluster_basic.yaml")),
+		Entry(newData("Simple Backup Configuration and 2 users", "data/atlascluster_backup.yaml")),
+		// Entry(newData("Multiregion, Backup and 2 users", "data/atlascluster_multiregion.yaml")), // TODO CLOUDP-83419
```
antonlisovenko, line 52: can we please create a ticket to uncomment this as soon as the fix for ReplicationSpecs arrives?

test/e2e/configuration_test.go
```diff
@@ -2,118 +2,138 @@ package e2e_test
...
+	By("Wait cluster creation", func() {
+		waitCluster(userSpec, "1")
+	})
+
+	By("check cluster Attribute", func() {
+		cluster := mongocli.GetClustersInfo(userSpec.projectID, userSpec.clusters[0].Spec.Name)
+		compareClustersSpec(userSpec.clusters[0].Spec, cluster)
+	})
+
+	By("Update cluster\n", func() {
```
antonlisovenko, line 104: I think this is ok for the current (simple) tests but we may need to do more context-dependent updates for more complicated specs.
So maybe when the replicationSpecs arrive - I'd suggest doing something about changing that field (adding replicas/changing regions etc)

This may require adding another parameter to the table params - the "update file"

test/e2e/configuration_test.go
```diff
@@ -2,118 +2,138 @@ package e2e_test
...
-		GinkgoWriter.Write([]byte(userSpec.projectID))
-		Eventually(kube.DeleteNamespace(userSpec.namespace)).Should(Say("deleted"))
+
+		if CurrentGinkgoTestDescription().Failed {
+			GinkgoWriter.Write([]byte("Resources wasn't clean"))
+			kube.GetManagerLogs(data.userSpec.namespace)
+		} else {
+			Eventually(kube.DeleteNamespace(data.userSpec.namespace)).Should(Say("deleted"))
+		}
+		// Eventually(kube.DeleteNamespace(data.userSpec.namespace)).Should(Say("deleted"))
```
antonlisovenko, line 33: [nit] remove these 2 comments?

antonlisovenko (commented):
Just a couple of questions and small comments.

Massive work!!

test/e2e/configuration_test.go
leo-ri: yes, you're right.

.github/workflows/test.yml
```diff
@@ -83,7 +83,7 @@ jobs:
       matrix:
         # k8s: ["1.17-kind", "1.19-kind", "1.17-opeshift"] # <supported platform version>-<platform>
         k8s: ["v1.18.15-kind"] # <K8sGitVersion>-<Platform>
-        test: ["sample all-in-one", "cluster wide operator"] # TODO refactor
+        test: ["cluster-ns", "cluster-wide"] # TODO refactor
```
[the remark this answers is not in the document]
leo-ri: to support "this test only for that platform". now it runs all cases from the matrix (I was thinking to rewrite it after I get a special test like that, so it is just for me - I can forget)

test/e2e/configuration_test.go
antonlisovenko: am I right that it didn't work out and we need to create the ns manually?

test/e2e/configuration_test.go
antonlisovenko: yes, you are right, we cannot use the hardcoded project name 👍

antonlisovenko (approved):
Amazing work!! 🥇

test/e2e/configuration_test.go
leo-ri: yeah, or I need create Kustomize patch with replace for namespace

Issue #126, opened 2021-02-22
Error when updating cluster autoscaling


matheusca:
Hi!

I'm using `mongodb-atlas-operator:0.3.0`

When updating cluster autoscaling, I'm getting an error.

The first initial atlas cluster configuration was:
```yaml
apiVersion: atlas.mongodb.com/v1
kind: AtlasCluster
metadata:
  annotations:
    app.aioros.io/revision: 32392083902832
  labels:
    app: chitauri-web-mongodb-platform
  name: service-name
  namespace: mongodb-atlas-system
spec:
  autoScaling:
    compute:
      enabled: true
      scaleDownEnabled: true
  clusterType: REPLICASET
  name: service-name
  projectRef:
    name: development
  providerBackupEnabled: true
  providerSettings:
    autoScaling:
      compute:
        maxInstanceSize: M40
        minInstanceSize: M30
    instanceSizeName: M30
    providerName: GCP
  replicationSpecs:
    - numShards: 1
      regionsConfig:
        EASTERN_US:
          analyticsNodes: 0
          electableNodes: 1
          priority: 6
          readOnlyNodes: 0
        SOUTH_AMERICA_EAST_1:
          analyticsNodes: 0
          electableNodes: 2
          priority: 7
          readOnlyNodes: 0
      zoneName: Zone 1
```

But when changing to:
```yaml
apiVersion: atlas.mongodb.com/v1
kind: AtlasCluster
metadata:
  annotations:
    app.aioros.io/revision: 32392083902832
  labels:
    app: chitauri-web-mongodb-platform
  name: service-name
  namespace: mongodb-atlas-system
spec:
  autoScaling:
    compute:
      enabled: true
      scaleDownEnabled: true
  clusterType: REPLICASET
  name: service-name
  projectRef:
    name: development
  providerBackupEnabled: true
  providerSettings:
    autoScaling:
      compute:
        maxInstanceSize: M50
        minInstanceSize: M40
    instanceSizeName: M40
    providerName: GCP
  replicationSpecs:
    - numShards: 1
      regionsConfig:
        EASTERN_US:
          analyticsNodes: 0
          electableNodes: 1
          priority: 6
          readOnlyNodes: 0
        SOUTH_AMERICA_EAST_1:
          analyticsNodes: 0
          electableNodes: 2
          priority: 7
          readOnlyNodes: 0
      zoneName: Zone 1
```

I got this error in `mongodb-atlas-operator`: 
```
[...]

PR #138, opened 2021-03-01
Add long-running test cases & implement slice merge for ReplicationSpecs

vasilevp:


matheusca:
Maybe this PR fixed error described here: https://github.com/mongodb/mongodb-atlas-kubernetes/issues/126

I think the problem it is, more attributes are being sent to the API during the cluster update and it is not necessary.

pkg/controller/atlascluster/cluster.go
```diff
@@ -84,22 +84,48 @@ func (r *AtlasClusterReconciler) ensureClusterState(ctx *workflow.Context, proje
...
+		return
+	}
+
+	if err = compat.JSONCopy(&result, spec); err != nil {
+		return
 	}
 
-	if err := compat.JSONCopy(&clusterMerged, spec); err != nil {
-		return false, err
+	// TODO: might need to do this with other slices
```
antonlisovenko, line 111: can we file a ticket for this? Seems like we need to cover all slices with tests...

pkg/util/compat/json_slice_merge.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	minLen := dstVal.Len()
+	if srcVal.Len() < minLen {
+		minLen = srcVal.Len()
+	}
+
+	// merge common elements
+	for i := 0; i < minLen; i++ {
+		dstX := dstVal.Index(i).Addr().Interface()
```
antonlisovenko, line 42: [q] am I right that this will work only if the items of the slice are pointers as well? (looking at the tests below seems only the destination slice must have pointers?)
If so - may need to add to the method documentation

pkg/util/compat/json_slice_merge_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	. "github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/compat"
+)
+
+func TestJSONSliceMerge(t *testing.T) {
```
antonlisovenko, line 11: Great to have the unit test!
Do you think we can add the use cases:
* when the src has smaller size than the dst?
* the src is empty/nil
* the dst is empty/nil

test/int/cluster_test.go
```diff
@@ -112,15 +112,66 @@ var _ = Describe("AtlasCluster", func() {
...
+				Expect(k8sClient.Create(context.Background(), expectedCluster)).ToNot(HaveOccurred())
+
+				Eventually(testutil.WaitFor(k8sClient, createdCluster, status.TrueCondition(status.ReadyType), validateClusterCreatingFunc()),
+					1800, interval).Should(BeTrue())
+
+				doCommonChecks()
+				checkAtlasState()
+			})
+
+			By("Increasing DiskSizeGB", func() {
```
antonlisovenko, line 166: [q] do you think we need the separate test for increasing DiscSize if we have the one decreasing it? https://github.com/mongodb/mongodb-atlas-kubernetes/blob/2ba4112c78a3257973e5c01dcaeb95ad8569fcf3/test/int/cluster_test.go#L189-L188

test/int/cluster_test.go
```diff
@@ -112,15 +112,66 @@ var _ = Describe("AtlasCluster", func() {
...
 		Expect(k8sClient.Update(context.Background(), createdCluster)).To(Succeed())
 
 		Eventually(testutil.WaitFor(k8sClient, createdCluster, status.TrueCondition(status.ReadyType), validateClusterUpdatingFunc()),
-			1200, interval).Should(BeTrue())
+			timeout, interval).Should(BeTrue())
 
 		lastGeneration++
 	}
 
+	Describe("Create cluster & change ReplicationSpecs", func() {
```
antonlisovenko, line 124: What do you think about adding another a bit more complicated use case and specifying two different regions for one replicationSpec (and some configuration for analytics/read-only nodes)?

The current use-case is nice as it covers our merging logic (take the things from Atlas and merge ours on top of the default ones) though we don't have the coverage for more complicated scenario when the user spreads his cluster over regions)

antonlisovenko (changes requested):
This looks great and the generic approach to slices is a good one (strong reflection 👍 )

I have some suggestions for unit and int tests - tell me what you think?

pkg/util/compat/json_slice_merge.go
vasilevp: They don't have to be pointers - I used pointers in the test case just to check that pointers work too

test/int/cluster_test.go
vasilevp: Yes, as part of #138

vasilevp: This one takes a lot of time due to disk provisioning, so I decided against including it in the "main" test suite

vasilevp:
@matheusca 
> 
> 
> Maybe this PR fixed error described here: #126
> 
> I think the problem it is, more attributes are being sent to the API during the cluster update and it is not necessary.

I doubt this PR will fix the issue right away - but it's the necessary first step to support this specific use case.

test/int/cluster_test.go
antonlisovenko: as discussed p2p it may make more sense to cover something different (like `providerSettings.instanceSizeName` or whatever) as DiscSize is already covered in the other testcase)

antonlisovenko: as discussed p2p it may make more sense to cover something different (like `providerSettings.instanceSizeName` or whatever) as DiscSize is already covered in the other testcase)

test/int/cluster_test.go
```diff
@@ -112,15 +112,111 @@ var _ = Describe("AtlasCluster", func() {
...
+					1800, interval).Should(BeTrue())
+
+				doCommonChecks()
+				checkAtlasState()
+			})
+
+			By("Change cluster to GEOSHARDED", func() {
+				createdCluster.Spec.ClusterType = "GEOSHARDED"
+				createdCluster.Spec.ProviderSettings.RegionName = ""
+				createdCluster.Spec.ReplicationSpecs = []mdbv1.ReplicationSpec{
```
antonlisovenko, line 193: [q] does this deploy 3 nodes in east_us and 0 - in west_us?

antonlisovenko (approved):
Thanks for addressing the comments and happy to have good coverage of replicationSpecs!

There's another ticket for global clusters - I think we can test them later.

Well done 👍

test/int/cluster_test.go
vasilevp: You're right - should I change it to 2/1 instead of 3/0?

antonlisovenko: yes, 2/1 looks good to me
What do you think if we add 1 readonly/analytics node to some region?

PR #144, opened 2021-03-02
CLOUDP-82588: normalize K8s ids

antonlisovenko:
This PR prepares some groundwork for the next one that creates the Connection Secret.
As discussed with the team the name of the secrets will be:

`<project-name>-<cluster-name>-<db-user>` 

and also the secret will have the label for cluster name.

Taking the user input and using as K8s identifiers/label values is risky so some "normalization" work is done.

### DIFF
--- a/pkg/util/kube/kube.go
+++ b/pkg/util/kube/kube.go
@@ -1,10 +1,23 @@
 package kube
 
 import (
+	"math"
+	"regexp"
+	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L155
+var labelValueRegexp = regexp.MustCompile("[-a-z0-9._]")
+
+// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L177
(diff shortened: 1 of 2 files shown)


leo-ri (approved):

pkg/util/kube/kube.go
```diff
@@ -1,15 +1,82 @@
...
+	if errs := validation.IsValidLabelValue(name); len(errs) == 0 {
+		return name
+	}
+	return normalize(name, 63, labelValueRegexp)
+}
+
+// Dev note: the algorithm tries to replace the invalid characters with '-' (or simply omit it replacing is not possible)
+// Note, that this algorithm is not ideal - e.g. it won't fix the following: "a.#b" ("a._b" is still not a valid output - as
+// nonalphanumeric symbols cannot go together) though this doesn't intend to work in ALL the cases but in the MAJORITY instead
+func normalize(name string, limit int, regexp *regexp.Regexp) string {
```
vasilevp, line 53: Suggestion: maybe we can simplify this by utilizing regexp a bit deeper?

```go
var invalidStartEnd = regexp.MustCompile(`(^[^a-z0-9]+)|([^a-z0-9]+$)`)
var nonIdentifierRegexp = regexp.MustCompile(`[^a-z0-9]+`)

func normalize(name string, limit int, regexp regexp.Regexp) string {
	name = strings.ToLower(name)[:limit]
	name = invalidStartEnd.ReplaceAllString(name, "") // makes sure start & end are alphanumeric
	name = regexp.ReplaceAllString(name, "-") // replaces every sequence of invalid runes with a single "-"
	return name
}
```

vasilevp (changes requested):

vasilevp (approved):

PR #146, opened 2021-03-03
CLOUDP-82588: list secrets by cluster or username

antonlisovenko:
This PR depends on https://github.com/mongodb/mongodb-atlas-kubernetes/pull/145

This PR adds support for getting the connection secrets either by cluster name or by user name.

This will be necessary while cleaning the secrets on changes (mostly removals) of cluster/dbUser resources

### DIFF
--- a/pkg/controller/connectionsecret/listsecrets.go
+++ b/pkg/controller/connectionsecret/listsecrets.go
@@ -38,16 +38,12 @@ func list(k8sClient client.Client, namespace, projectID, clusterName, dbUserName
 			result = append(result, s)
 		}
 		if dbUserName != "" {
-			secretData := make(map[string]string)
-			for k, v := range s.Data {
-				secretData[k] = string(v)
-			}
-			var userName string
+			var userName []byte
 			var ok bool
-			if userName, ok = secretData[userNameKey]; !ok {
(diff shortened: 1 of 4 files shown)


pkg/controller/connectionsecret/listsecrets.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+		if _, ok := s.Labels[ClusterLabelKey]; !ok {
+			continue
+		}
+		if clusterName != "" && s.Labels[ClusterLabelKey] == kube.NormalizeLabelValue(clusterName) {
+			result = append(result, s)
+		}
+		if dbUserName != "" {
+			secretData := make(map[string]string)
+			for k, v := range s.Data {
```
vasilevp, line 42: [q] why do we need to create a copy of the map?

antonlisovenko: `s.Data` is a `map[string][]byte` - so we need to read each value and transform to string

antonlisovenko: but I see what you mean - we can just read the value and cast to string. Will do

vasilevp (approved):

PR #147, opened 2021-03-04
CLOUDP-82259: save logs as an artifact

leo-ri:
if tests fail - save logs as an artifact
added:
make e2e <focus=cluster-ns> can focus particular test
create_kind_cluster.sh - looking for default cluster 'kind' (because you can have several of them)


.github/workflows/test.yml
```diff
@@ -145,4 +145,9 @@ jobs:
         go get github.com/onsi/gomega/...
         ginkgo --focus "${TEST_NAME}" -nodes=3 -x test/e2e/
 
-# TODO if int test failed - stop e2e, add job for cleanup Atlas projects/clusters
+    - name: Upload operator logs
+      if: ${{ failure() }}
+      uses: actions/upload-artifact@v2
```
antonlisovenko, line 150: Looks like an amazing feature! 🥇 
Where does it save the files? Somewhere internally on Github and then shows in the UI?

antonlisovenko (approved):
Looks really great, I love the upload-artifact (though so far not sure what it does) :)

.github/workflows/test.yml
leo-ri: image
in the summary field

PR #149, opened 2021-03-05
CLOUDP-82115: Implement delete retry loop

vasilevp:


pkg/controller/atlascluster/atlascluster_controller.go
```diff
@@ -160,12 +161,22 @@ func (r *AtlasClusterReconciler) Delete(e event.DeleteEvent) error {
...
-	_, err = atlasClient.Clusters.Delete(context.Background(), project.Status.ID, cluster.Spec.Name)
-	if err != nil {
-		return fmt.Errorf("cannot delete Atlas cluster: %w", err)
-	}
+	go func() {
+		timeout := time.Now().Add(workflow.DefaultTimeout)
+
+		for time.Now().Before(timeout) {
+			_, err = atlasClient.Clusters.Delete(context.Background(), project.Status.ID, cluster.Spec.Name)
+			if err != nil {
```
antonlisovenko, line 169: what do you think if we check here the error that is returned in case the cluster is already removed - and break the loop in this case?

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -134,12 +135,23 @@ func (r *AtlasProjectReconciler) Delete(e event.DeleteEvent) error {
...
-	_, err = atlasClient.Projects.Delete(context.Background(), project.Status.ID)
-	if err != nil {
-		return fmt.Errorf("cannot delete Atlas project: %w", err)
-	}
+	go func() {
+		timeout := time.Now().Add(workflow.DefaultTimeout)
+
+		for time.Now().Before(timeout) {
+			_, err = atlasClient.Projects.Delete(context.Background(), project.Status.ID)
+			if err != nil {
```
antonlisovenko, line 143: the same as above

test/int/cluster_test.go
```diff
@@ -58,19 +58,12 @@ var _ = Describe("AtlasCluster", func() {
 			if createdCluster != nil {
 				By("Removing Atlas Cluster " + createdCluster.Name)
 				Expect(k8sClient.Delete(context.Background(), createdCluster)).To(Succeed())
-				Eventually(checkAtlasClusterRemoved(createdProject.Status.ID, createdCluster.Name), 600, interval).Should(BeTrue())
+				Eventually(checkAtlasClusterRemoved(createdProject.Status.ID, createdCluster.Spec.Name), 600, interval).Should(BeTrue())
```
antonlisovenko, line 61: 🥇

test/int/dbuser_test.go
```diff
@@ -91,20 +93,23 @@ var _ = Describe("AtlasDatabaseUser", func() {
...
 				By("Removing Atlas Cluster " + createdClusterAWS.Name)
 				Expect(k8sClient.Delete(context.Background(), createdClusterAWS)).To(Succeed())
-				Eventually(checkAtlasClusterRemoved(createdProject.Status.ID, createdClusterAWS.Name), 600, interval).Should(BeTrue())
+				Eventually(checkAtlasClusterRemoved(createdProject.Status.ID, createdClusterAWS.Spec.Name), 600, interval).Should(BeTrue())
 			}
 
 			By("Removing Atlas Project " + createdProject.Status.ID)
-			Eventually(removeAtlasProject(createdProject.Status.ID), 600, interval).Should(BeTrue())
+			Expect(k8sClient.Delete(context.Background(), createdProject)).To(Succeed())
+			Eventually(checkAtlasProjectRemoved(createdProject.Status.ID), 60, interval).Should(BeTrue())
```
antonlisovenko, line 112: [q] do you know if we still need `Eventually` here? If there are no clusters then the project is supposed to get removed immediately or am I missing something?

antonlisovenko (changes requested):
Just one suggestion to deeper check the error on removal.

Thanks for finding the bug with `Spec.Name`!!

pkg/controller/atlascluster/atlascluster_controller.go
vasilevp: Good catch!

test/int/dbuser_test.go
vasilevp: Some time could pass between a call to `k8sClient.Delete` and the actual API call. Also, we could always stumble upon a network issue and fail prematurely

antonlisovenko: ah, right, we remove the resource in K8s, not in Atlas

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -134,12 +137,29 @@ func (r *AtlasProjectReconciler) Delete(e event.DeleteEvent) error {
...
-	}
-
-	log.Infow("Successfully deleted Atlas project", "projectID", project.Status.ID)
+	go func() {
+		timeout := time.Now().Add(workflow.DefaultTimeout)
+
+		for time.Now().Before(timeout) {
+			_, err = atlasClient.Projects.Delete(context.Background(), project.Status.ID)
+			var apiError *mongodbatlas.ErrorResponse
+			if errors.As(err, &apiError) && apiError.ErrorCode == atlas.ClusterNotFound {
```
antonlisovenko, line 146: should it be `atlas.ProjectNotFound`?

PR #163, opened 2021-03-10
Split integration tests to different jobs

antonlisovenko:


.github/actions/int-test/entrypoint.sh
```diff
@@ -7,4 +7,4 @@ export ATLAS_PRIVATE_KEY="${INPUT_ATLAS_PRIVATE_KEY}"
 # otherwise we may get strange "Detected Programmatic Focus - setting exit status to 197"
 export GINKGO_EDITOR_INTEGRATION="true"
 
-ginkgo -v -p -nodes=4 ./test/int
\ No newline at end of file
+ginkgo --focus "${INPUT_TEST_NAME}" -v -p -nodes=4 ./test/int
```
leo-ri, line 10: not important:
 `-p -nodes=4` I think we can leave only `-nodes=4` since it is already 'run in parallel'

leo-ri (approved):

PR #167, opened 2021-03-15
CLOUDP-82260: add users

leo-ri:


antonlisovenko (commented):
Nothing serious - a couple of nits and corrections only

antonlisovenko (approved):
Looks great!

PR #169, opened 2021-03-15
CLOUDP-83105: db users wait for clusters

antonlisovenko:
The PR supports waiting for the clusters to propagate changes and marks the db users as ready only after that


test/int/dbuser_test.go
```diff
@@ -328,38 +324,36 @@ type Person struct {
...
-				panic(err)
-			}
-		}()
-
-		collection := dbClient.Database(dbName).Collection(collectionName)
-
-		p := Person{
-			Name: "Patrick",
-			Age:  32,
+func tryWrite(projectID string, cluster mdbv1.AtlasCluster, user mdbv1.AtlasDatabaseUser, dbName, collectionName string) error {
```
antonlisovenko, line 327: no changes here - just removed the wrapping `func()...`

leo-ri (approved):

vasilevp (approved):

PR #171, opened 2021-03-16
Ensure secrets on cluster update

vasilevp:


pkg/controller/atlascluster/cluster.go
```diff
@@ -130,3 +132,37 @@ func clustersEqual(log *zap.SugaredLogger, clusterA mongodbatlas.Cluster, cluste
...
+			}
+
+			continue out
+		}
+
+		data := connectionsecret.ConnectionData{
+			DBUserName: dbUser.Username,
+			ConnURL:    cluster.ConnectionStrings.Standard,
+			SrvConnURL: cluster.ConnectionStrings.StandardSrv,
+			Password:   dbUser.Password,
```
antonlisovenko, line 156: I'm not sure this will work - the password will be empty on GET requests

pkg/controller/atlascluster/cluster.go
```diff
@@ -130,3 +132,37 @@ func clustersEqual(log *zap.SugaredLogger, clusterA mongodbatlas.Cluster, cluste
 
 	return d == ""
 }
+
+func ensureConnectionSecrets(ctx context.Context, wctx *workflow.Context, k8sClient client.Client, project *mdbv1.AtlasProject, cluster *mongodbatlas.Cluster) workflow.Result {
+	dbUsers, _, err := wctx.Client.DatabaseUsers.List(ctx, project.ID(), &mongodbatlas.ListOptions{})
```
antonlisovenko, line 137: we need to query K8s database users as we know the passwords for them only - not for Atlas db users

antonlisovenko (changes requested):
My hypothesis that querying the Atlas DB users won't work because the password won't be returned.

This can be proved/disproved by int tests I believe.

pkg/controller/atlascluster/cluster.go
vasilevp: Ah, that makes sense... Guess this is going to be a bit more complicated then :slightly_smiling_face:

pkg/controller/atlascluster/cluster.go
```diff
@@ -139,3 +141,31 @@ func ClustersEqual(log *zap.SugaredLogger, clusterAtlas mongodbatlas.Cluster, cl
 
 	return d == ""
 }
+
+func ensureConnectionSecrets(wctx *workflow.Context, k8sClient client.Client, project *mdbv1.AtlasProject, cluster *mongodbatlas.Cluster) workflow.Result {
+	databaseUsers := mdbv1.AtlasDatabaseUserList{}
```
antonlisovenko, line 146: should this be empty?

pkg/controller/atlascluster/cluster.go
```diff
@@ -139,3 +141,31 @@ func ClustersEqual(log *zap.SugaredLogger, clusterAtlas mongodbatlas.Cluster, cl
...
+
+	for _, dbUser := range databaseUsers.Items {
+		found := false
+
+		for _, scope := range dbUser.GetScopes(mdbv1.ClusterScopeType) {
```
antonlisovenko, line 155: what if db user has empty scopes?

test/int/cluster_test.go
```diff
@@ -476,6 +478,42 @@ var _ = Describe("AtlasCluster", func() {
...
 	})
 
+	Describe("Create DBUser before cluster & check secrets", func() {
+		It("Should Succeed", func() {
+			By(fmt.Sprintf("Creating password Secret %s", UserPasswordSecret), func() {
+				passwordSecret := buildPasswordSecret(UserPasswordSecret, DBUserPassword)
+				Expect(k8sClient.Create(context.Background(), &passwordSecret)).To(Succeed())
+			})
+
+			createdDBUser := mdbv1.DefaultDBUser(namespace.Name, "test-db-user", createdProject.Name).WithPasswordSecret(UserPasswordSecret)
```
antonlisovenko, line 488: can you please extend the test case and create two users - one with empty scope and another - with non-empty and referencing the fake cluster. The later one will fail with validation which is ok. We need to make sure that the connection secret for it wasn't built.

pkg/controller/atlascluster/cluster.go
```diff
@@ -139,3 +141,31 @@ func ClustersEqual(log *zap.SugaredLogger, clusterAtlas mongodbatlas.Cluster, cl
...
+			}
+
+			continue
+		}
+
+		if !found {
+			continue
+		}
+
+		result := atlasdatabaseuser.CreateOrUpdateConnectionSecrets(wctx, k8sClient, *project, dbUser)
```
antonlisovenko, line 168: I think another condition should be that the user is "ready = true" what do you think? Otherwise if the user is not ready - this means that the connectivity to the cluster using this user may not work

antonlisovenko (changes requested):

antonlisovenko (approved):
👍  as soon as the tests are green

PR #172, opened 2021-03-16
CLOUDP-84205: support different order of operations

antonlisovenko:
This PR plays with a different order of resources creation - `AtlasDatabaseUser` first and `AtlasCluster` then. Also it checks what happens if the update to the user happens while the cluster is being created. The code was added to support this - the `AtlasDatabaseUser` gets into `pending` state if the cluster exists but doesn't have connection strings yet


pkg/controller/atlasdatabaseuser/connectionsecrets.go
```diff
@@ -13,22 +13,30 @@ import (
...
+func createOrUpdateConnectionSecrets(ctx *workflow.Context, k8sClient client.Client, project mdbv1.AtlasProject, dbUser mdbv1.AtlasDatabaseUser) workflow.Result {
 	clusters, _, err := ctx.Client.Clusters.List(context.Background(), project.ID(), &mongodbatlas.ListOptions{})
 	if err != nil {
-		// TODO CLOUDP-84205 ignore the 404 exception in case no clusters exist by this time
-		return err
+		return workflow.Terminate(workflow.DatabaseUserConnectionSecretsNotCreated, err.Error())
 	}
 
 	secretNames := make(map[string]string)
+	requeue := false
```
antonlisovenko, line 24: `requeue` flag is needed to send the reconciliation requeue signal in the end of the loop - to make sure the maximum number of secrets were ensured

test/int/dbuser_test.go
```diff
@@ -76,20 +77,6 @@ var _ = Describe("AtlasDatabaseUser", func() {
 			Eventually(testutil.WaitFor(k8sClient, createdProject, status.TrueCondition(status.ReadyType)),
 				20, interval).Should(BeTrue())
 		})
-
```
antonlisovenko, line 79: this block was moved from the top of the tests to the specific `Describe` that works with creating multiple users and clusters.

vasilevp (approved):

PR #174, opened 2021-03-18
CLOUDP-82709: watch password secrets

antonlisovenko:
Depends on https://github.com/mongodb/mongodb-atlas-kubernetes/pull/172

This PR adds the code to watch the database user passwords Secrets. As soon as those secrets are changed by the user - the db user reconciliation is triggered and the update request is sent to Atlas.

Note, that to support this we have to keep the state of the password secret in the database user status. This is necessary as the controller needs to know when the passwords have changed as current "compare db users specs" code won't give this information (Atlas doesn't return passwords for db users in GET requests).

### DIFF
--- a/pkg/api/v1/atlasdatabaseuser_types.go
+++ b/pkg/api/v1/atlasdatabaseuser_types.go
@@ -132,6 +132,14 @@ func (p AtlasDatabaseUser) AtlasProjectObjectKey() client.ObjectKey {
 	return kube.ObjectKey(p.Namespace, p.Spec.Project.Name)
 }
 
+func (p AtlasDatabaseUser) PasswordSecretObjectKey() *client.ObjectKey {
+	if p.Spec.PasswordSecret != nil {
+		key := kube.ObjectKey(p.Namespace, p.Spec.PasswordSecret.Name)
+		return &key
+	}
+	return nil
+}
+
 func (p *AtlasDatabaseUser) GetStatus() status.Status {
 	return p.Status
 }
@@ -153,9 +161,13 @@ func (p *AtlasDatabaseUser) ReadPassword(kubeClient client.Client) (string, erro
 		if err := kubeClient.Get(context.Background(), kube.ObjectKey(p.Namespace, p.Spec.PasswordSecret.Name), secret); err != nil {
 			return "", err
 		}
-		if p, exist := secret.Data["password"]; !exist {
+		p, exist := secret.Data["password"]
+		switch {
+		case !exist:
 			return "", fmt.Errorf("secret %s is invalid: it doesn't contain 'password' field", secret.Name)
-		} else {
+		case len(p) == 0:
+			return "", fmt.Errorf("secret %s is invalid: the 'password' field is empty", secret.Name)
+		default:
 			return string(p), nil
 		}
 	}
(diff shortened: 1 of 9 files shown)


pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -150,12 +158,16 @@ func (p *AtlasDatabaseUser) UpdateStatus(conditions []status.Condition, options
 func (p *AtlasDatabaseUser) ReadPassword(kubeClient client.Client) (string, error) {
 	if p.Spec.PasswordSecret != nil {
 		secret := &corev1.Secret{}
-		if err := kubeClient.Get(context.Background(), kube.ObjectKey(p.Namespace, p.Spec.PasswordSecret.Name), secret); err != nil {
+		if err := kubeClient.Get(context.Background(), *p.PasswordSecretObjectKey(), secret); err != nil {
 			return "", err
 		}
-		if p, exist := secret.Data["password"]; !exist {
+		p, exist := secret.Data["password"]
```
antonlisovenko, line 164: `switch` was requested by the linter

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -72,8 +72,9 @@ func (r *AtlasProjectReconciler) Reconcile(context context.Context, req ctrl.Req
 		return result.ReconcileResult(), nil
 	}
 	if project.ConnectionSecretObjectKey() != nil {
+		// Note, that we are not watching the global connection secret - seems there is no point in reconciling all
```
vasilevp, line 75: [q] what if the `orgId` was changed? do we even support that case?

vasilevp (approved):

pkg/controller/atlasproject/atlasproject_controller.go
antonlisovenko: yes - in this case the new project will be created in that organization and the stale project will stay in the old organization until it's removed manually

vasilevp: what I mean is - don't we need to reconcile in this case?

antonlisovenko: this is a good thing to discuss!
If say the customer has 5 projects in the organization each with 2 clusters. What would be the motivation to change the  `orgID` and to copy all the projects (and eventually all the clusters) to the new organization? If they want the organization with a new name - then the name of organization can be edited directly in Atlas. Apart from that  automatic creation of 5 projects in the new organization (without automatic creation of clusters though) right after the global API key changes doesn't look very user-friendly to me... If we don't create the new project right away - the old one (with the same name and IP configs) with all the clusters will still be alive

That being said even watching the "project-level" secret doesn't look very important though it will affect smaller number of projects unlike the global one which may create a bunch of projects right away. So maybe we don't need to watch the project-level secret at all as well...

What do you think? Do you think we should watch the global secret as well?

vasilevp: I agree that this is probably not something the user would want to do, so this is more of a curious question than it is a feature request

So yeah, most likely we don't need to change this, just wanted to hear your thoughts on it 🙂

PR #177, opened 2021-03-19
Delete secrets on cluster removal

vasilevp:


test/int/cluster_test.go
```diff
@@ -392,6 +392,38 @@ var _ = Describe("AtlasCluster", func() {
...
+
+		passwordSecret := buildPasswordSecret(UserPasswordSecret, DBUserPassword)
+		Expect(k8sClient.Create(context.Background(), &passwordSecret)).To(Succeed())
+
+		createdDBUser := mdbv1.DefaultDBUser(namespace.Name, "test-db-user", createdProject.Name).WithPasswordSecret(UserPasswordSecret)
+		By(fmt.Sprintf("Creating the Database User %s", kube.ObjectKeyFromObject(createdDBUser)), func() {
+			Expect(k8sClient.Create(context.Background(), createdDBUser)).ToNot(HaveOccurred())
+
+			Eventually(testutil.WaitFor(k8sClient, createdDBUser, status.TrueCondition(status.ReadyType)),
+				20, interval).Should(BeTrue())
```
antonlisovenko, line 416: in my experience 20 seconds is often not enough, would recommend something around 80

test/int/cluster_test.go
```diff
@@ -392,6 +392,38 @@ var _ = Describe("AtlasCluster", func() {
...
+		})
+
+		By("Removing Atlas Cluster "+createdCluster.Name, func() {
+			Expect(k8sClient.Delete(context.Background(), createdCluster)).To(Succeed())
+			Eventually(checkAtlasClusterRemoved(createdProject.Status.ID, createdCluster.Spec.Name), 600, interval).Should(BeTrue())
+
+			secretNames := []string{kube.NormalizeIdentifier(fmt.Sprintf("%s-%s-%s", createdProject.Spec.Name, createdCluster.Spec.Name, createdDBUser.Spec.Username))}
+			Eventually(checkSecretsDontExist(namespace.Name, secretNames), 50, interval).Should(BeTrue())
```
antonlisovenko, line 424: just in case would suggest to check the total number of secrets as well:
```
checkNumberOfConnectionSecrets(k8sClient, *createdProject, 0)
```

antonlisovenko (approved):
LGTM after the tests get green

test/int/cluster_test.go
vasilevp: Thanks, that makes sense

PR #178, opened 2021-03-22
CLOUDP-84905: support user expiration

antonlisovenko:
This PR adds support for database user expiration. If the `spec.DeleteAfterDate` is less than the current date - no API requests are sent to Atlas, the resource is marked as `ready: false` and `DatabaseUserReadyType: false` and connection secrets are removed. Note, that the retry of the reconciliation doesn't happen in this case as it's the user that needs to update the spec to fix this


pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -171,24 +170,8 @@ func (r AtlasDatabaseUserReconciler) Delete(e event.DeleteEvent) error {
...
-			log.Errorf("Failed to remove connection Secret: %v", err)
-		} else {
-			log.Debugw("Removed connection Secret", "secret", kube.ObjectKeyFromObject(s))
-		}
-	}
-	if len(secrets) > 0 {
-		log.Infof("Removed %d connection secrets", len(secrets))
-	}
+	// We ignore the error as it will be printed by the function
+	_ = removeStaleSecretsByUserName(r.Client, project.ID(), userName, *dbUser, log)
```
antonlisovenko, line 174: moved to a single function as this code is reused by different places

pkg/controller/atlasdatabaseuser/databaseuser.go
```diff
@@ -180,6 +214,21 @@ func userMatchesSpec(log *zap.SugaredLogger, atlasSpec *mongodbatlas.DatabaseUse
 		return false, err
 	}
 
+	// performing some normalization of dates
```
antonlisovenko, line 217: this is necessary to correctly compare strings as the date returned by Atlas may be different than the one submitted by the Atlas Operator

vasilevp (approved):

PR #182, opened 2021-03-23
CLOUDP-84606: helm support

antonlisovenko:
This PR makes the changes to align the configuration with the Helm chart (https://github.com/mongodb/helm-charts/pull/25):
* adds support for the service account
* increases the limits resources
* doesn't create the CluterRole `metrics-reader` as it's not required by the Operator
* renames `WATCHED_NAMESPACE` to `WATCH_NAMESPACE` as the later is a standard naming in the other operators

### DIFF
--- a/config/rbac/clusterwide/auth_proxy_role_binding.yaml
+++ b/config/rbac/clusterwide/auth_proxy_role_binding.yaml
@@ -8,5 +8,5 @@ roleRef:
   name: proxy-role
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: mongodb-atlas-operator
   namespace: system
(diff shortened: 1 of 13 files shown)


config/rbac/clusterwide/auth_proxy_role_binding.yaml
```diff
@@ -8,5 +8,4 @@ roleRef:
   name: proxy-role
 subjects:
 - kind: ServiceAccount
-  name: default
-  namespace: system
+  name: operator
```
antonlisovenko, line 11: kustomize will replace this with `mongodb-atlas-operator` as it adds prefix to all the names

leo-ri (approved):

PR #185, opened 2021-03-25
CLOUDP-83733: more tests on clusters (create only)

antonlisovenko:
This PR focuses on more complicated testing of the clusters. Due to the size only the creation is happening - the update will be addressed in the next PR (found some issues there).

The changes were reviewed and agreed on with Pavel p2p.

* fixed the situation when the user specifies the `providerSettings.regionName` together with `replicationSpecs`. In this case Atlas accepts the update request but erazes the former field so the infinite reconciliation happens as `ClustersEqual` method constantly returns false.
* added more checks to the integration test to check all the fields
* more unit tests to make sure the `MergedCluster` and `ClustersEqual` work correctly
* increased the timeout for project create/update as 20 seconds is not enough often

### DIFF
--- a/pkg/controller/atlascluster/cluster.go
+++ b/pkg/controller/atlascluster/cluster.go
@@ -15,42 +15,42 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/compat"
 )
 
-func (r *AtlasClusterReconciler) ensureClusterState(ctx *workflow.Context, project *mdbv1.AtlasProject, cluster *mdbv1.AtlasCluster) (c *mongodbatlas.Cluster, _ workflow.Result) {
-	c, resp, err := ctx.Client.Clusters.Get(context.Background(), project.Status.ID, cluster.Spec.Name)
+func (r *AtlasClusterReconciler) ensureClusterState(ctx *workflow.Context, project *mdbv1.AtlasProject, cluster *mdbv1.AtlasCluster) (atlasCluster *mongodbatlas.Cluster, _ workflow.Result) {
+	atlasCluster, resp, err := ctx.Client.Clusters.Get(context.Background(), project.Status.ID, cluster.Spec.Name)
 	if err != nil {
 		if resp == nil {
-			return c, workflow.Terminate(workflow.Internal, err.Error())
+			return atlasCluster, workflow.Terminate(workflow.Internal, err.Error())
 		}
 
 		if resp.StatusCode != http.StatusNotFound {
-			return c, workflow.Terminate(workflow.ClusterNotCreatedInAtlas, err.Error())
+			return atlasCluster, workflow.Terminate(workflow.ClusterNotCreatedInAtlas, err.Error())
 		}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/atlascluster/cluster.go
```diff
@@ -15,42 +15,42 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/compat"
 )
 
-func (r *AtlasClusterReconciler) ensureClusterState(ctx *workflow.Context, project *mdbv1.AtlasProject, cluster *mdbv1.AtlasCluster) (c *mongodbatlas.Cluster, _ workflow.Result) {
-	c, resp, err := ctx.Client.Clusters.Get(context.Background(), project.Status.ID, cluster.Spec.Name)
+func (r *AtlasClusterReconciler) ensureClusterState(ctx *workflow.Context, project *mdbv1.AtlasProject, cluster *mdbv1.AtlasCluster) (atlasCluster *mongodbatlas.Cluster, _ workflow.Result) {
+	atlasCluster, resp, err := ctx.Client.Clusters.Get(context.Background(), project.Status.ID, cluster.Spec.Name)
```
antonlisovenko, line 19: did some renamings of variables as it was a bit hard to follow which cluster is from Atlas and which - from the Operator later when we did merging magic

pkg/controller/atlascluster/cluster.go
```diff
@@ -99,9 +99,9 @@ func cleanupCluster(cluster mongodbatlas.Cluster) mongodbatlas.Cluster {
 	return cluster
 }
 
-// mergedCluster will return the result of merging AtlasClusterSpec with Atlas Cluster
-func mergedCluster(cluster mongodbatlas.Cluster, spec mdbv1.AtlasClusterSpec) (result mongodbatlas.Cluster, err error) {
-	if err = compat.JSONCopy(&result, cluster); err != nil {
+// MergedCluster will return the result of merging AtlasClusterSpec with Atlas Cluster
+func MergedCluster(atlasCluster mongodbatlas.Cluster, spec mdbv1.AtlasClusterSpec) (result mongodbatlas.Cluster, err error) {
```
antonlisovenko, line 103: made public to use inside integration tests

test/int/cluster_test.go
```diff
@@ -90,14 +109,10 @@ var _ = Describe("AtlasCluster", func() {
 			atlasCluster, _, err := atlasClient.Clusters.Get(context.Background(), createdProject.Status.ID, createdCluster.Spec.Name)
 			Expect(err).ToNot(HaveOccurred())
 
-			createdAtlasCluster, err := createdCluster.Spec.Cluster()
+			mergedCluster, err := atlascluster.MergedCluster(*atlasCluster, createdCluster.Spec)
```
antonlisovenko, line 112: using the `MergedCluster` to compare clusters in total.
This could be error-prone in case if `MergedCluster` has bugs in it - that's why the unit tests above were extended

vasilevp (approved):

PR #186, opened 2021-03-29
CLOUDP-84848-bundle

leo-ri:


test/e2e/bundle_test.go
```diff
@@ -0,0 +1,128 @@
...
+		By("Wait cluster creation", func() {
+			waitCluster(userSpec, "1")
+		})
+
+		By("Check attributes", func() {
+			uCluster := mongocli.GetClustersInfo(userSpec.ProjectID, userSpec.Clusters[0].Spec.Name)
+			compareClustersSpec(userSpec.Clusters[0].Spec, uCluster)
+		})
+
+		By("check database users Attibutes", func() {
```
antonlisovenko, line 105: Can I suggest checking the test app and connectivity to the atlas cluster here? They should work for sure as in the other test but it will be good to ensure that again.

antonlisovenko (approved):
This is amazing - now we can make changes to CSVs and be sure we haven't broken anything!

The only suggestion is to extend the tests and check the connectivity to the cluster.

Doesn't need a second look from me!

PR #192, opened 2021-03-31
Replace GCP with AWS/Azure (almost) everywhere

vasilevp:


Makefile
```diff
@@ -110,16 +110,14 @@ generate: controller-gen ## Generate code
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: controller-gen
-# Download controller-gen locally if necessary
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
-controller-gen:
+controller-gen: ## Download controller-gen locally if necessary
```
vasilevp, line 114: moved the comment so these targets appear in `make help`

antonlisovenko (approved):
LGTM as soon as the tests are green 👍

antonlisovenko:
@vasilevp seems like one test failed due to timeout.. could be the problem in Atlas or just some glitch. I'm suggesting to increase the timeout a bit, but not too much - 40/45 min maybe?

PR #196, opened 2021-04-05
CLOUDP-82258: unstuck e2e

leo-ri:
- tests: remove unnecessary `eventually` waiting
- ci/cd: more nodes + stream (probably better to leave it like that for a while)
- tests: fix ports in the test
- test-app: update test-app(add logs) & create pipeline for auto building
- ci/cd: delete projects older than 1 day at the end


.github/workflows/build-test-app.yml
```diff
@@ -0,0 +1,29 @@
+name: Build test application and push it to docker repository
+
+on:
+  push:
+    paths:
```
antonlisovenko, line 5: wow, nice! Good to know 👍

scripts/e2e_local.sh
```diff
@@ -28,4 +28,4 @@ export MCLI_OPS_MANAGER_URL="https://cloud-qa.mongodb.com/"
 export MCLI_PUBLIC_API_KEY="${public_key}"
 export MCLI_PRIVATE_API_KEY="${private_key}"
 export MCLI_ORG_ID="${org_id}"
-ginkgo --focus "${focus_key}" -nodes=3 -x -v test/e2e/
+ginkgo -focus "${focus_key}" -stream -nodes=6 -x -v test/e2e/
```
antonlisovenko, line 31: [q] what does `stream` do?

test/e2e/operator_type_wide_test.go
```diff
@@ -51,9 +52,8 @@ var _ = Describe("[cluster-wide] Users (Norton and Nimnul) can work with one Clu
 					NimnulSpec.Namespace,
 				)
 			} else {
-				kube.Delete(NortonSpec.Clusters[0].ClusterFileName(NortonSpec), "-n", NortonSpec.Namespace)
-				kube.Delete(NimnulSpec.Clusters[0].ClusterFileName(NimnulSpec), "-n", NimnulSpec.Namespace)
-				// do not wait it
+				Eventually(kube.DeleteNamespace(NortonSpec.Namespace)).Should(Say("deleted"), "Cant delete namespace after testing")
```
antonlisovenko, line 55: 👍

test/e2e/user_input_test.go
```diff
@@ -19,26 +19,27 @@ import (
...
+		1, kube.GetGeneration(input.Namespace, input.Clusters[0].GetClusterNameResource()),
+	).Should(Equal(generation))
+	ExpectWithOffset(1, kube.GetK8sClusterStateName(
 		input.Namespace, input.Clusters[0].GetClusterNameResource()),
-		"45m", "1m",
 	).Should(Equal("IDLE"), "Kubernetes resource: Cluster status should be IDLE")
 
-	Expect(
-		mongocli.GetClusterStateName(input.ProjectID, input.Clusters[0].Spec.Name),
+	ExpectWithOffset(
```
antonlisovenko, line 34: [q] what does `ExpectWithOffset` do?

antonlisovenko (approved):
Amazing work, thanks for addressing the cleanup! 🥇 
just a couple of questions

scripts/e2e_local.sh
leo-ri: it will enumerate each node test and print it as is - node output will be mixed (for configuration test only). So I was thinking to leave it like this and see if there are any more failures

test/e2e/user_input_test.go
leo-ri: if we have test asserts in another package, and the test fails on this assert, it shows another package code line, and we usually want to know the test code line

antonlisovenko: oh I see - very useful feature, will use in integration tests as well, thanks!

PR #213, opened 2021-04-20
CLOUDP-82912: annotations to avoid removing resources from Atlas

antonlisovenko:
This PR adds the support for annotations of type `mongodb.com/atlas-resource-policy="keep"`
When set to any CR it will tell the AO to skip removal of the resource in Atlas when the K8s resource is removed.
Note, that the connection secrets are still removed in this case


test/int/project_test.go
```diff
@@ -124,6 +126,25 @@ var _ = Describe("AtlasProject", func() {
...
+				Expect(k8sClient.Create(context.Background(), createdProject)).ToNot(HaveOccurred())
+
+				Eventually(testutil.WaitFor(k8sClient, createdProject, status.TrueCondition(status.ReadyType)),
+					ProjectCreationTimeout, interval).Should(BeTrue())
+			})
+			By("Deleting the project", func() {
+				Expect(k8sClient.Delete(context.Background(), createdProject)).To(Succeed())
+				time.Sleep(10 * time.Second)
+				Expect(checkAtlasProjectRemoved(createdProject.Status.ID)()).Should(BeFalse())
+				createdProject = nil
```
vasilevp, line 143: [q] shouldn't we remove this project from Atlas to keep it cleaner?

antonlisovenko: I thought the project cleaner will do it but better do manually, right 👍

leo-ri (approved):

PR #214, opened 2021-04-21
Update yq to version 4

fabritsius:
The latest V4 version of `yq` has new syntax,
this small change fixes issues caused by the migration.

Along with this change local `yq` should be updated to the latest release.
This can be done via brew: `brew upgrade yq` (https://pypi.org/project/yq/)

P.S. there is a solution which doesn't involve the update. The use of V3 is not recommended but possible.
In case V3 syntax is left as it is I suggest to update `start-guide.md` with: `brew install yq@3`


antonlisovenko (approved):
Nice fix, thanks @fabritsius ! 🥇

PR #221, opened 2021-04-29
Fix "make deploy"

antonlisovenko:
Our `make deploy` is broken - for some reasons `act -j ..` doesn't work for local custom actions.
Changed this to a script directly - this is anyway faster than calling `act`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,8 +87,8 @@ uninstall: manifests kustomize ## Uninstall CRDs from a cluster
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
 .PHONY: deploy
-deploy: run-kind ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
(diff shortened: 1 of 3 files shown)


leo-ri (approved):

Makefile
```diff
@@ -87,8 +87,8 @@ uninstall: manifests kustomize ## Uninstall CRDs from a cluster
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
 .PHONY: deploy
-deploy: run-kind ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-	@ act -j deploy -s KUBE_CONFIG_DATA='$(shell kind get kubeconfig | yq eval -j )'
```
fabritsius, line 91: This is the same line I changed with my yq V4 commit so maybe that's the reason it started failing.
`yq eval` won't work with yq version 3 which in turn will spoil act -j command.

In any case now it's better =)

fabritsius (commented):
Great job 👍

PR #240, opened 2021-05-25
CLOUDP-88737: Do not try to delete non-existing Atlas projects

fabritsius:
I've added an extra check in the Delete method for Projects.

- I thought it makes sense to have this new check right after the first function comment
(so it is clear that that's a part of delete process)
- No error is thrown

### DIFF
--- a/pkg/controller/atlasproject/atlasproject_controller.go
+++ b/pkg/controller/atlasproject/atlasproject_controller.go
@@ -134,7 +134,7 @@ func (r *AtlasProjectReconciler) Delete(e event.DeleteEvent) error {
 	log.Infow("-> Starting AtlasProject deletion", "spec", project.Spec)
 
 	if project.Status.ID == "" {
-		log.Infof("Project does not exist, nothing to remove")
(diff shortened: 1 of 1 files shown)


pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -133,6 +133,11 @@ func (r *AtlasProjectReconciler) Delete(e event.DeleteEvent) error {
 	log := r.Log.With("atlasproject", kube.ObjectKeyFromObject(project))
 	log.Infow("-> Starting AtlasProject deletion", "spec", project.Spec)
 
+	if project.Status.ID == "" {
+		log.Infof("Project does not exist, nothing to remove")
```
antonlisovenko, line 137: [nit] maybe we could add "does not exist in Atlas"?

antonlisovenko (approved):
Looks great! 👍

pkg/controller/atlasproject/atlasproject_controller.go
fabritsius: I agree, better to be concrete 👌

PR #248, opened 2021-06-03
CLOUDP-90594: CRD for private link

antonlisovenko:
This follows the specification described in the Spec for the PrivateLink


fabritsius (approved):
Great job 👍

We will have to add more fields in the future but I think this is enough for now.

pkg/api/v1/provider/provider.go
```diff
@@ -0,0 +1,10 @@
+package provider
+
+type ProviderName string
+
+const (
+	ProviderAWS    ProviderName = "AWS"
+	ProviderGCP    ProviderName = "GCP"
+	ProviderAzure  ProviderName = "AZURE"
+	ProviderTenant ProviderName = "TENANT"
+)
```
fabritsius, line 10: Here the formatter just wants an extra new line at the end.

PR #273, opened 2021-07-08
CLOUDP-92613 flaky

leo-ri:
probably fixed geosharded (if it is because long ginkgo await)
change openshift test order (lock earlier)
update kind action (global e2e fails)


fabritsius (approved):
Great Job 👍

Issue #270, opened 2021-07-06
Cross Namespace Watching Does Not Work for Global API Key


Freyert:
1. Deploy operator into namespace : `mongodb`
2. set environment variable WATCH_NAMESPACE: `ns2`
3. receive error `reason":"AtlasCredentialsNotProvided","message":"Secret \"mongodb-atlas-operator-api-key\" not found"`

I used the debugger to discover that the cache for the manager's kubernetes client uses the `WATCH_NAMESPACE` to populate its cache of `Secrets`. Therefore the cache will not have `mongodb/mongodb-atlas-operator-api-key`.

This is the documented behavior of the `Namespace` option for controller mangers:
https://github.com/kubernetes-sigs/controller-runtime/blob/ef5c8a3ffd288667b89005aebad5ff6e0cf0c5e4/pkg/manager/manager.go#L188-L194

> Namespace if specified restricts the manager's cache to watch objects in the desired namespace Defaults to all namespaces

>Note: If a namespace is specified, controllers can still Watch for a cluster-scoped resource (e.g Node). For namespaced resources the cache will only hold objects from the desired namespace.

So for the GlobalAPIKey secret to be usable I think we need to be able to mount the secret as a volume or otherwise configure a different kubernetes client for retrieving the GlobalAPIKey secret.

Please let me know if there's a workaround/configuration I'm missing! Thank you for your help.

Freyert:
Related to #215

jamesbroadhead:
Fixed in #282

PR #277, opened 2021-07-16
Refactor to use Builder package and filter namespaces with Predicates

Freyert:
I was going to use a [`MultiNamespaceCache`] to enable cross namespace watching, but after looking into it more there are some performance warnings when using a large number of namespaces and a preference for using [`Predicates`] for including/excluding namespaces.

Swapping to the builder pattern aligns with the expected use of `controller-runtime` from what I can tell and makes it very easy to leverage [`Predicates`], model `parent-child` references, and other commons tasks besides.

Since this may be a somewhat significant change I did a review of features that may be better implemented using the common flow of [`Predicates`]:

1. https://github.com/mongodb/mongodb-atlas-kubernetes/pull/272 could be done by adding a predicate to disable listening.
2. https://github.com/mongodb/mongodb-atlas-kubernetes/pull/213/files added +200 lines to the code base, but may have been implemented with a [`Predicate`] based on `Delete` requests that ignores the request if the label exists.
3. https://github.com/mongodb/mongodb-atlas-kubernetes/issues/270 works fine if you watch all namespaces, but if you want to restrict namespaces it doesn't because you can't find the Global API Key. This PR fixes that issue using predicates.

Using the controller [`Builder`] package also connects you to other facets of `controller-runtime` that you may be interested in pursuing. For instance you can define `Owns` relationships between `AtlasClusters` and `ConnectionSecrets` so that the controller knows to garbage collect `ConnectionSecrets` if the owner disappears. ([Docs](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/builder#Builder.Owns)) Or just separating the control logic for `ConnectionSecrets` into their own controller.

I hope this is helpful! Thank you for letting me review your project and let me know if there was a reason for the previous code and I can close this PR. If you can't tell we're very excited about your work :)

[`Predicate`]: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/predicate
[`Predicates`]: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/predicate
[`MultiNamespaceCache`]: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/cache#MultiNamespacedCacheBuilder
[`Builder`]: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"log"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/go-logr/zapr"
@@ -29,6 +30,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	ctrzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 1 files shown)


Freyert:
Documented reasons to not use MultiNamespaceCache:
> Note that this is not intended to be used for excluding namespaces, this is better done via a Predicate. Also note that you may face performance issues when using this with a high number of namespaces.

Freyert:
I see [`SetupWithManager`](https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/pkg/controller/atlasproject/atlasproject_controller.go#L183-L201) sets up the watches etc.

PR #281, opened 2021-07-20
CLOUDP-92613 flaky (part 2: polling intervals)

leo-ri:
- add/split polling intervals, add defaults
- add new trigger for tests


fabritsius (approved):
LGTM 👍

PR #278, opened 2021-07-16
Controller Namespace Predicate

Freyert:
After reviewing, https://github.com/mongodb/mongodb-atlas-kubernetes/pull/277 I began to understand how controllers are currently configured in `mongodb-atlas-kubernetes` operator. This is a redo of https://github.com/mongodb/mongodb-atlas-kubernetes/pull/277 to more seamlessly incorporate a namespace predicate into controllers.

Fixes: https://github.com/mongodb/mongodb-atlas-kubernetes/issues/270

PR #282, opened 2021-07-21
CLOUDP-94516: Testing of Controller Namespace Predicate Changes

fabritsius:
This a copy of #278 PR to allow the tests to run.


Freyert:
These failures seem hard to tie directly to the changes I've made. Could these be flakes or test infrastructure issues?

PR #298, opened 2021-08-02
Configurable integration domain

Freyert:
I wanted to be able to run the integration tests and needed to reconfigure the Atlas domain being used to do so.

users can `export ATLAS_DOMAIN="https://cloud.mongodb.com"` to run the tests using the public Atlas platform.

Perhaps there are other places to modify, or a better way? Currently I'm hooking into the `package` `init` function to set the variable via an env var or use a default value.

Also, what's the best way to speed these tests up? Should I throw more ginkgo threads at them?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,8 @@ require (
 	go.mongodb.org/atlas v0.7.3-0.20210315115044-4b1d3f428c24
 	go.mongodb.org/mongo-driver v1.7.0
 	go.uber.org/zap v1.18.1
+	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
+	golang.org/x/tools v0.1.5 // indirect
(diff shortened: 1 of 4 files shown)


Freyert:
I imagine these tests shouldn't run if I'm not part of the MongoDB organization. Seems like the values necessary to kick off the builds don't exist.

leo-ri:
Thank you for your contribution

> Also, what's the best way to speed these tests up? Should I throw more ginkgo threads at them?

In the current state threads could help (it creates real clusters = takes a long time), another way - rewrite tests with mocking where possible

> I imagine these tests shouldn't run if I'm not part of the MongoDB organization. Seems like the values necessary to kick off the builds don't exist.

Forked repos do not have access. Another trigger has been added for that (old created branches without update will still run with both triggers, and some could fail)

PR #310, opened 2021-08-16
CLOUDP-97762: Fix Dependabot error, remove Anton from reviewers

jamesbroadhead:
Dependabot was failing to assign reviewers, as Anton is no longer on the
team

Remove Anton's username from the OLM readme also

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,6 @@ updates:
       interval: weekly
       day: tuesday
     reviewers:
-      - "antonlisovenko"
       - "fabritsius"
       - "leo-ri"
   - package-ecosystem: github-actions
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -6,7 +6,6 @@ updates:
       interval: weekly
       day: tuesday
     reviewers:
-      - "antonlisovenko"
```
gssbzn, line 9: [q] have you thought about having a group instead, see https://github.com/mongodb/mongocli/blob/aba18a22fa124537e22241a3d56db95358e554e7/.github/dependabot.yml#L9

jamesbroadhead: that's a good idea

priyolahiri (approved):
LGTM

PR #324, opened 2021-09-28
Cloudp-101523: Add cluster termination to cleaner, and manual workflow for it, change helm-update test

leo-ri:
- Add cluster termination to cleaner
- Add manual workflow for running cleaner
- Add case to helm-update with a different version of charts values file


fabritsius (approved):
Looks great 👍

fabritsius (approved):

PR #331, opened 2021-10-08
CLOUDP-88816 second deletion test

leo-ri:
please do not merge until the bug fix
UPD: expecting that the problem/bug is gone, so need to merge it for running whose tests  for a while


leo-ri:
Test Forked / E2E tests (latest-openshift, openshift) is not available for now

fabritsius (approved):

PR #344, opened 2021-11-17
Adding private connection string counterparts.

jhill072:
Adding the two private connection strings to the generated secret, since we only expose the peering IPs through the firewall.

Please let me know if I need to provide anything further, next steps, etc. as this is my first open source PR!


jhill072:
Just to add further context, our company is using the private peering option in Atlas, which means we are able to use our private IPs during communications. Thus, we have also only allowed those IPs through to the Atlas firewall. However, when using the connection strings currently being produced by the Atlas operator, they are forced to go through the public internet, losing their private IPs.

Luckily, the Atlas API already returns the private connection strings alongside the standard ones, and they were already typed in the project. I am attempting to simply add those as two new keys to the generated secret, if they exist (since not everyone will enable peering)

jamesbroadhead:
@jhill072 Thank you for this PR! 
Could you sign our [CLA](https://www.mongodb.com/legal/contributor-agreement) please?  
That will enable us to merge your patch

jhill072:
Sorry about that @jamesbroadhead! I have just signed the agreement, so if there is anything else I need to do to get this merged let me know.

Thank you!

pkg/controller/connectionsecret/ensuresecret.go
```diff
@@ -50,20 +52,32 @@ func Ensure(client client.Client, namespace, projectName, projectID, clusterName
...
+	var connURL, srvConnURL, pvtConnURL, pvtSrvConnURL string
 	var err error
 	if connURL, err = AddCredentialsToConnectionURL(data.ConnURL, data.DBUserName, data.Password); err != nil {
 		return err
 	}
 	if srvConnURL, err = AddCredentialsToConnectionURL(data.SrvConnURL, data.DBUserName, data.Password); err != nil {
 		return err
 	}
+	if data.PvtConnURL == "" {
+		pvtConnURL = ""
```
fabritsius, line 64: In both cases strings are empty by default. We can check the opposite and add credentials only if they are not empty:
```
if data.PvtConnURL != "" {
    if pvtConnURL, err = AddCredentialsToConnectionURL(data.PvtConnURL, data.DBUserName, data.Password); err != nil {
        return err
    }
}
```

jhill072: I like it! Haven't had the pleasure of truly jumping into golang yet so still learning some of those basics. Thank you for the suggestion!

jhill072:
I am attempting to get the E2E tests running locally, but I wanted to ask if this was something flaky? The error logs are saying the dependent helm chart isn't where it is supposed to be. I'm not sure how my changes could have affected something like that, so just wanted to throw feelers out there as to what may be going wrong here while I try to reproduce.

leo-ri:
@jhill072  I will take a look

leo-ri:
@jhill072 could you please update your branch with the changes from `main`?

jhill072:
@leo-ri main has been fetched, thank you for the update!

leo-ri (approved):

leo-ri:
@jhill072 "Merging is blocked. The base branch requires all commits to be signed." could you please sign your commits?

jamesbroadhead:
@jhill072  ^ no need - thank you for the contribution!

jhill072:
Thank you all for the support during the process as well! This has been a
fantastic first PR experience! 😀

PR #352, opened 2021-12-01
CLOUDP-106704: e2e helm-test: dependency as file

leo-ri:
- changed paths : absolute >> relative
- work with dependency: delete, install separately >> change the source as a file
- remove unsetting for backingProvider from the test (the chart can manage it now without creating a separate value file)


fabritsius (approved):
Looks great 👍

fabritsius (approved):

PR #355, opened 2021-12-03
Add private endpoint for AWS

fabritsius:


pkg/controller/atlasproject/private_endpoint.go
```diff
@@ -0,0 +1,250 @@
...
+
+	peStatus := convertToStatus(atlasPEs)
+	ctx.Log.Debugw("Private Endpoints Status", "status", peStatus)
+
+	ctx.EnsureStatusOption(status.AtlasProjectPrivateEnpointOption(peStatus))
+	return workflow.OK()
+}
+
+func createOrDeletePEInAtlas(ctx *workflow.Context, projectID string, operatorPrivateEndpoints []project.PrivateEndpoint, log *zap.SugaredLogger) (result workflow.Result, atlasPrivateEndpoints []AtlasProjectPrivateEndpoint) {
+	provider := "AWS"
```
jasonmimick, line 34: Why is this hardcoded?

pkg/controller/atlasproject/private_endpoint.go
```diff
@@ -0,0 +1,250 @@
...
+
+			result = workflow.InProgress(workflow.ProjectPEServiceIsNotReadyInAtlas, "Private Endpoint is creating")
+		}
+	}
+
+	return result
+}
+
+func DeleteAllPrivateEndpoints(client mongodbatlas.Client, projectID string, log *zap.SugaredLogger) error {
+	provider := "AWS"
```
jasonmimick, line 133: ?

pkg/controller/workflow/reason.go
```diff
@@ -12,9 +12,11 @@ const (
...
 // Atlas Project reasons
 const (
-	ProjectNotCreatedInAtlas   ConditionReason = "ProjectNotCreatedInAtlas"
-	ProjectIPAccessInvalid     ConditionReason = "ProjectIPAccessListInvalid"
-	ProjectIPNotCreatedInAtlas ConditionReason = "ProjectIPAccessListNotCreatedInAtlas"
+	ProjectNotCreatedInAtlas                ConditionReason = "ProjectNotCreatedInAtlas"
+	ProjectIPAccessInvalid                  ConditionReason = "ProjectIPAccessListInvalid"
+	ProjectIPNotCreatedInAtlas              ConditionReason = "ProjectIPAccessListNotCreatedInAtlas"
+	ProjectPEServiceIsNotReadyInAtlas       ConditionReason = "ProjectPrivateEndpointServiceIsNotReadyInAtlas"
+	ProjectPrivateEndpointNotCreatedInAtlas ConditionReason = "ProjectPrivateEndpointNotCreatedInAtlas"
```
jasonmimick, line 19: Is this "Not Created" or "We tried to create it but got an error"? - the name of the condition is not so important but want to make sure we understand the workflow and what these will really mean if customers see them.

pkg/controller/atlasproject/private_endpoint.go
fabritsius: This was temporary.
The initial reason is that fetch all endpoints from all providers is considered outdated and fetch each provider separately is the preferred way. While developing for one provider I used this string for convenience.

pkg/controller/atlasproject/private_endpoint.go
fabritsius: Also temporary. Same reason as before.

pkg/controller/workflow/reason.go
fabritsius: Good point. I will think about these names. Keep in mind that there is also a message for each such error which is more descriptive.

priyolahiri (approved):

PR #357, opened 2021-12-07
soft delete projects

Freyert:
### All Submissions:

When adding the `mongodb.com/atlas-resource-policy: keep` annotation the resource actually needs to be deleted from K8S even if we keep it in Atlas. Otherwise the controller will continuously try to delete the K8S CR.

If something tries to delete an object and the infinite delete loop occurs I don't think it's possible to update or change the object being deleted in any way. At least helm will refuse to do it.


fabritsius:
The PR #388 was merged instead so closing this one. Thank you 🙏

PR #373, opened 2022-01-10
Specify registry for the operator image, and bypass kind cluster creation.

chatton:
### All Submissions:

This is a small patch to enable us to use clusters other than kind when deploying the operator.

### DIFF
--- a/scripts/create_kind_cluster.sh
+++ b/scripts/create_kind_cluster.sh
@@ -2,6 +2,11 @@
 # sample from https://kind.sigs.k8s.io/docs/user/local-registry/
 set -o errexit
 
+if [[ "${USE_KIND:-}" = "false" ]]; then
+  echo "skipping creation of kind cluster"
+  exit 0
+fi
+
 if [[ "$(kind get clusters | grep -x kind)" = "kind" ]]; then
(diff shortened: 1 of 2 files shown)


scripts/create_kind_cluster.sh
```diff
@@ -2,6 +2,11 @@
 # sample from https://kind.sigs.k8s.io/docs/user/local-registry/
 set -o errexit
 
+if [[ "${USE_KIND:-}" = "false" ]]; then
```
fabritsius, line 5: I like the idea behind this change. It makes sense that we don't always use kind. We can change this script even further, make it generic and also change the name accordingly (something without `kind`).

Alternatively we can create a separate script for other types and add similar branching in the `Makefile`. This way we won't need to skip `kind` cluster creation inside of `create_kind_cluster.sh` script.

leo-ri (approved):

scripts/create_kind_cluster.sh
chatton: I think in the future we can make something more flexible! For now I think it makes sense to de-couple the creation of a kind cluster with the operator deployment.

PR #392, opened 2022-01-18
Cancel previous test runs

chatton:
### All Submissions:

This new step cancels previous test runs from the same branch. This means if we make and push changes while a test is still running we can cancel the previous test as it is no longer required.


leo-ri:
self-requested : ) Can we not merge this one until  CLOUDP-110780, CLOUDP-110604 are done? (I will notify)

leo-ri:
`GitHub actions` released a new feature for canceling pipelines not long time ago - using the `vanilla` way for now.

PR #411, opened 2022-02-07
CLOUDP-110604: update ginkgo

leo-ri:


.golangci.yml
```diff
@@ -82,13 +82,13 @@ issues:
         - gochecknoglobals
         - wrapcheck
         - nlreturn
+        - errcheck
```
leo-ri, line 85: Ginkgo default reporter return err - we do not check it in tests

fabritsius (approved):
Looks great 👍

PR #414, opened 2022-02-10
Enable x509 user auth

fabritsius:


scripts/create_x509.go
```diff
@@ -0,0 +1,88 @@
+package main
```
igor-karpukhin, line 1: one suggestion: maybe it worth to add this thing to the Makefile and then mention in the documentation that one can build it and use the binary to generate certificates?

igor-karpukhin (approved):
Looks good. Should we also add a test for this feature?

scripts/create_x509.go
fabritsius: Good call, I actually had such Makefile section during the development 😅

scripts/create_x509.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	if err := os.MkdirAll(basePath, os.ModePerm); err != nil {
+		fmt.Println("failed to create path", err)
+	}
+
+	// save private key
+	pkey := x509.MarshalPKCS1PrivateKey(privatekey)
+	pkeyPath := filepath.Join(basePath, "private.key")
+	_ = ioutil.WriteFile(pkeyPath, pkey, 0600)
```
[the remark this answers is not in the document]
fabritsius: Sure 👌

pkg/api/v1/status/atlasproject.go
```diff
@@ -72,4 +72,7 @@ type AtlasProjectStatus struct {
 
 	// The list of private endpoints configured for current project
 	PrivateEndpoints []ProjectPrivateEndpoint `json:"privateEndpoints,omitempty"`
+
+	// X509Enabled signifies if self-managed X.509 authentication is configured
+	X509Enabled bool `json:"x509Enabled,omitempty"`
```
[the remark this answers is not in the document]
fabritsius: Agreed, thanks for the idea 💡

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -73,10 +73,13 @@ type AtlasDatabaseUserSpec struct {
...
 
 	// PasswordSecret is a reference to the Secret keeping the user password.
-	PasswordSecret *ResourceRef `json:"passwordSecretRef"`
+	PasswordSecret *ResourceRef `json:"passwordSecretRef,omitempty"`
 
 	// Username is a username for authenticating to MongoDB.
 	Username string `json:"username"`
+
+	// X509Type is X.509 method by which the database authenticates the provided username
+	X509Type string `json:"x509Type,omitempty"`
```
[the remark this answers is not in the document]
fabritsius: I think explicit X.509 is better but. `X509Type` is an actual field of the resulting request and it can be something other than `CUSTOMER` (so we just exposing this field to the user), this way it's easier to follow from the existing Atlas documentation. At the same time, I will think again about it (maybe there is actually a better solution, like you've suggested).

pkg/api/v1/status/atlasproject.go
```diff
@@ -72,4 +73,43 @@ type AtlasProjectStatus struct {
 
 	// The list of private endpoints configured for current project
 	PrivateEndpoints []ProjectPrivateEndpoint `json:"privateEndpoints,omitempty"`
+
+	// AuthModes contains a list of configured authentication modes
+	// "SCRAM" is default authentication method and requires a password for each user
+	// "X509" signifies that self-managed X.509 authentication is configured
+	AuthModes []authmode.AuthMode `json:"AuthModes,omitempty"`
```
fabritsius, line 80: Added an AuthModes list to the Status.
Just realised that I need to put the default `SCRAM` somewhere, even though the field is not used at this stage.

config/manager/kustomization.yaml
```diff
@@ -7,5 +7,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: mongodb-atlas-controller
-  newTag: latest
+  newName: localhost:5000/mongodb-atlas-kubernetes-operator
```
leo-ri, line 10: typo

fabritsius: Thank you 🙏

pkg/api/v1/status/atlasproject.go
fabritsius: Also noticed that this status field isn't passed properly, fixing...

fabritsius: I manually add "SCRAM" to the status. I is never removed cause there is no way to disable default auth through the operator.

igor-karpukhin (approved):
Looks good!

chatton (approved):
LGTM! Nice work with this one 👍

PR #416, opened 2022-02-11
e2e(bundles): add logs

leo-ri:


igor-karpukhin (approved):
Looks good! Quick question: do we need `SaveOperatorLogs` and other two methods if we use `SaveClusterDump`?

fabritsius (approved):
Good Code 👍

PR #426, opened 2022-02-21
CLOUDP-111985: Create advanced clusters in atlas

chatton:
### All Submissions:

This PR adds support for advanced cluster support. With a basic integration test.

A follow up PR will will update the helm repo to allow configuration for advancedCluster through helm, and add e2e tests (+ more integration tests)


pkg/controller/atlascluster/advanced_cluster.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"go.mongodb.org/atlas/mongodbatlas"
+	"go.uber.org/zap"
+
+	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/workflow"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/compat"
+)
+
+func (r *AtlasClusterReconciler) ensureAdvancedClusterState(ctx *workflow.Context, project *mdbv1.AtlasProject, cluster *mdbv1.AtlasCluster) (*mongodbatlas.AdvancedCluster, workflow.Result) {
```
chatton, line 18: this function is very similar to the regular cluster. There are changes to all of the types to the various functions (api client, merge and clean functions and return type)

I looked into defining an interface, but it ended up being messier than I had hoped.

pkg/api/v1/atlascluster_types.go
```diff
@@ -116,6 +116,99 @@ type AtlasClusterSpec struct {
 	Project ResourceRefNamespaced `json:"projectRef"`
 
 	ClusterSpec *ClusterSpec `json:"clusterSpec,omitempty"`
```
fabritsius, line 118: We can also mark ClusterSpec as optional to signify that you need one or the other.

pkg/controller/atlascluster/advanced_cluster.go
fabritsius: We can try to refactor this again after Go 18, some of the added features might be useful here.

pkg/controller/atlasproject/private_endpoint.go
```diff
@@ -91,6 +91,7 @@ func getStatusForInterfaceConnections(ctx *workflow.Context, projectID string) w
 			return workflow.Terminate(workflow.Internal, err.Error())
 		}
 
+		// interfaceEndpoint.Status is for the interface endpoint AZURE
```
fabritsius, line 94:
```suggestion
		// interfaceEndpoint.Status is for the Azure and GCP interface endpoints
```

Until recently it was `AzureConnectionStatus` but they changed this to just Status with the added support of GCP.

fabritsius (approved):
Great Code 👍

pkg/controller/atlascluster/advanced_cluster.go
chatton: yes this was actually a perfect candidate for generics, I was thinking about this when I was writing it!

pkg/api/v1/atlascluster_types.go
chatton: I believe the default is optional unless you specify required, but we can add it to be explicit

config/crd/bases/atlas.mongodb.com_atlasclusters.yaml
```diff
@@ -36,6 +36,225 @@ spec:
           spec:
             description: AtlasClusterSpec defines the desired state of AtlasCluster
             properties:
+              advancedClusterSpec:
```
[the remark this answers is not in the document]
chatton: the generation happens automatically, and it looks like it is alphabetical.

PR #428, opened 2022-02-21
fix pe azure test

leo-ri:


chatton (approved):
LGTM Thanks for fixing this test!

PR #442, opened 2022-03-02
Added validation for cluster resource

chatton:
### All Submissions:

This PR validates the new advancedCluster/regularCluster spec changes.

The validation functions for the other types isn't strictly required, but I think it's a good idea to have all resources showing the same conditions.

### DIFF
--- a/pkg/controller/validate/validate.go
+++ b/pkg/controller/validate/validate.go
@@ -1,0 +1,29 @@
+package validate
+
+import (
+	"errors"
+
+	"github.com/hashicorp/go-multierror"
+
+	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
+)
+
+func ClusterSpec(clusterSpec mdbv1.AtlasClusterSpec) error {
+	var err error
+	if clusterSpec.AdvancedClusterSpec == nil && clusterSpec.ClusterSpec == nil {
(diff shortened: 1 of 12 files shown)


pkg/controller/validate/validate.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/hashicorp/go-multierror"
+
+	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
+)
+
+func ClusterSpec(clusterSpec mdbv1.AtlasClusterSpec) error {
+	var err error
+	if clusterSpec.AdvancedClusterSpec == nil && clusterSpec.ClusterSpec == nil {
+		err = multierror.Append(err, errors.New("expected exactly one of spec.clusterSpec or spec.advancedClusterSpec, neither were present"))
```
igor-karpukhin, line 14: Any advantages of using multi-error here?

igor-karpukhin (approved):
Looks good

pkg/controller/validate/validate.go
chatton: in this specific use case no, since we have only one error ever returned, but in general I think using multi error to include a list of all validation errors is better than returning the first one we encounter.

fabritsius (approved):
Looks good 👍

PR #497, opened 2022-04-12
CLOUDP-114365: Allow to use cert-manager for X.509 Certificates

fabritsius:


pkg/controller/atlasproject/x509_auth.go
```diff
@@ -70,14 +73,37 @@ func readX509CertFromSecret(kubeClient client.Client, secretRef client.ObjectKey
 		return "", err
 	}
 
-	if len(secret.Data) != 1 {
-		return "", errors.New("the secret should contain one X.509 certificate")
+	defaultName := "ca.crt"
```
igor-karpukhin, line 76: I guess this can be a constant

igor-karpukhin (approved):
Looks good! It would be nice to have an integration test for this feature.

pkg/controller/atlasproject/x509_auth.go
fabritsius: Good point 👍

PR #553, opened 2022-06-09
Improvment of start-guide.md

Sugar-pack:
### All Submissions:

Add a few required dependencies to the list of dependencies
Correct the syntax error in the How-To header
Add known issues section

### DIFF
--- a/docs/dev/start-guide.md
+++ b/docs/dev/start-guide.md
@@ -4,6 +4,8 @@
 2. Install dependencies
 ```
 # on Mac
+brew install kubebuilder # or https://book.kubebuilder.io/quick-start.html
+brew install kustomize # or https://kubectl.docs.kubernetes.io/installation/kustomize/
 brew install act # or https://github.com/nektos/act#installation
 brew install yq # or https://pypi.org/project/yq/
 brew install shellcheck # or https://github.com/koalaman/shellcheck#installing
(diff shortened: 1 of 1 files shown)


docs/dev/start-guide.md
```diff
@@ -39,3 +41,7 @@ When running integration tests from an IDE the following environment variables n
 `ATLAS_PUBLIC_KEY=<..>`
 `ATLAS_PRIVATE_KEY=<..>`
 `GINKGO_EDITOR_INTEGRATION=true`
+
+# Known issues
+If you get an error `controller-gen: command not found` or `kustomize: command not found` while running `make deploy`
```
fabritsius, line 46: This section can probably be safely removed as this issue won't happen if you follow the commands above and errors like "command not found" are pretty self-explanatory. What do you think?

Sugar-pack: Yes, I understand what you mean, my point is that the kustomize and the controller-gen are installed in the  $GOPATH/bin folder, and not in the standard path.

igor-karpukhin: That's right, but the $GOBIN should be a part of $PATH

fabritsius: All of the tools can be installed in various ways and for different operating systems. This guide is for other developers so we can assume they will manage =)

Sugar-pack: Got you) Done

fabritsius (approved):

igor-karpukhin (approved):

PR #563, opened 2022-06-21
CLOUDP-126252: Maintenance Windows support

Julien-Ben:
### All Submissions:

# Maintenance window feature guide 

## Example of definition of a maintenance window for an atlas project :

```
projectMaintenanceWindow:
  dayOfWeek: 5
  hourOfDay: 1
  autoDefer: false
  startASAP: false
  defer: false
```
If no field specified, maintenance window is deleted (if it exists)
User must always specify at least `dayOfWeek`. `hourOfDay` is defaulted to 0 (midnight), `autoDefer` is defaulted to false
Only one of (`startASAP`, `defer`) can be true

`dayOfWeek`, `hourOfDay` and `autoDefer` represent the actual state of the window in Atlas 
`defer` and `startASAP` are used to send one request only (and then the user should reset them to `false`)

## How the operator works :

1. Check if window needs to be deleted
2. Check if window needs to be updated (`hourOfDay`, `dayOfWeek`)
    1. If yes, update (`hourOfDay`, `dayOfWeek`, `autoDefer`) according to YAML spec
    2. If no, check if `autoDefer` must be toggled (there is a specific request in Atlas API for this field)
3. Check if a `startASAP` request needs to be sent, if yes send and return
4. Check if a `defer` request needs to be sent, if yes send and return

Although it doesn’t make much sense, it is possible both to toggle auto-deferral and to ask the maintenance to start immediately through the API, so the operator reflects this possibility as well 
Note : `defer` works only if a maintenance is scheduled (a notification has been sent to the user 72h prior to the beginning), the operator doesn’t check for that and simply send the deferral request  which might fail if no maintenance is scheduled

### DIFF
--- a/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
+++ b/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
@@ -320,10 +320,6 @@ spec:
                     description: Flag indicating whether any scheduled project maintenance
                       should be deferred automatically for one week.
                     type: boolean
-                  autoDeferOnceEnabled:
-                    description: Flag that indicates whether you want to defer all
-                      maintenance windows one week they would be triggered.
-                    type: boolean
                   dayOfWeek:
                     description: Day of the week when you would like the maintenance
                       window to start as a 1-based integer. Sunday 1, Monday 2, Tuesday
--- a/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
+++ b/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
@@ -324,6 +324,8 @@ spec:
                     description: Day of the week when you would like the maintenance
                       window to start as a 1-based integer. Sunday 1, Monday 2, Tuesday
                       3, Wednesday 4, Thursday 5, Friday 6, Saturday 7
+                    maximum: 7
+                    minimum: 1
                     type: integer
                   defer:
                     description: Flag indicating whether the next scheduled project
@@ -333,6 +335,8 @@ spec:
                     description: Hour of the day when you would like the maintenance
                       window to start. This parameter uses the 24-hour clock, where
                       midnight is 0, noon is 12.
+                    maximum: 23
+                    minimum: 0
                     type: integer
                   startASAP:
                     description: Flag indicating whether project maintenance has been
--- a/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
+++ b/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
@@ -227,6 +227,40 @@ spec:
                       type: object
                   type: object
                 type: array
+              maintenanceWindow:
+                description: MaintenanceWindow allows to specify a preferred time
+                  in the week to run maintenance operations. See more information
+                  at https://www.mongodb.com/docs/atlas/reference/api/maintenance-windows/
+                properties:
+                  autoDefer:
+                    description: Flag indicating whether any scheduled project maintenance
+                      should be deferred automatically for one week.
+                    type: boolean
+                  dayOfWeek:
+                    description: Day of the week when you would like the maintenance
+                      window to start as a 1-based integer. Sunday 1, Monday 2, Tuesday
+                      3, Wednesday 4, Thursday 5, Friday 6, Saturday 7
+                    maximum: 7
+                    minimum: 1
+                    type: integer
+                  defer:
+                    description: Flag indicating whether the next scheduled project
+                      maintenance should be deferred for one week. Cannot be specified
+                      if startASAP is true
+                    type: boolean
+                  hourOfDay:
+                    description: Hour of the day when you would like the maintenance
+                      window to start. This parameter uses the 24-hour clock, where
+                      midnight is 0, noon is 12.
+                    maximum: 23
+                    minimum: 0
+                    type: integer
+                  startASAP:
+                    description: Flag indicating whether project maintenance has been
+                      directed to start immediately. Cannot be specified if defer
+                      is true
+                    type: boolean
+                type: object
               name:
                 description: Name is the name of the Project that is created in Atlas
                   by the Operator if it doesn't exist yet.
@@ -311,38 +345,6 @@ spec:
                       type: string
                   type: object
                 type: array
-              projectMaintenanceWindow:
-                description: ProjectMaintenanceWindow allow to specify a preferred
-                  time in the week to run maintenance operations. See more information
-                  at https://www.mongodb.com/docs/atlas/reference/api/maintenance-windows/
-                properties:
-                  autoDefer:
-                    description: Flag indicating whether any scheduled project maintenance
-                      should be deferred automatically for one week.
-                    type: boolean
-                  dayOfWeek:
-                    description: Day of the week when you would like the maintenance
-                      window to start as a 1-based integer. Sunday 1, Monday 2, Tuesday
-                      3, Wednesday 4, Thursday 5, Friday 6, Saturday 7
-                    maximum: 7
-                    minimum: 1
-                    type: integer
-                  defer:
-                    description: Flag indicating whether the next scheduled project
-                      maintenance should be deferred for one week.
-                    type: boolean
-                  hourOfDay:
-                    description: Hour of the day when you would like the maintenance
-                      window to start. This parameter uses the 24-hour clock, where
-                      midnight is 0, noon is 12.
-                    maximum: 23
-                    minimum: 0
-                    type: integer
-                  startASAP:
-                    description: Flag indicating whether project maintenance has been
-                      directed to start immediately.
-                    type: boolean
-                type: object
               withDefaultAlertsSettings:
                 default: true
                 description: Flag that indicates whether to create the new project
--- a/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
+++ b/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
@@ -311,6 +311,30 @@ spec:
                       type: string
                   type: object
                 type: array
+              projectMaintenanceWindow:
+                description: ProjectMaintenanceWindow allow to specify a preferred
+                  time in the week to run maintenance operations. See more information
+                  at https://www.mongodb.com/docs/atlas/reference/api/maintenance-windows/
+                properties:
+                  autoDeferOnceEnabled:
+                    description: Flag that indicates whether you want to defer all
+                      maintenance windows one week they would be triggered.
+                    type: boolean
+                  dayOfWeek:
+                    description: Day of the week when you would like the maintenance
+                      window to start as a 1-based integer. Sunday 1, Monday 2, Tuesday
+                      3, Wednesday 4, Thursday 5, Friday 6, Saturday 7
+                    type: integer
+                  hourOfDay:
+                    description: Hour of the day when you would like the maintenance
+                      window to start. This parameter uses the 24-hour clock, where
+                      midnight is 0, noon is 12.
+                    type: integer
+                  startASAP:
+                    description: Flag indicating whether project maintenance has been
+                      directed to start immediately.
+                    type: boolean
+                type: object
               withDefaultAlertsSettings:
                 default: true
                 description: Flag that indicates whether to create the new project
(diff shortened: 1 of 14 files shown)


pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -201,6 +201,12 @@ func (r *AtlasProjectReconciler) Reconcile(context context.Context, req ctrl.Req
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	if result = ensureMaintenanceWindow(ctx, projectID, project); !result.IsOk() {
+		ctx.SetConditionTrue(status.MaintenanceWindowReadyType)
+		r.EventRecorder.Event(project, "Normal", string(status.MaintenanceWindowReadyType), "")
```
fabritsius, line 206: Looks like this creates a "Normal" event only in case of a "not OK" result.

Julien-Ben: Thank you for your feedback ! I've indeed noticed it this morning and it has been fixed in a commit
I am still reviewing my code to spot bugs and I will mark the PR as ready for review when I'm done

fabritsius: Nice 👍

config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
```diff
@@ -311,6 +311,38 @@ spec:
                       type: string
                   type: object
                 type: array
+              projectMaintenanceWindow:
+                description: ProjectMaintenanceWindow allow to specify a preferred
```
igor-karpukhin, line 315:
```suggestion
                description: ProjectMaintenanceWindow allows to specify a preferred
```

config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
```diff
@@ -311,6 +311,38 @@ spec:
...
+                      maintenance should be deferred for one week.
+                    type: boolean
+                  hourOfDay:
+                    description: Hour of the day when you would like the maintenance
+                      window to start. This parameter uses the 24-hour clock, where
+                      midnight is 0, noon is 12.
+                    maximum: 23
+                    minimum: 0
+                    type: integer
+                  startASAP:
```
igor-karpukhin, line 341: Consider mentioning that `defer` and `startASAP` are mutually exclusive options

config/samples/atlas_v1_atlasproject.yaml
```diff
@@ -6,4 +6,10 @@ spec:
...
     - ipAddress: "192.0.2.15"
-      comment: "IP address for Application Server A"
\ No newline at end of file
+      comment: "IP address for Application Server A"
+  projectMaintenanceWindow:
+    dayOfWeek: 7
+    hourOfDay: 18
+    autoDefer: false
+    #startASAP: true
+    #defer: true
```
igor-karpukhin, line 15: Consider adding another example file (e.g. `atlas_v1_atlasproject_with_maintenance_window.yaml`) with this object. Also, please remove commented fields if they are not a part of the example.

config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
```diff
@@ -311,6 +311,38 @@ spec:
                       type: string
                   type: object
                 type: array
+              projectMaintenanceWindow:
```
igor-karpukhin, line 314: I think we can rename that to be just `maintenanceWindow`, because it is already a part of a `project` resource

pkg/api/v1/atlasproject_types.go
```diff
@@ -55,6 +55,11 @@ type AtlasProjectSpec struct {
 	// +optional
 	ProjectIPAccessList []project.IPAccessList `json:"projectIpAccessList,omitempty"`
 
+	// ProjectMaintenanceWindow allow to specify a preferred time in the week to run maintenance operations. See more
+	// information at https://www.mongodb.com/docs/atlas/reference/api/maintenance-windows/
+	// +optional
+	ProjectMaintenanceWindow project.MaintenanceWindow `json:"projectMaintenanceWindow,omitempty"`
```
igor-karpukhin, line 61: Same thing with renaming

pkg/api/v1/project/maintenancewindow.go
```diff
@@ -0,0 +1,75 @@
...
+	// Flag indicating whether the next scheduled project maintenance should be deferred for one week.
+	// +optional
+	Defer bool `json:"defer,omitempty"`
+}
+
+// ToAtlas converts the ProjectMaintenanceWindow to native Atlas client format.
+func (m MaintenanceWindow) ToAtlas() (*mongodbatlas.MaintenanceWindow, error) {
+	hourOfDayValue := m.HourOfDay
+	startASAPValue := m.StartASAP
+	autoDeferValue := m.AutoDefer
```
igor-karpukhin, line 35: I assume these assignments are just for taking references later below. Please consider taking references directly with `&m.StartASAP` and etc.

pkg/controller/atlasproject/maintenancewindow.go
```diff
@@ -0,0 +1,164 @@
...
+		// To avoid any unexpected behavior, we send a request to the API containing only the StartASAP flag,
+		// although the API should ignore other fields in that case
+		if result := createOrUpdateInAtlas(ctx.Client, projectID, project.NewMaintenanceWindow().WithStartASAP(true)); !result.IsOk() {
+			return result
+		}
+		// Nothing else should be done after sending a StartASAP request
+		return workflow.OK()
+	}
+
+	if windowSpec.Defer {
```
igor-karpukhin, line 66: Shouldn't `startASAP` and `Defer` be mutually exclusive? What happens if I specify both?

pkg/controller/atlasproject/maintenancewindow.go
```diff
@@ -0,0 +1,164 @@
...
+func maxOneFlag(window project.MaintenanceWindow) bool {
+	sum := 0
+	if window.StartASAP {
+		sum++
+	}
+	if window.Defer {
+		sum++
+	}
+	return sum <= 1
+}
```
igor-karpukhin, line 99: I assume you want to check if only one flag is set? Consider simplifying it like that: `!(window.StartASAP && window.Defer)`

pkg/api/v1/project/maintenancewindow.go
```diff
@@ -0,0 +1,75 @@
...
+	return &mongodbatlas.MaintenanceWindow{
+		DayOfWeek:            m.DayOfWeek,
+		HourOfDay:            &hourOfDayValue,
+		StartASAP:            &startASAPValue,
```
igor-karpukhin, line 39: This is quite a weird combination of pointers usage. Let's not forget to fix the in the atlas-go library as well.

igor-karpukhin (changes requested):
In general, looks great 👍 Let's rename `projectMaintenanceWindow` to `maintenanceWindow` and I guess we're good to go.
cc: @fabritsius @leo-ri please take a look as well

config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
Julien-Ben: Should we do the same with `projectIpAccessList` for consistency then ?

pkg/controller/atlasproject/maintenancewindow.go
Julien-Ben: The function `validateMaintenanceWindow` should take care of that, should I also do the check in the Reconcile loop ?

pkg/controller/atlasproject/maintenancewindow.go
Julien-Ben: Good catch, I've implemented it like that because there were more than 2 flags originally, but now it simplifies the code a lot.

priyolahiri (approved):
LGTM

config/samples/atlas_v1_atlasproject_with_maintenance_window.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  name: Test Atlas Operator Project
+  projectIpAccessList:
+    - ipAddress: "192.0.2.15"
+      comment: "IP address for Application Server A"
+  maintenanceWindow:
+    dayOfWeek: 3
+    hourOfDay: 5
+    autoDefer: true
+    startASAP: false
```
fabritsius, line 14: Do we really need to define all of the fields here? I think it's better to omit optional fields with default values.

pkg/api/v1/project/maintenancewindow.go
```diff
@@ -0,0 +1,74 @@
...
+	// +optional
+	StartASAP bool `json:"startASAP,omitempty"`
+	// Flag indicating whether the next scheduled project maintenance should be deferred for one week.
+	// Cannot be specified if startASAP is true
+	// +optional
+	Defer bool `json:"defer,omitempty"`
+}
+
+// ToAtlas converts the MaintenanceWindow to native Atlas client format.
+func (m MaintenanceWindow) ToAtlas() (*mongodbatlas.MaintenanceWindow, error) {
```
fabritsius, line 34: We can safely remove error output since it's always nil and there is no need to check for it. I don't think there is some prototype which needs ToAtlas method with a specific definition.

pkg/util/testutil/maintenancewindow_matcher.go
```diff
@@ -0,0 +1,50 @@
...
+type maintenanceWindowMatcher struct {
+	ExpectedMaintenanceWindow project.MaintenanceWindow
+}
+
+func (m *maintenanceWindowMatcher) Match(actual interface{}) (success bool, err error) {
+	var c *mongodbatlas.MaintenanceWindow
+	var ok bool
+	if c, ok = actual.(*mongodbatlas.MaintenanceWindow); !ok {
+		actualType := reflect.TypeOf(actual)
+		panic("Expected *mongodbatlas.MaintenanceWindow but received type " + actualType.String())
```
fabritsius, line 30: I think we can swap this for an error for consistency =)

pkg/controller/atlasproject/maintenancewindow.go
```diff
@@ -0,0 +1,157 @@
...
+// that hour of day and day of week are in the bounds - this will be done by Atlas.
+func validateMaintenanceWindow(window project.MaintenanceWindow) error {
+	if isEmptyWindow(window) || (windowSpecified(window) && maxOneFlag(window)) {
+		return nil
+	}
+	return errors.New(`
+		projectMaintenanceWindow must respect the following constraints, or be empty :
+			1) dayOfWeek must be specified, hourOfDay is 0 by default, autoDeferral is false by default
+			2) only one of (startASAP, defer) is true
+	`)
```
fabritsius, line 104: This multiline comment looks great in the code but will be ugly in the state. Better to change this message so it looks fine in the state.

pkg/controller/atlasproject/maintenancewindow.go
```diff
@@ -0,0 +1,157 @@
...
+// in Atlas but is not specified in the CR, it is deleted.
+func ensureMaintenanceWindow(ctx *workflow.Context, projectID string, atlasProject *mdbv1.AtlasProject) workflow.Result {
+	windowSpec := atlasProject.Spec.MaintenanceWindow
+	if err := validateMaintenanceWindow(windowSpec); err != nil {
+		return workflow.Terminate(workflow.ProjectWindowInvalid, err.Error())
+	}
+
+	if isEmptyWindow(windowSpec) {
+		ctx.Log.Debugw("Window empty or undefined, deleting in Atlas")
+		if result := deleteInAtlas(ctx.Client, projectID); !result.IsOk() {
```
fabritsius, line 25: This will send `DELETE` request each reconcile call.
We can send delete only if `status` is not empty which `spec` is empty.

fabritsius (changes requested):
Looks great but better to make a few changes 👍

pkg/controller/atlasproject/maintenancewindow.go
Julien-Ben: The function `isEmptyWindow` will return true only if none of the fields are specified. In that case we will indeed delete the window.
Otherwise no delete request should be sent.

Do you think the behaviour should be different ?

fabritsius: Default project has no maintenance window defined so `isEmptyWindow` will return `true` and we will make the `DELETE` request on each call of Reconcile. This isn't critical but it is possible to avoid it by setting and checking the status for the maintenance window.

Just a suggestion though =)

fabritsius: We can merge my PR with conditions removal logic. You will have to rebase but afterwards you will be able to use the presence of condition for determining if you need to delete anything in Atlas.

fabritsius (approved):
Great Job 👍

leo-ri (approved):

PR #566, opened 2022-06-23
CLOUDP-125096: Unset conditions if the resource is no longer in use

fabritsius:


igor-karpukhin (approved):
Looks good 👍

Julien-Ben (approved):
I like the changes to the Maintenance Window feature, it makes the reconciliation loop easier to read and understand

Maybe `validateMaintenanceWindow` should be called outside of `syncAtlasWithSpec`

fabritsius:
Moving it may look something like this:
```

	if err := validateMaintenanceWindow(atlasProject.Spec.MaintenanceWindow); err != nil {
		result := workflow.Terminate(workflow.ProjectWindowInvalid, err.Error())
		ctx.SetConditionFromResult(status.MaintenanceWindowReadyType, result)
		return result
	}

	if result := syncAtlasWithSpec(ctx, projectID, atlasProject.Spec.MaintenanceWindow); !result.IsOk() {
		ctx.SetConditionFromResult(status.MaintenanceWindowReadyType, result)
		return result
	}
```

One downside is that we will need an extra call of `SetConditionFromResult`. Not sure if it's better. What do you think?

igor-karpukhin (approved):
Looks good 👍

leo-ri (approved):

Issue #588, opened 2022-07-04
Creating AtlasDeployment with advancedDeploymentSpec fails with diskGB


sunchill06:
**What did you do to encounter the bug?**

```console
❯ cat <<EOF | kubectl apply -f -
apiVersion: atlas.mongodb.com/v1
kind: AtlasDeployment
metadata:
  name: new-cluster-1
spec:
  projectRef:
    name: test-project
  advancedDeploymentSpec:
      backupEnabled: true
      diskSizeGB: 80
      mongoDBMajorVersion: "4.4"
      pitEnabled: false
      replicationSpecs:
      - numShards: 3
        regionConfigs:
        - electableSpecs:
            instanceSize: M40
            nodeCount: 3
          autoScaling:
            compute:
              enabled: true
              maxInstanceSize: M60
              minInstanceSize: M40
              scaleDownEnabled: true
            diskGB:
              enabled: true
          providerName: GCP
          regionName: WESTERN_EUROPE
EOF
error: error validating "STDIN": error validating data: ValidationError(AtlasDeployment.spec.advancedDeploymentSpec.replicationSpecs[0].regionConfigs[0].autoScaling): unknown field "diskGB" in com.mongodb.atlas.v1.AtlasDeployment.spec.advancedDeploymentSpec.replicationSpecs.regionConfigs.autoScaling; if you choose to ignore these errors, turn validation off with --validate=false

❯ helm ls
NAME                  	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART                       	APP VERSION
mongodb-atlas-operator	mongodb  	22      	2022-07-04 12:08:47.672815367 +0000 UTC	deployed	mongodb-atlas-operator-1.1.0	1.1.0 

```

However in the atlas [documentation](https://www.mongodb.com/docs/atlas/reference/api/cluster-advanced/create-one-cluster-advanced/) I can see 
<img width="1470" alt="Screenshot 2022-07-04 at 16 40 37" src="https://user-images.githubusercontent.com/26785680/177186334-c45869bb-c412-40cf-a617-b129571784e3.png">

**What did you expect?**
I was expecting the spec to run just fine. Following works fine though.

```console
❯ cat <<EOF | kubectl apply -f -
apiVersion: atlas.mongodb.com/v1
kind: AtlasDeployment
metadata:
  name: new-cluster-1
spec:
  projectRef:
    name: test-project
  advancedDeploymentSpec:
      backupEnabled: true
[...]

PR #592, opened 2022-07-05
Change diskGB field in advancedDeploymentSpec (#588)

Sugar-pack:
Add new AdvancedAutoScalingSpec struct for AdvancedDeploymentSpec
Change existed field to the new struct

### All Submissions:


pkg/api/v1/atlasdeployment_types.go
```diff
@@ -250,6 +250,25 @@ type AutoScalingSpec struct {
...
 }
 
+// AdvancedAutoScalingSpec configures your deployment to automatically scale its storage
+type AdvancedAutoScalingSpec struct {
+	// Flag that indicates whether autopilot mode for Performance Advisor is enabled.
+	// The default is false.
+	AutoIndexingEnabled *bool `json:"autoIndexingEnabled,omitempty"`
+	// Flag that indicates whether disk auto-scaling is enabled. The default is true.
+	// +optional
+	DiskGBEnabled *DiskGBEnabled `json:"diskGB,omitempty"`
```
fabritsius, line 260: I think we can rename the field to `DiskGB` to match json object.

Also potentially can rename `DiskGBEnabled` struct to something general like `Enabled` so it can be reused somewhere (even thought there is nowhere to reuse it for now).

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -250,6 +250,25 @@ type AutoScalingSpec struct {
 	Compute *ComputeSpec `json:"compute,omitempty"`
 }
 
+// AdvancedAutoScalingSpec configures your deployment to automatically scale its storage
+type AdvancedAutoScalingSpec struct {
+	// Flag that indicates whether autopilot mode for Performance Advisor is enabled.
+	// The default is false.
+	AutoIndexingEnabled *bool `json:"autoIndexingEnabled,omitempty"`
```
fabritsius, line 257: Found no mention of `autoIndexingEnabled` field in [the docs](https://www.mongodb.com/docs/atlas/reference/api/cluster-advanced/create-one-cluster-advanced/).
I think `autoScaling` has only `diskGB` and `compute` subfields.

This is not related to the issue but we can fix in bulk =)

pkg/api/v1/atlasdeployment_types.go
Sugar-pack: Renamed to DiskGB.
Second option is too general. Reusing fields with common names was the cause of this bug.

pkg/api/v1/atlasdeployment_types.go
Sugar-pack: Done. Thanks)

pkg/api/v1/atlasdeployment_types.go
fabritsius: Thanks for the change :)

In the first part I was actually talking about the `DiskGBEnabled` field itself, not the type.
Can agree with you regarding the potential change 👍

fabritsius (approved):
Nice Work 👍

fabritsius (approved):

PR #594, opened 2022-07-07
CLOUDP-127362: Add validation for deployment name

fabritsius:
The name of the deployment can only contain ASCII letters, numbers, and hyphens.
This changes fixes the example in the README and adds regex validation for the `name` field.

If you try to create a deployment with the incorrect name you will get something like this:
```
The AtlasDeployment "my-atlas-deployment" is invalid: spec.deploymentSpec.name: Invalid value: "test deployment": spec.deploymentSpec.name in body should match '^[\w-]+$'
```

I wonder if regex like this one is better `^[A-Za-z0-9-]+$` for readability 🤔 

This PR also corrects "deployment" back to the "cluster" in some links.


Julien-Ben:
I think we should go for the more readable regex `^[A-Za-z0-9-]+$`, it is a bit more "user-friendly".

Julien-Ben (approved):
LGTM 👍 

It's also nice that you started fixing the documentation links (`reference/api/clusters-advanced/`).

igor-karpukhin (approved):
Looks good 👍

Sugar-pack (approved):
Great work!

config/crd/bases/atlas.mongodb.com_atlasdeployments.yaml
```diff
@@ -141,7 +141,7 @@ spec:
                   name:
                     description: Name of the deployment as it appears in Atlas. After
                       Atlas creates the deployment, you can't change its name.
-                    pattern: ^[\w-]+$
+                    pattern: ^[A-Za-z0-9-]+$
```
igor-karpukhin, line 144: I think the underscore sign is missing in this regex

PR #596, opened 2022-07-12
CLOUDP-127414: PE cleanup action

Sugar-pack:
### All Submissions:

Add a new step to Clean Atlas Organization action. It is needed when e2e tests are stopped manually or panicked. In this case, there may be unremoved private endpoints.


.github/actions/cleanup-pe/cleantagged.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"google.golang.org/api/compute/v1"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/actions/cloud"
+	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/api/gcp"
+)
+
+func cleanAllTaggedAWSPE(region, tagName, tagValue string) error {
```
fabritsius, line 20: All of the "tagged" function versions look a lot like their not-tagged counterparts (close to duplication). What do you think about using the same cleaning function for tagged and non-tagged? You can add a flag or just make a decision based on the fact that tagName and tagValue are both empty.

.github/actions/cleanup-pe/cleanPE.go
```diff
@@ -0,0 +1,93 @@
...
+	switch cleanAllEnv {
+	case "true":
+		cleanOnlyTaggedPE = true
+	case "false":
+	default:
+		log.Fatal("CLEAN_TAGGED_PE must be set to true or false")
+	}
+	err := SetGCPCredentials()
+	if err != nil {
+		log.Fatal(err)
```
Julien-Ben, line 26: Should we interrupt execution after a fatal error occurs ?

.github/actions/cleanup-pe/cleantagged.go
fabritsius: Or make 2 wrapper functions on top of the same core function might be also a great idea

.github/actions/cleanup-pe/cleanPE.go
Sugar-pack: I think so. The program will stop if it doesn't get the credentials or if it receives an error from another function. Log.Fatal is equivalent to Print() followed by a call to os.Exit(1).

Julien-Ben: Oh I wasn't aware that the logging library was taking care of interrupting the execution, all good then !

fabritsius (approved):
Looks Great 👍

fabritsius (approved):

PR #607, opened 2022-07-15
CLOUDP-127788 : Import resources from Atlas 

Julien-Ben:
### All Submissions:

# Feature : Import resources from Atlas into Kubernetes

## Demo

I've made a step by step demo of the script in action. It shows how to use it and the resulting resources inside my local k8s cluster (with Lens).

https://youtu.be/GN1zBy-RFBM

## How to use ?

⚠️ To avoid losing any Kubernetes resources, run the make target in a separate namespace, this is done by default in the current version of the Makefile ⚠️ 
- Ensure your `.actrc` contains your Atlas credentials.
- Ensure Docker is running
- `make run-import`

The Makefile takes care of :
- Creating a kubernetes cluster (the import script should automatically retrieve the corresponding configuration)
- Creating a namespace for imported resources through `kubectl`
- Applying necessary CRDs to the cluster
- Deleting any previously instantiated resources in the namespace, including ⚠️ **any secrets** ⚠️ 
- Retrieving your credentials from `.actrc` 

You can change the import command in the `run-import` target. By default, the `all` command is runned.

## Available commands

```
//Import selected deployments from a project
atlas-import project “projectID1” --deployments=”deploymentID1, deploymentID2, deploymentID3”
//Import all deployments from a project
atlas-import project “projectID2” --all
//Import every Atlas resources accessible from provided credentials
atlas-import all

The following flags are needed in addition, for the 3 above commands :
--org $(ORG) --publickey $(PUBLICKEY) --privatekey $(PRIVATEKEY)
In order to specify Atlas Organization ID, public and private keys

Optionally, the user can specify a kubernetes namespace in which to import resources : 
--import-namespace “example-namespace”

Optionally, the script can be run in verbose mode, to display DEBUG logs :
--verbose

All the above configurations can be passed through a YAML file as well :
//Read import configuration from a file
atlas-import “from-config config.yaml”
```

## For reviewers

This PR is quite long and will take some time to review (750+ lines for the biggest file). Through this guide and the demo, I'll try to make it easier to understand.
The commit history is broken because of the last rebase I performed.

## Other useful documentation

- [Tecnical Specifications](https://docs.google.com/document/d/1R5bqi7pHJz94x34ajjwhDXKCGB2Lvi77VCofcihVnVE/edit#heading=h.my8l8kwcujpo) Provides an overview of the algorithm, the different resources existing in Atlas, and some implementation considerations

### More tests
I would advise to use existing methods inside `atlasdeployment_types_test.go ` in order to compare atlas resources to kubernetes resources when writing e2e tests.

`toLowercaseAlphaNumeric` and `stringNameConcatenation` should be moved inside a util folder and unit tested. It is likely that they will be used for e2e testing as well.

### Concurrent run
The script performance is strongly network-bound. I advise to import each project in a separate go routine, and even each deployments.

### Import resources into a file
Following discussions on the Spec document, it might be useful for users to not directly import their resources in their kubernetes cluster, but instead save them in a file, to apply them later (through GitOps for example).
This work would require quite some refactoring because at the moment, the script needs a kubernetes client in multiple places and not only for the final steps, for example when converting ThirdPartyIntegrations from Atlas to Kubernetes.

### Credential management
As suggested on the Spec document, the management of Atlas credentials could be simplified.

### Letting users pick other types of resources to import
For example, all database users are imported with a project. We should let users decide which one they want to import.


cmd/atlas-import/cli.go
```diff
@@ -0,0 +1,169 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+)
+
+func generateBaseConfig(cmd *cobra.Command) AtlasImportConfig {
+	baseConfig := AtlasImportConfig{
```
jamesbroadhead, line 14: I quite strongly think that this should share config & API keys with the atlascli project

cmd/atlas-import/cli.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func main() {
+	importer.Log, _ = zap.NewDevelopment()
+	Execute()
+}
+
+func run(config importer.AtlasImportConfig) {
+	println(config.OrgID)
+	print(config.PrivateKey)
```
igor-karpukhin, line 89: Are these for debugging?

cmd/atlas-import/importer/import.go
```diff
@@ -0,0 +1,774 @@
...
+// setUpAtlasClient instantiate the client to interact with the Atlas API
+// Credentials are provided in the import configuration
+func setUpAtlasClient(config *AtlasImportConfig) (*mongodbatlas.Client, error) {
+	Log.Debug("Creating AtlasClient")
+	credentials := atlas.Connection{
+		OrgID:      config.OrgID,
+		PublicKey:  config.PublicKey,
+		PrivateKey: config.PrivateKey,
+	}
+	atlasDomain := "https://cloud.mongodb.com/"
```
igor-karpukhin, line 69: I think you can add this as a default value to the `domainFlag`

PR #617, opened 2022-07-26
Fix backup export and policies

Julien-Ben:
### All Submissions:

* [X] Update docs/release-notes/release-notes.md if your changes should be included in the release notes for the next release.

The export field should be a pointer, to avoid instantiating the underlying struct when not specified in Atlas.
The enum validation was wrong, the specified set of value works only for hourly backup, but not for the other frequencies.


igor-karpukhin (approved):
Looks good 👍

PR #628, opened 2022-08-03
Fixed Kustomize downloader

igor-karpukhin:
### All Submissions:

Fixed `kustomize` tool downloader

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,10 +145,12 @@ controller-gen: ## Download controller-gen locally if necessary
 .PHONY: kustomize
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -145,7 +145,10 @@ controller-gen: ## Download controller-gen locally if necessary
 .PHONY: kustomize
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.0.1)
+	rm -f ./kustomize
```
fabritsius, line 148: ```## Download kustomize locally if necessary```

I think we should check if there is a binary and install only if it's missing

igor-karpukhin: Good point. Fixed!

Sugar-pack (approved):

Issue #611, opened 2022-07-19
Decreasing the Cluster disk size integration test logic doesn't seem right 


yzdann:
**What did you do to encounter the bug?**
- Run this Integration test 
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/9719113ca7ca49171da3ded7458a69c83b1e3e0a/test/int/deployment_test.go#L438

**What did you expect?**
The test scenario decreases the disk size by changing from a higher number to a lower number.

**What happened instead?**
This line creates an `M10` with a 10GB cluster and runs the test against the cluster:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/9719113ca7ca49171da3ded7458a69c83b1e3e0a/test/int/deployment_test.go#L476

On line `505` the test tries to set the disk to 10GB but the disk size was already 10GB, so no decrease will happen on this scenario as far as i understand:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/9719113ca7ca49171da3ded7458a69c83b1e3e0a/test/int/deployment_test.go#L504-L514

**Screenshots**
(nothing to add)

**Operator Information**
- v1.1.0 (not related to the issue)

**Kubernetes Cluster Information**
- (not related to the issue)

yzdann:
Also even if we ran this test on an instance with a 20GB disk (for example) ... the disk size won't decrease to 10GB 

https://github.com/mongodb/mongodb-atlas-kubernetes/blob/9719113ca7ca49171da3ded7458a69c83b1e3e0a/test/int/deployment_test.go#L505
because in `DefaultAWSCluster` the `atlasdeployments.spec.deploymentSpec.autoScaling.diskGBEnabled` is true and reconciliation will ignore `DiskSizeGB` after first intiliaztion.

https://github.com/mongodb/mongodb-atlas-kubernetes/blob/445d66aa1c044b70f233a2dc49d9cffbc959d1b5/pkg/controller/atlasdeployment/deployment.go#L137-L139

PR #634, opened 2022-08-06
CLOUDP-130483: Fixed diskSizeGB decreasing

igor-karpukhin:
### All Submissions:

Fixed a case when disc size couldn't be decreased if autoscaling is enabled

* Closes: https://github.com/mongodb/mongodb-atlas-kubernetes/issues/611


test/int/deployment_test.go
```diff
@@ -460,6 +460,36 @@ var _ = Describe("AtlasDeployment", Label("int", "AtlasDeployment"), func() {
...
+			})
+
+			By("Decreasing the Deployment disk size", func() {
+				createdDeployment.Spec.DeploymentSpec.DiskSizeGB = intptr(12)
+				performUpdate(30 * time.Minute)
+				doRegularDeploymentStatusChecks()
+				checkAtlasState(func(c *mongodbatlas.Cluster) {
+					Expect(*c.DiskSizeGB).To(BeEquivalentTo(*createdDeployment.Spec.DeploymentSpec.DiskSizeGB))
+
+					// check whether https://github.com/mongodb/go-client-mongodb-atlas/issues/140 is fixed
```
Sugar-pack, line 487: Is this comment still relevant?

igor-karpukhin: Yes, but we have a different issue for this one. It will be removed later.

test/int/deployment_test.go
```diff
@@ -490,7 +520,7 @@ var _ = Describe("AtlasDeployment", Label("int", "AtlasDeployment"), func() {
 			})
 
 			By("Decreasing the Deployment disk size", func() {
-				createdDeployment.Spec.DeploymentSpec.DiskSizeGB = intptr(10)
+				createdDeployment.Spec.DeploymentSpec.DiskSizeGB = intptr(15)
```
fabritsius, line 523: I don't think we set `DiskSizeGB` for the `DefaultAWSDeployment` which probably means it is 10 (which is the minimum accepted value). Which probably means this line is actually increases the `DiskSizeGB`.

pkg/controller/atlasdeployment/deployment.go
```diff
@@ -133,10 +133,6 @@ func removeOutdatedFields(removeFrom *mongodbatlas.Cluster, lookAt *mongodbatlas
 		} else {
 			result.ProviderSettings.AutoScaling.Compute = &mongodbatlas.Compute{}
 		}
-
-		if lookAt.AutoScaling.DiskGBEnabled != nil && *lookAt.AutoScaling.DiskGBEnabled {
```
yzdann, line 137: Hey @igor-karpukhin,
Are you changing the whole concept of the operator should not touch disk/instance size if autoscaling is enabled?

For example, we have a deploymentA(Kubernetes object) with a 10GB disk, and disk autoscaling is enabled.
After some time, autoscaler on the Atlas platform decided to increase the disk size to 15GB, the increasing process is finished and now our atlas cluster has a 15GB disk.
What value should we expect in the Kubernetes object related to that deploymentA 10GB or 15GB?

yzdann:
resolve: https://github.com/mongodb/mongodb-atlas-kubernetes/issues/611

test/int/deployment_test.go
igor-karpukhin: I changed the default disk size to 20GB, but It looks like I forgot to add this file, thanks!

pkg/controller/atlasdeployment/deployment.go
igor-karpukhin: The API technically allows disk size to be configured with autoscaling enabled. But, I just revert it to the previous state so it's not configurable if the autoscaling is enabled.

fabritsius (approved):
Looks Great 👍

PR #654, opened 2022-08-25
CLOUDP-134285: Add a small doc for annotations

fabritsius:


docs/annotations.md
```diff
@@ -0,0 +1,26 @@
...
+  name: my-project
+  annotations:
+    mongodb.com/atlas-resource-policy: keep
+spec:
+  name: Test Atlas Operator Project
+```
+
+### mongodb.com/atlas-resource-policy=keep
+
+If `mongodb.com/atlas-resource-policy` is set to `keep` operator will not delete the Atlas resource when you delete the k8s resource.
```
Sugar-pack, line 20: Maybe "custom resource/k8s custom resource" is better?

fabritsius: In that case it's not just any k8s resource but the one which corresponds to Atlas resource.
If there is some confusion here maybe I should change to:
```
when you delete the corresponding k8s resource.
```
Adding "custom" might sound a bit better but I decided not to add it as it actually doesn't mean anything useful, all operator resources are custom.

Sugar-pack (approved):
LGTM

PR #664, opened 2022-08-30
CLOUDP-135171: Fix error message for Delete method

Sugar-pack:
Fix test function usage

### All Submissions:

### DIFF
--- a/pkg/controller/atlasdeployment/atlasdeployment_controller.go
+++ b/pkg/controller/atlasdeployment/atlasdeployment_controller.go
@@ -259,19 +259,11 @@ func (r *AtlasDeploymentReconciler) handleDeploymentBackupSchedule(ctx *workflow
 	}
 	apiScheduleRes.Policies = []mongodbatlas.Policy{apiPolicy}
 
-	currentSchedule, response, err := ctx.Client.CloudProviderSnapshotBackupPolicies.Delete(context.Background(), projectID, cName)
+	currentSchedule, _, err := ctx.Client.CloudProviderSnapshotBackupPolicies.Delete(context.Background(), projectID, cName)
 	if err != nil {
 		errMessage := "unable to delete current backup policy for project"
-		r.Log.Debugf("%s: %v:%v, %v", errMessage, projectID, cName, err)
-		return fmt.Errorf("%s: %v:%v, %w", errMessage, projectID, cName, err)
-	}
-	if currentSchedule == nil {
-		r.Log.Debugf("got empty response for deleting backup schedule for project: %v:%v", projectID, cName)
-		if response != nil {
-			return fmt.Errorf("can't delete сurrent backupschedule. got an empty response from the api. response status: %s", response.Status)
-		} else {
-			return fmt.Errorf("can't delete сurrent backupschedule. got an empty response from the api. response is nil")
-		}
+		r.Log.Debugf("%s: %s:%s, %v", errMessage, projectID, cName, err)
+		return fmt.Errorf("%s: %s:%s, %w", errMessage, projectID, cName, err)
 	}
 
 	r.Log.Debugf("successfully deleted backup policy. Default schedule received: %v", currentSchedule)
--- a/pkg/controller/atlasdeployment/atlasdeployment_controller.go
+++ b/pkg/controller/atlasdeployment/atlasdeployment_controller.go
@@ -261,15 +261,15 @@ func (r *AtlasDeploymentReconciler) handleDeploymentBackupSchedule(ctx *workflow
 
 	currentSchedule, response, err := ctx.Client.CloudProviderSnapshotBackupPolicies.Delete(context.Background(), projectID, cName)
 	if err != nil {
-		r.Log.Debugf("unable to delete current backup policy for project: %v:%v, %v", projectID, cName, err)
-		return fmt.Errorf("unable to delete current backup policy for project: %v:%v, %w", projectID, cName, err)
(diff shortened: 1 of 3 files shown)


pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -259,13 +259,19 @@ func (r *AtlasDeploymentReconciler) handleDeploymentBackupSchedule(ctx *workflow
...
 	}
 	if currentSchedule == nil {
-		r.Log.Debugf("got empty response for deleting backup schedule for project: %v:%v", projectID, cName)
-		return fmt.Errorf("can't delete сurrent backupschedule. got an empty response from the api: %v", currentSchedule)
+		if response != nil {
+			r.Log.Debugf("got empty response for deleting backup schedule for project: %v:%v", projectID, cName)
+			return fmt.Errorf("can't delete сurrent backupschedule. got an empty response from the api. err: %w, response: %v", err, response)
+		} else {
+			r.Log.Debugf("got empty response for deleting backup schedule for project: %v:%v", projectID, cName)
+			return fmt.Errorf("can't delete сurrent backupschedule. got an empty response from the api. err: %w", err)
```
fabritsius, line 273: Well, I suggest we rewrite this section. At this stage the `error` is likely to be <nil> and it's actually better to log something from the response as long as it isn't empty. Response is an object but I think we can just log only status from it.

Sugar-pack: Done. Thank you

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -259,13 +259,19 @@ func (r *AtlasDeploymentReconciler) handleDeploymentBackupSchedule(ctx *workflow
 	}
 	apiScheduleRes.Policies = []mongodbatlas.Policy{apiPolicy}
 
-	currentSchedule, _, err := ctx.Client.CloudProviderSnapshotBackupPolicies.Delete(context.Background(), projectID, cName)
+	currentSchedule, response, err := ctx.Client.CloudProviderSnapshotBackupPolicies.Delete(context.Background(), projectID, cName)
 	if err != nil {
 		r.Log.Debugf("unable to delete current backup policy for project: %v:%v, %v", projectID, cName, err)
+		return fmt.Errorf("unable to delete current backup policy for project: %v:%v, %w", projectID, cName, err)
```
igor-karpukhin, line 265: code dup nit: print the error message into a string and then use it in both calls

Sugar-pack: Done. Thank you!

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -259,13 +259,19 @@ func (r *AtlasDeploymentReconciler) handleDeploymentBackupSchedule(ctx *workflow
 	}
 	apiScheduleRes.Policies = []mongodbatlas.Policy{apiPolicy}
 
-	currentSchedule, _, err := ctx.Client.CloudProviderSnapshotBackupPolicies.Delete(context.Background(), projectID, cName)
+	currentSchedule, response, err := ctx.Client.CloudProviderSnapshotBackupPolicies.Delete(context.Background(), projectID, cName)
 	if err != nil {
-		r.Log.Debugf("unable to delete current backup policy for project: %v:%v, %v", projectID, cName, err)
+		errMessage := "unable to delete current backup policy for project"
+		r.Log.Debugf("%s: %v:%v, %v", errMessage, projectID, cName, err)
+		return fmt.Errorf("%s: %v:%v, %w", errMessage, projectID, cName, err)
```
fabritsius, line 266:
```suggestion
		return fmt.Errorf("%s: %s:%s, %w", errMessage, projectID, cName, err)
```

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -259,13 +259,19 @@ func (r *AtlasDeploymentReconciler) handleDeploymentBackupSchedule(ctx *workflow
...
 	}
 	if currentSchedule == nil {
 		r.Log.Debugf("got empty response for deleting backup schedule for project: %v:%v", projectID, cName)
-		return fmt.Errorf("can't delete сurrent backupschedule. got an empty response from the api: %v", currentSchedule)
+		if response != nil {
+			return fmt.Errorf("can't delete сurrent backupschedule. got an empty response from the api. response status: %s", response.Status)
```
fabritsius, line 271: `got an empty response from the api` seems like it should be removed from both here and from of the `else` statement. If the request was made `response` is never empty, if the response is empty, the request was never made, but the `error` was supposed to be returned (meaning we will `return` prior to this line).

So we can assume that response is not `nil` on this line, no need to check.

To be fair I'm not sure this logic is correct as `response` is returned both for success and failures, meaning

fabritsius (changes requested):
Left a couple of suggestions :)

pkg/controller/atlasdeployment/atlasdeployment_controller.go
Sugar-pack: Done

pkg/controller/atlasdeployment/atlasdeployment_controller.go
Sugar-pack: Done

fabritsius (approved):
Great Fix 👍

PR #666, opened 2022-09-02
CLOUDP-135177: Fix deployment CR deletion if token invalid

### LINKED ISSUE #421: Cluster CR deleted while using expired API Keys
**What did you do to encounter the bug?**
Create An API Key on the Cloud Atlas
Create a k8 secret on the Kubernetes cluster
Deploy the Project CR on the Kubernetes. Project created on the Cloud Atlas
Deploy the Cluster CR on the Kubernetes. Cluster created on the Cloud Atlas
Using Atlas console , Delete the API Key
Delete the Cluster CR . Cluster CR deleted while the cluster on the Cloud Atlas remain active

**What did you expect?**
The Delete CR operation should register an error and continue reconcile until a valid K8 secret is created. This will avoid having an orphan cluster and allows user to take corrective action.

**What happened instead?**
A clear and concise description of what happened instead
The Cluster CR is deleted, while the cluster on the Atlas Cloud remains active. On the MongoDB Operator POD , the reconcile keep getting HTTP error 401. 

**Screenshots**

**Operator Information**
 - Operator Version

 - Distribution
 - Version
 - Image Registry location (quay, or an internal registry)

**Additional context**

If possible, please include:
 - `kubectl describe` output
 - yaml definitions for your objects
 - log files for the operator and database pods
[log (3).txt](https://github.com/mongodb/mongodb-atlas-kubernetes/files/8063495/log.3.txt)

Sugar-pack:
closes #421
### All Submissions:

### DIFF
--- a/pkg/controller/atlasdeployment/atlasdeployment_controller.go
+++ b/pkg/controller/atlasdeployment/atlasdeployment_controller.go
@@ -302,13 +302,16 @@ func (r *AtlasDeploymentReconciler) handleDeploymentBackupSchedule(customContext
 	}
 	apiScheduleRes.Policies = []mongodbatlas.Policy{apiPolicy}
 
+
 	currentSchedule, _, err := customContext.Client.CloudProviderSnapshotBackupPolicies.Delete(ctx, projectID, cName)
 	if err != nil {
-		r.Log.Debugf("unable to delete current backup policy for project: %v:%v, %v", projectID, cName, err)
+		errMessage := "unable to delete current backup policy for project"
+		r.Log.Debugf("%s: %s:%s, %v", errMessage, projectID, cName, err)
+		return fmt.Errorf("%s: %s:%s, %w", errMessage, projectID, cName, err)
 	}
-	if currentSchedule == nil {
-		r.Log.Debugf("got empty response for deleting backup schedule for project: %v:%v", projectID, cName)
-		return fmt.Errorf("can't delete сurrent backupschedule. got an empty response from the api: %v", currentSchedule)
+
+	if currentSchedule == nil && response != nil {
+		return fmt.Errorf("can't delete сurrent backupschedule. response status: %s", response.Status)
 	}
 
 	r.Log.Debugf("successfully deleted backup policy. Default schedule received: %v", currentSchedule)
--- a/pkg/controller/atlasdeployment/atlasdeployment_controller.go
+++ b/pkg/controller/atlasdeployment/atlasdeployment_controller.go
@@ -83,7 +83,7 @@ type AtlasDeploymentReconciler struct {
 
 // +kubebuilder:rbac:groups="",namespace=default,resources=events,verbs=create;patch
 
-func (r *AtlasDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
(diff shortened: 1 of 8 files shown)


pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -95,61 +94,112 @@ func (r *AtlasDeploymentReconciler) Reconcile(context context.Context, req ctrl.
...
+			deployment.SetFinalizers(append(deployment.GetFinalizers(), finalizer))
+			if err := r.Client.Update(ctx, deployment); err != nil {
+				result = workflow.Terminate(workflow.Internal, err.Error())
+				customContext.SetConditionFromResult(status.DeploymentReadyType, result)
+				return result.ReconcileResult(), nil
+			}
+		}
+	}
+
+	if !deployment.GetDeletionTimestamp().IsZero() {
```
igor-karpukhin, line 155: Can be replaced with `else` by joining with previous `if` statement?

fabritsius: I think this is the same style as for projects. Sometimes it's considered a good practice to avoid `else` statement as it requires the developer to remember what was in the first `if` statement.

In that case I think both variants are Ok as the blocks aren't big.

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -84,7 +83,7 @@ type AtlasDeploymentReconciler struct {
 
 // +kubebuilder:rbac:groups="",namespace=default,resources=events,verbs=create;patch
 
-func (r *AtlasDeploymentReconciler) Reconcile(context context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *AtlasDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
fabritsius, line 86: Do not agree with this change. It is true that `ctx` is used for `context.Context` in most projects and it was I bit confusing for me too. But this project was created with a different notation and short `ctx` is used for custom object as it is more common.

I think it might be a good idea to refactor this and maybe change the naming a bit. But this has to be a global change for the entire project, otherwise it is confusing. I also think `customContext` is too long and there is actually no reason for our context to even be called "context", and we can just switch it to `state` or something along these lines.

We also don't use `context.Context` properly. So such refactoring with proper `context` usage might be a separate change which we discuss and implement in the future.

fabritsius (changes requested):
Looks good but I think we should revert one of the changes you've made 👍

pkg/controller/atlasdeployment/atlasdeployment_controller.go
Sugar-pack: Done

fabritsius (approved):
Great Job 👍

fabritsius (approved):
Looks Good 👍 
Left a comment)

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -450,27 +502,22 @@ func (r *AtlasDeploymentReconciler) Delete(e event.DeleteEvent) error {
 
 	log.Infow("-> Starting AtlasDeployment deletion", "spec", deployment.Spec)
 
+	ctx := context.Background()
```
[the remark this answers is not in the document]
Sugar-pack: Done

PR #674, opened 2022-09-08
CLOUDP-126255: Add encryption at rest

fabritsius:


pkg/controller/workflow/reason.go
```diff
@@ -29,6 +29,8 @@ const (
 	ProjectIntegrationReady                  ConditionReason = "ProjectIntegrationReady"
 	ProjectPrivateEndpointIsNotReadyInAtlas  ConditionReason = "ProjectPrivateEndpointIsNotReadyInAtlas"
 	ProjectNetworkPeerIsNotReadyInAtlas      ConditionReason = "ProjectNetworkPeerIsNotReadyInAtlas"
+	ProjectEncryptionAtRestReady             ConditionReason = "ProjectEncryptionAtRestReady"
+	ProjectEncryptionAtRestFailed            ConditionReason = "ProjectEncryptionAtRestFailed"
```
Sugar-pack, line 33: If status with such reason have "false" status, is Encryption at rest ready or not?

test/e2e/encryption_at_rest_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// todo: fill in
+}
+
+func fillKMSforGCP(encAtRest *v1.EncryptionAtRest) {
+	if (encAtRest.GoogleCloudKms == v1.GoogleCloudKms{}) {
+		return
+	}
+
+	// todo: fill in
```
Sugar-pack, line 189: Does it need to be completed, or can it be removed?

test/e2e/encryption_at_rest_test.go
```diff
@@ -0,0 +1,190 @@
...
+			},
+			[]cloudaccess.Role{
+				{
+					Name: utils.RandomName(awsRoleNameBase),
+					AccessRole: v1.CloudProviderAccessRole{
+						ProviderName: "AWS",
+					},
+				},
+			},
+		),
```
Sugar-pack, line 101: It seems that it is worth adding tests for GCP and Azure.

pkg/controller/workflow/reason.go
fabritsius: Initially plan was to set this status if parameters are wrong and user has to change something but at this point I've changed this since so this status can be safely removed. Thanks for the catch 👍🏻

test/e2e/encryption_at_rest_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+	encAtRest.AwsKms.CustomerMasterKeyID = CustomerMasterKeyID
+}
+
+func fillVaultforAzure(encAtRest *v1.EncryptionAtRest) {
+	if (encAtRest.AzureKeyVault == v1.AzureKeyVault{}) {
+		return
+	}
+
+	// todo: fill in
```
igor-karpukhin, line 181: Here as well. Is some functionality missing?

test/e2e/encryption_at_rest_test.go
fabritsius: Yeah, this is a placeholder for future Azure tests. Can stay as is for now, I think 🤔

test/e2e/encryption_at_rest_test.go
fabritsius: I don't think they are quite necessary in this initial version as operator logic is actually identical for each provider, only the object changes. We can use the same logic as for `integrations` and omit them for now.

test/e2e/encryption_at_rest_test.go
fabritsius: Save as above. Just a placeholder for future tests. I think it's ok to leave as is for now and fill in at a later point.

Sugar-pack (approved):
Great work :sunglasses:

igor-karpukhin (approved):
Good job 👍

PR #715, opened 2022-10-14
CLOUDP-136586: Allow autoscaling instanceSize and diskSizeGB outside of the manually set instanceSize and diskSizeGB

helderjs:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasdeployment/advanced_deployment.go
+++ b/pkg/controller/atlasdeployment/advanced_deployment.go
@@ -241,6 +241,7 @@ func normalizeInstanceSize(currentInstanceSize string, autoscaling *mdbv1.Advanc
 	return currentInstanceSize
 }
 
+// extractNumberFromInstanceTypeName get the existing number from a given instance type name, fail when the name is incorrect
 func extractNumberFromInstanceTypeName(instanceTypeName string) int {
 	name := strings.TrimPrefix(instanceTypeName, "M")
 	name = strings.TrimPrefix(name, "R")
--- a/pkg/controller/atlasdeployment/advanced_deployment.go
+++ b/pkg/controller/atlasdeployment/advanced_deployment.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strconv"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -110,9 +112,9 @@ func cleanupTheSpec(deployment *mdbv1.AdvancedDeploymentSpec) {
 //	if region config has enabled compute autoscaling
 func handleAutoscaling(kubeDeployment *mdbv1.AdvancedDeploymentSpec) {
 	isDiskAutoScaled := false
-	cleanupInstanceSize := func(s *mdbv1.Specs) {
+	syncInstanceSize := func(s *mdbv1.Specs, as *mdbv1.AdvancedAutoScalingSpec) {
 		if s != nil {
-			s.InstanceSize = ""
+			s.InstanceSize = normalizeInstanceSize(s.InstanceSize, as)
 		}
 	}
 	for _, repSpec := range kubeDeployment.ReplicationSpecs {
@@ -127,9 +129,9 @@ func handleAutoscaling(kubeDeployment *mdbv1.AdvancedDeploymentSpec) {
 				if regConfig.AutoScaling.Compute != nil &&
 					regConfig.AutoScaling.Compute.Enabled != nil &&
 					*regConfig.AutoScaling.Compute.Enabled {
-					cleanupInstanceSize(regConfig.ElectableSpecs)
-					cleanupInstanceSize(regConfig.AnalyticsSpecs)
-					cleanupInstanceSize(regConfig.ReadOnlySpecs)
+					syncInstanceSize(regConfig.ElectableSpecs, regConfig.AutoScaling)
+					syncInstanceSize(regConfig.AnalyticsSpecs, regConfig.AutoScaling)
+					syncInstanceSize(regConfig.ReadOnlySpecs, regConfig.AutoScaling)
 				}
 			}
 		}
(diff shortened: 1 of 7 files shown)


pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -222,3 +224,29 @@ func GetAllDeploymentNames(client mongodbatlas.Client, projectID string) ([]stri
...
+
+	return currentInstanceSize
+}
+
+func extractNumberFromInstanceTypeName(instanceTypeName string) int {
+	name := strings.TrimPrefix(instanceTypeName, "M")
+	name = strings.TrimPrefix(name, "R")
+	name = strings.TrimSuffix(name, "_NVME")
+
+	number, _ := strconv.Atoi(name)
```
igor-karpukhin, line 249: Why ignore an error here? We do not have any kubebuilder validations on the `instanceSizeName` field. I'd say it's better to explicitly return an error and put it into the status that the instance size name is incorrect.

pkg/controller/atlasdeployment/advanced_deployment_test.go
```diff
@@ -453,3 +559,72 @@ func TestAdvancedDeployment_handleAutoscaling(t *testing.T) {
...
+	for name, test := range data {
+		t.Run(name, func(t *testing.T) {
+			assert.Equal(t, test.Number, extractNumberFromInstanceTypeName(test.Name))
+		})
+	}
+}
+
+func boolptr(b bool) *bool {
+	return &b
+}
```
igor-karpukhin, line 630: Consider replacing with `toptr.MakePtr(true)`

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -222,3 +224,29 @@ func GetAllDeploymentNames(client mongodbatlas.Client, projectID string) ([]stri
...
+	}
+
+	if currentSize > maxSize {
+		return autoscaling.Compute.MaxInstanceSize
+	}
+
+	return currentInstanceSize
+}
+
+func extractNumberFromInstanceTypeName(instanceTypeName string) int {
```
Sugar-pack, line 244: It seems to me that this function is worth providing related context(comment, link to documentation).

pkg/controller/atlasdeployment/advanced_deployment.go
igor-karpukhin: This code probably won't handle `M50` and `M50_NVME` instances

fabritsius:
Potentially we can just check for the difference of values. This will allow to handle such cases like:
`MaxInstanceSize: M40` => `MaxInstanceSize: M40_NVME`

Like if both AutoScaling margins are the same, remove the AutoScaling section, and leave it otherwise 🤔

fabritsius:
Potentially we can just check for the difference of values. This will allow to handle such cases like:
`MaxInstanceSize: M40` => `MaxInstanceSize: M40_NVME`

Like if both AutoScaling margins are the same, remove the AutoScaling section, and leave it otherwise 🤔

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -222,3 +245,32 @@ func GetAllDeploymentNames(client mongodbatlas.Client, projectID string) ([]stri
 	}
 	return deploymentNames, nil
 }
+
+func normalizeInstanceSize(ctx *workflow.Context, currentInstanceSize string, autoscaling *mdbv1.AdvancedAutoScalingSpec) (string, error) {
```
igor-karpukhin, line 249: @helderjs Will this function handle the situation when the previous instance size is equal the current one from a CR? In that case the operator shouldn't send it.

helderjs: > when the previous instance size is equal the current one
@igor-karpukhin Do you mean the when instance size didn't change? if yes, does it make a difference to send the same value or not? will atlas try to "scale" to the same size?

igor-karpukhin: @helderjs Yes, it will try to scale to the same size. The problem is, if someone modifies anything else in the CR, we will send the same value for the instance size again.

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -108,14 +111,25 @@ func cleanupTheSpec(deployment *mdbv1.AdvancedDeploymentSpec) {
...
-	cleanupInstanceSize := func(s *mdbv1.Specs) {
+	syncInstanceSize := func(s *mdbv1.Specs, as *mdbv1.AdvancedAutoScalingSpec) error {
 		if s != nil {
-			s.InstanceSize = ""
+			size, err := normalizeInstanceSize(ctx, s.InstanceSize, as)
+			if err != nil {
+				return err
+			}
+
+			if size == currentDeployment.ReplicationSpecs[0].RegionConfigs[0].ElectableSpecs.InstanceSize {
```
fabritsius, line 123: Potentially can wrap this into something like `hasInstanceSizeChanged()` just to make it more readable :)

fabritsius (approved):
Looks Great 👍

Sugar-pack (approved):
Great work!

igor-karpukhin (approved):
Good job 👍 LGTM

pkg/controller/validate/validate.go
```diff
@@ -59,3 +74,50 @@ func allAreNil(values ...interface{}) bool {
...
+
+			if regionSpec.ElectableSpecs.InstanceSize != instanceSize {
+				return err
+			}
+
+			if regionSpec.ReadOnlySpecs.InstanceSize != instanceSize {
+				return err
+			}
+
+			if regionSpec.AnalyticsSpecs.InstanceSize != instanceSize {
```
igor-karpukhin, line 96: @helderjs `Specs` needs to be checked if they are nil. They are not always filled. This is the reason why helm tests are failing.

igor-karpukhin (changes requested):
Missing nil pointer checks

igor-karpukhin (approved):
Looks good now, thanks 👍

PR #716, opened 2022-10-17
CLOUDP-136888: Add Project Settings

fabritsius:


pkg/controller/atlasproject/project_settings.go
```diff
@@ -0,0 +1,73 @@
...
+func areSettingsInSync(atlas, spec *v1.ProjectSettings) bool {
+	if areProjectSettingsEmpty(spec) {
+		return true
+	}
+
+	return reflect.DeepEqual(atlas, spec)
+}
+
+func patchSettings(ctx *workflow.Context, projectID string, spec *v1.ProjectSettings) error {
+	path := fmt.Sprintf("api/atlas/v1.0/groups/%s/settings", projectID)
```
Sugar-pack, line 59: Maybe it's better to make this path constant.

pkg/controller/atlasproject/project_settings.go
```diff
@@ -0,0 +1,73 @@
...
+func syncProjectSettings(ctx *workflow.Context, projectID string, project *v1.AtlasProject) workflow.Result {
+	spec := project.Spec.Settings
+
+	atlas, err := fetchSettings(ctx, projectID)
+	if err != nil {
+		return workflow.Terminate(workflow.ProjectSettingsReady, err.Error())
+	}
+
+	if !areSettingsInSync(atlas, spec) {
+		if err := patchSettings(ctx, projectID, spec); err != nil {
```
helderjs, line 38: I see here the we take k8s spec as source of truth. Considering the import feature, shouldn't we merge k8s spec into existing atlas settings instead of overwrite it? (this would require a upper-cased info in the docs to avoid misunderstandings).
Ignore my comment if this behavior was previously discussed and defined as it is.

pkg/controller/atlasproject/project_settings.go
```diff
@@ -0,0 +1,73 @@
...
+	if areProjectSettingsEmpty(spec) {
+		return true
+	}
+
+	return reflect.DeepEqual(atlas, spec)
+}
+
+func patchSettings(ctx *workflow.Context, projectID string, spec *v1.ProjectSettings) error {
+	path := fmt.Sprintf("api/atlas/v1.0/groups/%s/settings", projectID)
+	_, err := ctx.Client.NewRequest(context.Background(), http.MethodPatch, path, spec)
```
helderjs, line 60: Should we consider visit again this call (and other already in place) and decorate atlas client with not yet implemented endpoints? we avoid spread custom direct call to atlas api in the controllers and centralize this on the client (where it should be)

test/e2e/project_settings_test.go
```diff
@@ -0,0 +1,75 @@
...
+})
+
+func projectSettingsFlow(userData *model.TestDataProvider, settings *v1.ProjectSettings) {
+	By("Add project-settings to the project", func() {
+		userData.Project.Spec.Settings = settings
+		Expect(userData.K8SClient.Update(userData.Context, userData.Project)).Should(Succeed())
+		actions.WaitForConditionsToBecomeTrue(userData, status.ProjectSettingsReadyType, status.ReadyType)
+	})
+
+	By("Remove Auditing from the project", func() {
```
helderjs, line 70: Is this some leftover from c&p?

fabritsius: Yeah 😁

pkg/controller/atlasproject/project_settings.go
fabritsius: Export feature will just take the current settings via GET request and create a spec with these Project Settings so we don't need to do any merging here. And also the spec should be taken as source of truth in most cases.

pkg/controller/atlasproject/project_settings.go
fabritsius: Not sure. This PATCH call override is only necessary because there is a mistake in the lib implementation so as soon as this mistake is gone we can swap it for a regular lib call. So at this point I'm actually happy it is contained in this single module and the rest of the project doesn't know that this walk-around is being used.

I'm thinking of personally making a PR to the lib so it shouldn't stay like this for long:)

fabritsius: Not sure. This PATCH call override is only necessary because there is a mistake in the lib implementation so as soon as this mistake is gone we can swap it for a regular lib call. So at this point I'm actually happy it is contained in this single module and the rest of the project doesn't know that this walk-around is being used.

I'm thinking of personally making a PR to the lib so it shouldn't stay like this for long:)

pkg/controller/atlasproject/project_settings.go
fabritsius: I think it makes sense to keep in contained here until the lib is updated and we won't have to have this string here 🤔

pkg/controller/atlasproject/project_settings.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/workflow"
+)
+
+func ensureProjectSettings(ctx *workflow.Context, projectID string, project *v1.AtlasProject) (result workflow.Result) {
+	if result = syncProjectSettings(ctx, projectID, project); !result.IsOk() {
+		ctx.SetConditionFromResult(status.ProjectSettingsReadyType, result)
+		return result
+	}
+
+	if areProjectSettingsEmpty(project.Spec.Settings) {
```
igor-karpukhin, line 18: Maybe there is no need to create a separate function for that. Up to you though 😃

.github/actions/gen-install-scripts/entrypoint.sh
```diff
@@ -14,7 +14,7 @@ mkdir -p "${crds_dir}"
 mkdir -p "${openshift}"
 
 # Generate configuration and save it to `all-in-one`
-controller-gen crd:crdVersions=v1 rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+controller-gen crd:crdVersions=v1 rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases crd:allowDangerousTypes=true
```
Sugar-pack, line 17: I don't think this is a good idea. It is better to store floats in strings.

pkg/controller/atlasproject/project_settings.go
fabritsius: Yeah, this function was used in multiple places and was supposed to be more complex.
Here just checking for `nil` is fine :)

.github/actions/gen-install-scripts/entrypoint.sh
fabritsius: No need to worry, this is a temporary thing 😅

igor-karpukhin (approved):
Looks good 👍

helderjs (approved):
Looks Good

pkg/controller/atlasproject/project_settings.go
```diff
@@ -0,0 +1,95 @@
...
+	if !areSettingsInSync(atlas, spec) {
+		if err := patchSettings(ctx, projectID, spec); err != nil {
+			return workflow.Terminate(workflow.ProjectSettingsReady, err.Error())
+		}
+	}
+
+	return workflow.OK()
+}
+
+func areSettingsInSync(atlas, spec *v1.ProjectSettings) bool {
```
Sugar-pack, line 44: Maybe its better to use `isOneContainedInOther` func here

pkg/controller/atlasproject/project_settings.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, err
+	}
+	ctx.Log.Debugw("Got Project Settings", "data", data)
+
+	settings := v1.ProjectSettings(*data)
+	return &settings, nil
+}
+
+func isOneContainedInOther(one, other *v1.ProjectSettings) bool {
```
Sugar-pack, line 69: I don't fully understand the advantage of using the reflect package. This piece of code can be rewritten without it and more explicitly. For example:
```
func isOneContainedInOther(one, other *v1.ProjectSettings) bool {
	if one == nil {
		return true
	}

	if other == nil {
		return false
	}

	return compareBoolPtr(one.IsCollectDatabaseSpecificsStatisticsEnabled, other.IsCollectDatabaseSpecificsStatisticsEnabled) &&
		compareBoolPtr(one.IsDataExplorerEnabled, other.IsDataExplorerEnabled) &&
		compareBoolPtr(one.IsPerformanceAdvisorEnabled, other.IsPerformanceAdvisorEnabled) &&
		compareBoolPtr(one.IsRealtimePerformancePanelEnabled, other.IsRealtimePerformancePanelEnabled)
}

func compareBoolPtr(oneBool, otherBool *bool) bool {
	if oneBool == nil {
		return true
	}
	if otherBool == nil {
		return false
	}
	return *oneBool == *otherBool
}
```

pkg/controller/atlasproject/project_settings.go
fabritsius: I understand what you mean but I intentionally left it as is. I think having something like `isOneContainedInOther` on top level might be confusing and it reads better with `areSettingsInSync`. It's also possible to put the content of `isOneContainedInOther` into `areSettingsInSync` and remove the first func entirely but I'm also not a fan of this approach. This way the function is kind of like a in-code-comment and it explains the sync func and what it actually does.

Also lot's of the latest "project features" have a certain similarities in the code, including `..InSync` func, and I want to keep these similarities :)

Sugar-pack: Yes, my original idea was to move the body of the contains function into this function. And add a comment to that function, which explains the logic of comparing structures.

pkg/controller/atlasproject/project_settings.go
fabritsius: Imagine if they add 10 more fields to the Project Settings 🤔 
The "beauty" of your solution is dependent on the amount of fields in the struct so I think it's not as good :)

fabritsius: Also at some point in the future isOneContainedInOther can be moved out like this:
```
func isOneContainedInOther[T](one, other *T) bool
```
And it can potentially be reused for other stuff :)

Sugar-pack: But there are four fields in this structure. And if it changes, maybe there will be not only bool ptrs. 
And I don't think it's a good idea to use this approach for any type. Because this would require a lot of complicated code with type checks and other reflect magic.

fabritsius: Well, there is some logic in what you're saying)

In general I think it makes sense to create one general function (even with complex logic inside) if it allows not to call simple function on every single struct field :)

pkg/controller/atlasproject/project_settings.go
fabritsius: I think my version is better as we basically get the same readability but without any comments :)

Let's keep as is 😁

pkg/controller/atlasproject/project_settings.go
Sugar-pack: At the moment, this function is named very generically. At the same time, this is a tricky comparator that works with only one type.
Also, I do not understand why you use the reflection package here. Because it is currently needed to save 3 lines of code, but at the same time it brings several methods that replace well-known operators. And it does not help to reduce the code, nor to make this code more readable.
And I do not know of other cases in which this function would be useful in its current form.

fabritsius: The function logic is actually quite simple. We need to iterate over struct fields and `reflect` is the only way I know of in `Go`.

At this point it is an unexported function with types hardcoded so there is no expectation that it will be reused in its current form. I see no reason to over-engineer this function now so that we can reuse it later. I am just saying that since the taken `reflect-approach` is quite a general one, similar logic can be reused in the future in the exported utils func.

Sugar-pack (approved):
LGTM

PR #717, opened 2022-10-17
CLOUDP-136887: Add alert configuration feature

Sugar-pack:
### All Submissions:


igor-karpukhin (approved):
Good job 👍

pkg/api/v1/atlasproject_types.go
```diff
@@ -65,6 +65,16 @@ type AtlasProjectSpec struct {
...
 
+	// AlertConfiguration is a list of Alert Configurations configured for the current Project.
+	AlertConfigurations []AlertConfiguration `json:"alertConfigurations,omitempty"`
+
+	// AlertConfigurationsEnabled is a flag that enables/disables Alert Configurations sync for the current Project.
+	// If true - project alert configurations will be synced according to AlertConfigurations.
+	// If not - alert configurations will not be modified by the operator. They can be managed through API, cli, UI.
+	//kubebuilder:default:=false
+	// +optional
+	AlertConfigurationsEnabled bool `json:"alertConfigurationEnabled,omitempty"`
```
helderjs, line 76: Would `alertConfigurationSyncEnabled` be a better name?
At first, I was confused why there were two flags to enable the feature, this and the other inside `AletConfigurations`. They look the same except the other is nested in the struct.

Sugar-pack: Done.
Good point. Thanks

pkg/api/v1/atlasproject_types.go
```diff
@@ -65,6 +65,16 @@ type AtlasProjectSpec struct {
...
 
+	// AlertConfiguration is a list of Alert Configurations configured for the current Project.
+	AlertConfigurations []AlertConfiguration `json:"alertConfigurations,omitempty"`
+
+	// AlertConfigurationSyncEnabled is a flag that enables/disables Alert Configurations sync for the current Project.
+	// If true - project alert configurations will be synced according to AlertConfigurations.
+	// If not - alert configurations will not be modified by the operator. They can be managed through API, cli, UI.
+	//kubebuilder:default:=false
+	// +optional
+	AlertConfigurationSyncEnabled bool `json:"alertConfigurationEnabled,omitempty"`
```
helderjs, line 76: Won't you also update the name for the CRD?

helderjs:
@Sugar-pack, I was wondering. Most of the functions defined in the `atlasproject` controller package are exported. Do we want to do that?

pkg/api/v1/atlasproject_types.go
Sugar-pack: Yeah 😄 
Done

helderjs (approved):

pkg/api/v1/status/alert_configurations.go
```diff
@@ -0,0 +1,292 @@
...
+	// VictorOps API key. Populated for the VICTOR_OPS notifications type. If the key later becomes invalid, Atlas sends an email to the project owner and eventually removes the key.
+	VictorOpsAPIKey string `json:"victorOpsApiKey,omitempty"`
+	// VictorOps routing key. Populated for the VICTOR_OPS notifications type. If the key later becomes invalid, Atlas sends an email to the project owner and eventually removes the key.
+	VictorOpsRoutingKey string `json:"victorOpsRoutingKey,omitempty"`
+	// The following roles grant privileges within a project.
+	Roles []string `json:"roles,omitempty"`
+}
+
+func NotificationFromAtlas(notification mongodbatlas.Notification) Notification {
+	return Notification{
```
fabritsius, line 116: Why not use `JSONCopy` here?
```
result := & status.Notification{}
err := compat.JSONCopy(result, s)
return result, err
```

pkg/api/v1/status/alert_configurations.go
```diff
@@ -0,0 +1,292 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"go.mongodb.org/atlas/mongodbatlas"
+)
+
+type AlertConfiguration struct {
```
fabritsius, line 11: All of these look like a straight copies from the lib, it might be possible (or might not be) to swap all of these definitions to something like:
```
type AlertConfiguration mongodbatlas.AlertConfiguration
```

pkg/api/v1/status/alert_configurations.go
Sugar-pack: Then this error will have to be returned and handled somehow. And the current approach makes it so that the Operator can put any structure from the Altas library into our structure.

pkg/api/v1/status/alert_configurations.go
Sugar-pack: Doing so creates a dependency cycle

pkg/controller/atlasproject/alert_configurations.go
```diff
@@ -0,0 +1,216 @@
...
+	return checkAlertConfigurationStatuses(newStatuses)
+}
+
+func checkAlertConfigurationStatuses(statuses []status.AlertConfiguration) (workflow.Result, status.ConditionType) {
+	for _, alertConfigurationStatus := range statuses {
+		if alertConfigurationStatus.ErrorMessage != "" {
+			return workflow.Terminate(workflow.ProjectAlertConfigurationIsNotReadyInAtlas, fmt.Sprintf("failed to create alert configuration: %s", alertConfigurationStatus.ErrorMessage)), status.AlertConfigurationReadyType
+		}
+	}
+	return workflow.OK(), status.AlertConfigurationReadyType
```
fabritsius, line 70: I think there is one place in the project where where we return a condition because there can be multiple options. For this feature there is only one condition so I think it's better to remove second output. You can still avoid reperion by something like this:
```
                 condition := status.AlertConfigurationReadyType

                 if len(specToSync) == 0 {
			service.UnsetCondition(condition)
			return workflow.OK()
		}
		result := syncAlertConfigurations(ctx, service, groupID, specToSync)
		if !result.IsOk() {
			service.SetConditionFromResult(condition, result)
			return result
		}
		service.SetConditionTrue(condition)

File: pkg/api/v1/status/alert_configurations.go (hunk 4 above)
Comment (Sugar-pack) [reply pkg/api/v1/status/alert_configurations.go:11]: There is a different ErrorMessage field

**Review by fabritsius — APPROVED** [review]
Great Job 👏 
Left a few suggestions :)

File: pkg/controller/atlasproject/alert_configurations.go (hunk 5 above)
Comment (Sugar-pack) [reply pkg/controller/atlasproject/alert_configurations.go:70]: I don't fully understand what you mean.

File: pkg/api/v1/status/alert_configurations.go (hunk 4 above)
Comment (fabritsius) [reply pkg/api/v1/status/alert_configurations.go:11]: Ok. We might still be able to do it for others than?
```
type Notification mongodbatlas.Notification
type Threshold mongodbatlas.Threshold
type MetricThreshold mongodbatlas.MetricThreshold
```

pkg/api/v1/status/alert_configurations.go
fabritsius: But this is a better long-term solution as no changes will have to be made in case of the changes in structs, assuming these structs are also not hardcoded ofc.

And it's also much shorter even with error handling :)

fabritsius (approved):

pkg/api/v1/status/alert_configurations.go
Sugar-pack: The only thing that can be done this way is Notifications. And in this case, there will be no description in CRD.

pkg/api/v1/status/alert_configurations.go
Sugar-pack: I don't think so. If this structure changes in the Atlas lib, then we still have to add new fields to our structure. 
BTW after each bump of Atlas lib, we have to check if some of the JSON tags change. So it doesn't look good for long-term support.

PR #743, opened 2022-10-27
CLOUDP-141796: Remake cleaner

Sugar-pack:
### All Submissions:


.github/actions/cleanup/pe.go
```diff
@@ -0,0 +1,78 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"go.mongodb.org/atlas/mongodbatlas"
+)
+
+func deleteAllPE(ctx context.Context, client mongodbatlas.PrivateEndpointsService, projectID string) error {
```
fabritsius, line 10: I wonder why functions for Private Endpoints are not exported while for network peers are exported? What's the difference?

.github/actions/cleanup/pe.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if len(connection.EndpointGroupNames) != 0 {
+		return connection.EndpointGroupNames[0]
+	}
+
+	return ""
+}
+
+func getAllPrivateEndpoints(ctx context.Context, client mongodbatlas.PrivateEndpointsService, projectID string) ([]mongodbatlas.PrivateEndpointConnection, error) {
```
fabritsius, line 60: Not sure but I think some of these function can be taken from the operator. I think it's ok to use operator code as a dependency. Not an issue though since you've already coded it up :)

fabritsius (approved):
Looks Great 👍

.github/actions/cleanup/pe.go
Sugar-pack: What kind of functions do you mean? DeleteAllPrivateEndpoint function from atlas operator:
```
func DeleteAllPrivateEndpoints(ctx *workflow.Context, projectID string) workflow.Result {
	atlasPEs, err := getAllPrivateEndpoints(ctx.Client, projectID)
	if err != nil {
		return workflow.Terminate(workflow.Internal, err.Error())
	}

	endpointsToDelete := getEndpointsNotInSpec([]mdbv1.PrivateEndpoint{}, atlasPEs)
	return deletePrivateEndpointsFromAtlas(ctx, projectID, endpointsToDelete)
}
```
I cant use it for cleaner

.github/actions/cleanup/pe.go
Sugar-pack: Operator functions are non-exportable. And most of them are not suitable for the cleaner purpose.

.github/actions/cleanup/main.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	projectList, err := PrepareListToDelete(projects, config.DeleteAll, config.Lifetime)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	ctx := context.Background()
+	ok := DeleteProjects(ctx, client, projectList)
+	if !ok {
+		log.Printf("Not all project deleted. Please run cleaner again later")
```
igor-karpukhin, line 35: Maybe it worth to return a different exit code here?

igor-karpukhin (approved):
Looks good 👍

.github/actions/cleanup/config.go
```diff
@@ -0,0 +1,61 @@
...
+	switch deleteAllStr {
+	case "true":
+		deleteAll = true
+	case "false":
+		deleteAll = false
+	default:
+		return nil, fmt.Errorf("INPUT_CLEAN_ALL must be either true or false")
+	}
+
+	lifetimeStr := os.Getenv("MAX_PROJECT_LIFETIME")
```
helderjs, line 32: as in the previous comment, could we have defaults here and follow the same logic suggested before?

.github/actions/cleanup/pe.go
helderjs: I think it would make sense if those functions were on their own module, making the internal behavior of the controller exported seems not a good idea.

.github/actions/cleanup/project.go
```diff
@@ -0,0 +1,59 @@
...
+			}
+			if time.Since(createdTime) < time.Duration(lifetimeInHours)*time.Hour {
+				projectList = append(projectList, project)
+			}
+		}
+	}
+	return projectList, nil
+}
+
+func DeleteProjects(ctx context.Context, client mongodbatlas.Client, projectList []*mongodbatlas.Project) bool {
```
helderjs, line 30: Are other resources(alerts, custom roles...) deleted with the project or will we need to update the "cleaner" always we add new features?

.github/workflows/cleanup-all.yml
```diff
@@ -31,22 +31,20 @@ jobs:
...
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
           go-version: '1.19'
 
+      - name: Run atlas project cleaner
+        env:
+          MCLI_PUBLIC_API_KEY: ${{ secrets.ATLAS_PUBLIC_KEY }}
+          MCLI_PRIVATE_API_KEY: ${{ secrets.ATLAS_PRIVATE_KEY }}
+          MCLI_OPS_MANAGER_URL: "https://cloud-qa.mongodb.com/"
```
helderjs, line 43: once we target QA only, we could use as default in the code and let the env as a custom configuration. What do you think?

helderjs (changes requested):

.github/actions/cleanup/project.go
Sugar-pack: Only a small amount of entities may prevent the project from deleting. At the moment, the cleaner deletes everything that the operator supports and can prevent the project from being deleted.

.github/actions/cleanup/pe.go
fabritsius: Well, ofc, not as is, I mean like it is possible to change these functions so you can reuse them. My point was not to have the same basic logic in two places but it's fine as is.

.github/actions/cleanup/main.go
Sugar-pack: Done

.github/actions/cleanup/config.go
```diff
@@ -0,0 +1,61 @@
...
+type Config struct {
+	DeleteAll  bool
+	Lifetime   int // lifetime in hours
+	PublicKey  string
+	PrivateKey string
+	ManagerUrl string
+}
+
+func NewConfig() (*Config, error) {
+	deleteAllStr := os.Getenv("CLEAN_ALL")
```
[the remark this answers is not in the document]
Sugar-pack: Done

.github/actions/cleanup/config.go
Sugar-pack: Done

.github/workflows/cleanup-all.yml
Sugar-pack: Done

helderjs (approved):
Great Job!

PR #745, opened 2022-10-27
CLOUDP-136889: Project Custom Roles

helderjs:
### All Submissions:


test/e2e/custom_roles_test.go
```diff
@@ -0,0 +1,144 @@
...
+	v1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/status"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/toptr"
+	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/actions"
+	kubecli "github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/cli/kubecli"
+	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/data"
+	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/model"
+)
+
+var _ = FDescribe("UserLogin", Label("custom-roles"), func() {
```
igor-karpukhin, line 19: It looks like it should be just `Describe`

igor-karpukhin (commented):
Looks good 👍 But, one comment regarding tests

pkg/api/v1/custom_roles.go
```diff
@@ -0,0 +1,73 @@
...
+type Resource struct {
+	// Flag that indicates whether to grant the action on the cluster resource. If true, MongoDB Cloud ignores Database and Collection parameters.
+	Cluster *bool `json:"cluster,omitempty"`
+	// Human-readable label that identifies the database on which you grant the action to one MongoDB user.
+	Database *string `json:"database,omitempty"`
+	// Human-readable label that identifies the collection on which you grant the action to one MongoDB user.
+	Collection *string `json:"collection,omitempty"`
+}
+
+func (cr CustomRole) ToAtlas() *mongodbatlas.CustomDBRole {
```
Sugar-pack, line 39: Generated code has pointer receiver methods. I think it's better to use the same approach for all methods.

pkg/controller/atlasproject/custom_roles.go
```diff
@@ -0,0 +1,201 @@
...
+		ctx.UnsetCondition(status.ProjectCustomRolesReadyType)
+	}
+
+	return workflow.OK()
+}
+
+func fetchCustomRoles(ctx *workflow.Context, projectID string) ([]v1.CustomRole, error) {
+	data, _, err := ctx.Client.CustomDBRoles.List(context.Background(), projectID, nil)
+	if err != nil {
+		return nil, err
```
Sugar-pack, line 59: It might be worth wrapping up errors because they will be in status. And without context, it might be hard to understand what exactly happened.

pkg/controller/atlasproject/custom_roles.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+	}
+
+	ctx.Log.Debugw("Custom Roles to be added", "NumItems", len(toCreate))
+
+	for _, customRole := range toCreate {
+		_, _, err := ctx.Client.CustomDBRoles.Create(context.Background(), projectID, customRole.ToAtlas())
+
+		if err != nil {
+			return err
```
Sugar-pack, line 184: I'm not sure if it's worth returning an error here. For example, the user provided invalid data for the first role. And the Atlas API returns an error when it is created. Shouldn't we try to create all the other roles? Because the current code will try to create the first non-valid role and get stuck there.

Sugar-pack: Other feature has structures in the spec and structures in the status. 
The status for each element contains related information: errors, IDs, and all related fields from the API response.
Thus, it is possible to correlate the element from the spec (in this case, it is a role) and the element in the status, thereby finding out the state of the element.

Sugar-pack (changes requested):
Please provide the status for each role. More information in the comments

pkg/controller/atlasproject/custom_roles.go
Sugar-pack: Also, if the users provide two (or more) non-valid role specs, then they will receive only one error. So they will fix one role spec. And one more error will appear. 
They have to fix one error per edit. But the status for each role approach allows users to fix all errors.

helderjs: @Sugar-pack, it was an excellent point to add context to the errors and apply all the changes rather than fail quickly. I've applied those changes. However, for custom roles specifically, I believe is not necessary to have structured status once there's no complex operation going on on the main or in the nested structures.

pkg/controller/atlasproject/custom_roles.go
```diff
@@ -0,0 +1,207 @@
...
+			err = multierror.Append(err, fmt.Errorf("failed to create custom role \"%s\": %w", customRole.Name, creErr))
+		}
+
+		ctx.Log.Debugw("Created Custom Role in current AtlasProject", "custom role", customRole.Name)
+	}
+
+	return err
+}
+
+func mapCustomRolesByName(customRoles []v1.CustomRole) map[string]v1.CustomRole {
```
Sugar-pack, line 199: I'm not sure that using the name as the identifier is a good idea. For example, if I create a project with such spec:
```
spec:
  customRoles:
    - actions:
        - name: LIST_SESSIONS
          resources:
            - cluster: true
        - name: KILL_ANY_SESSION
          resources:
            - cluster: true
      inheritedRoles:
        - database: admin
          name: enableSharding
        - database: admin
          name: backup
      name: ShardingAdmin
    - inheritedRoles:
        - database: test
          name: readWrite
        - database: test
          name: dbAdmin
      name: ShardingAdmin
  maintenanceWindow: {}
  name: Test project 1
  withDefaultAlertsSettings: true
```
The operator will create only one role.

pkg/controller/atlasproject/custom_roles.go
Sugar-pack: For two non-valid custom roles CR still contains only one error. 
```
status:
  authModes:
    - SCRAM
  conditions:
    - lastTransitionTime: '2022-11-02T06:16:13Z'
      status: 'False'
      type: Ready
    - lastTransitionTime: '2022-10-26T06:25:35Z'
      status: 'True'
      type: ValidationSucceeded
    - lastTransitionTime: '2022-11-01T12:52:17Z'
      status: 'True'
      type: ProjectReady
    - lastTransitionTime: '2022-11-02T06:16:20Z'
      message: >-
        POST
        https://cloud-qa.mongodb.com/api/atlas/v1.0/groups/635fd15382bd8930618b378c/customDBRoles/roles:
        400 (request "ATLAS_CUSTOM_ROLE_INVALID_NAME") The role name,
        "ShardingAdmin@@@@", is invalid. A custom role name must be a non-empty
        string made up of only ASCII letters, numbers, hyphens, and underscores,
        and should begin with a letter or number.
      reason: ProjectCustomRolesReady
      status: 'False'
      type: ProjectCustomRolesReady
  id: 635fd15382bd8930618b378c
  observedGeneration: 7
spec:
  customRoles:
    - actions:
        - name: LIST_SESSIONS
          resources:
            - cluster: true
        - name: KILL_ANY_SESSION
          resources:
            - cluster: true
      inheritedRoles:
        - database: admin
          name: enableSharding
        - database: admin
          name: backup
      name: ShardingAdmin@@@@
    - inheritedRoles:
        - database: test
          name: readWrite
        - database: test
          name: dbAdmin
      name: ShardingAdmin@@
  maintenanceWindow: {}
  name: Test project 1
  withDefaultAlertsSettings: true

```

Sugar-pack: The status for each role isn't just about complex entities. If any of the roles is invalid, then the CR clearly shows which roles are created and which are not. For example:
CR spec contains roles:
```
role1 - non-valid name
role2 - non-valid action
role3 - valid
```
The current approach:
`role1 contains a non-valid name, and role2 contains a non-valid action`
Approach with status for each role:
```
role1 - failed: contains a non-valid name
role2 - failed: contains a non-valid action
role3 - created
```
The user is explicitly told the status of each role. And user sees that role3  has been created. But in the current approach, the user needs to know additional information about our operator. Moreover, all other features work through individual statuses for each element.

pkg/controller/atlasproject/custom_roles.go
igor-karpukhin: @Sugar-pack @helderjs  I think it make sense to mention that the role names should be unique. In this case the operator may show an error in the status, saying duplicates are not allowed.

pkg/controller/atlasproject/custom_roles.go
helderjs: @Sugar-pack While I agree the status for each role is easier to read, I believe that the API and the entity(custom role) itself don't provide enough information to make a clear set of infos which represents the current status of a certain custom role.
what would be the status when updating it? what if we update it again and again? and what if we delete it?

maybe @igor-karpukhin and @fabritsius have an opinion here?

pkg/controller/atlasproject/custom_roles.go
helderjs: @Sugar-pack @igor-karpukhin As the name is unique and it's the identifier, wouldn't be better to add this rule as validation?

Sugar-pack (approved):
Looks great!

fabritsius (approved):
Looks Great 👍

igor-karpukhin (approved):
Good job 👍

PR #767, opened 2022-11-14
CLOUDP-142977: Atlas Teams

helderjs:
### All Submissions:


pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func createTeamContextFromParent(
+	team *v1.AtlasTeam,
+	kubeClient client.Client,
+	atlasConnection atlas.Connection,
+	atlasDomain string,
+	logger *zap.SugaredLogger,
+) (*workflow.Context, error) {
+	teamCtx := customresource.MarkReconciliationStarted(kubeClient, team, logger)
```
igor-karpukhin, line 91: I guess it should be first checked if this resource is not marked with `mongodb.com/atlas-reconciliation-policy: skip` label

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -0,0 +1,265 @@
...
+
+	atlasTeam, err = renameTeam(ctx, workflowCtx, atlasTeam, team.Spec.Name)
+	if err != nil {
+		return "", workflow.Terminate(workflow.TeamNotUpdatedInAtlas, err.Error())
+	}
+
+	return atlasTeam.ID, workflow.OK()
+}
+
+func ensureTeamUsesAreInSync(ctx context.Context, workflowCtx *workflow.Context, teamID string, team *v1.AtlasTeam) workflow.Result {
```
igor-karpukhin, line 145:
```suggestion
func ensureTeamUsersAreInSync(ctx context.Context, workflowCtx *workflow.Context, teamID string, team *v1.AtlasTeam) workflow.Result {
```

pkg/controller/atlasproject/teams.go
```diff
@@ -0,0 +1,253 @@
...
+			r.Log, resourcesToWatch...,
+		)
+		r.Log.Debugf("watching team resources: %v\r\n", r.WatchedResources)
+	}()
+
+	teamsToAssign := map[string]*v1.Team{}
+	for _, assignedTeam := range project.Spec.Teams {
+		assignedTeam := assignedTeam
+
+		if assignedTeam.TeamRef.Name == "" {
```
igor-karpukhin, line 42: In this case, will the `AtlasTeam` resource still be watched by the controller?

igor-karpukhin (commented):
Massive work here 👍 Added a few comments

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -0,0 +1,265 @@
+package atlasproject
```
Sugar-pack, line 1: Why is the team reconciler in the project package?

pkg/controller/atlasproject/teams.go
```diff
@@ -0,0 +1,253 @@
...
+		r.EnsureMultiplesResourcesAreWatched(
+			types.NamespacedName{Namespace: project.Namespace, Name: project.Name},
+			r.Log, resourcesToWatch...,
+		)
+		r.Log.Debugf("watching team resources: %v\r\n", r.WatchedResources)
+	}()
+
+	teamsToAssign := map[string]*v1.Team{}
+	for _, assignedTeam := range project.Spec.Teams {
+		assignedTeam := assignedTeam
```
Sugar-pack, line 40: I think that it's better to rename this var.

test/e2e/actions/actions.go
```diff
@@ -90,3 +90,27 @@ func DeleteFirstUser(data *model.TestDataProvider) {
 		data.Users = data.Users[1:]
 	})
 }
+
+func AddTeamResourcesWithNUsers(data *model.TestDataProvider, teams []v1.Team, n int) {
+	By("Setup Teams", func() {
+		aClient := atlas.GetClientOrFail()
+		users, err := aClient.GetOrgUsers(data.Project.ID())
```
Sugar-pack, line 97: Will it work if the organization does not have the required number of users?

pkg/controller/atlasproject/team_reconciler.go
igor-karpukhin: It's the same thing as for `BackupSchedules` and `BackupPolicies`. Atlas teams CR changes processed only if they referenced by an `AtlasProject` CR

pkg/api/v1/status/project_team.go
```diff
@@ -0,0 +1,14 @@
+package status
+
+import "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/common"
+
+type ProjectTeamStatus struct {
+	Teams  []ProjectTeamRef `json:"teams"`
+	Status bool             `json:"status"`
```
fabritsius, line 7: As global status can be deduced from the condition we might want to restructure this section and make this status a list of objects with references and separate statuses for each one 💡

pkg/controller/atlasproject/teams.go
helderjs: @Sugar-pack do you think is really needed? the reassignment is only to avoid implicit memory aliasing

pkg/api/v1/status/project_team.go
helderjs: I thought about this initially, but it wouldn't make sense once the assignment to the project is a batch and not an individual operation. Maybe make sense to remove this status flag from here. what do you think @fabritsius?

pkg/controller/atlasproject/teams.go
Sugar-pack: Yeah. I think that it's better to avoid shadowing. But it's up to you.

Sugar-pack (approved):
Great work!

pkg/api/v1/status/project_team.go
fabritsius: Yeah, since we have a condition I think it's ok to remove it

igor-karpukhin (approved):
Looks good 👍 @helderjs please also create a PR for `helm-chart` repo to add new `atlasteams` CRD (https://github.com/mongodb/helm-charts/tree/main/charts/atlas-operator-crds)

fabritsius (approved):
Great Work 👍

PR #773, opened 2022-11-17
Make sure PEs are always added to status

fabritsius:
When there is an error with PEs the **privateEndpoints** status is missing. This change fixes that.


Sugar-pack (approved):
LGTM

PR #778, opened 2022-11-24
CLOUDP-146481: Convert OplogMinRetentionHours field properly

fabritsius:


pkg/api/v1/atlasdeployment_types.go
```diff
@@ -304,10 +307,28 @@ type ProcessArgs struct {
...
 
+func convertOplogMinRetentionHours(specArgs *ProcessArgs, atlasArgs *mongodbatlas.ProcessArgs) error {
+	if specArgs != nil && specArgs.OplogMinRetentionHours != "" {
+		OplogMinRetentionHours, err := strconv.ParseFloat(specArgs.OplogMinRetentionHours, 64)
+		if err != nil {
+			return err
+		}
+
+		atlasArgs.OplogMinRetentionHours = &OplogMinRetentionHours
+		specArgs.OplogMinRetentionHours = ""
```
helderjs, line 326: Don't you think changing the spec value in that way may lead to side effects? Someone working in this field would not expect it was altered after calling `ToAtlas()`

helderjs (changes requested):

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -328,21 +349,30 @@ func (specArgs ProcessArgs) IsEqual(newArgs interface{}) bool {
...
 			return false
 		}
 	}
 
 	return true
 }
 
+func stringValue(v interface{}) string {
+	return regexp.
+		MustCompile(`\.[0]*$`).
```
igor-karpukhin, line 372: `MustCompile` can be called once, not every time you call this method

igor-karpukhin (approved):
Looks good, with one minor thing with regex

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -304,10 +307,28 @@ type ProcessArgs struct {
 
 func (specArgs ProcessArgs) ToAtlas() (*mongodbatlas.ProcessArgs, error) {
```
Sugar-pack, line 308: Why do we use value receiver here? Autogenerated methods use pointer receiver. I might be a bit confusing

pkg/api/v1/atlasdeployment_types.go
fabritsius: Good point 👌

pkg/api/v1/atlasdeployment_types.go
fabritsius: Yeah, I agree that is should fine in this case as the changes only affect `ToAtlas()` scope.

helderjs (approved):
Good job! :+1:

Sugar-pack (approved):

PR #862, opened 2023-01-31
CLOUDP-155410: Fix operator bug with existing m0 cluster

Sugar-pack:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasdeployment/advanced_deployment.go
+++ b/pkg/controller/atlasdeployment/advanced_deployment.go
@@ -15,6 +15,8 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/compat"
 )
 
+const FreeTier = "M0"
+
 func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.Context, project *mdbv1.AtlasProject, deployment *mdbv1.AtlasDeployment) (*mongodbatlas.AdvancedCluster, workflow.Result) {
 	advancedDeploymentSpec := deployment.Spec.AdvancedDeploymentSpec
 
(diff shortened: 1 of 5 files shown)


pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -210,12 +212,16 @@ func handleAutoscaling(ctx *workflow.Context, desiredDeployment *mdbv1.AdvancedD
 
 // MergedAdvancedDeployment will return the result of merging AtlasDeploymentSpec with Atlas Advanced Deployment
 func MergedAdvancedDeployment(atlasDeploymentAsAtlas mongodbatlas.AdvancedCluster, specDeployment mdbv1.AdvancedDeploymentSpec) (mergedDeployment mdbv1.AdvancedDeploymentSpec, atlasDeployment mdbv1.AdvancedDeploymentSpec, err error) {
+	if IsFreeTierAdvancedDeployment(&atlasDeploymentAsAtlas) {
+		atlasDeploymentAsAtlas.DiskSizeGB = nil
```
fabritsius, line 216: Just in case it is possible to have `512 MB` for this field in places other than free tier I suggest to actually check the value and make the decision based on that. We technically don't even need to check for the actual `InstanceSize` value.

```
if spec == nil && atlas < 1 { make nil }
```

Sugar-pack: M0 supports only 512mb disk size

fabritsius: Ok:)

fabritsius (approved):
Looks Great 👍

igor-karpukhin (approved):

PR #943, opened 2023-04-26
CLOUDP-165326: Add data federation

igor-karpukhin:
### All Submissions:

Add data federation and private endpoints feature


pkg/controller/atlasdatafederation/private_endpoint.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return workflow.OK()
+}
+
+func syncPrivateEndpointsWithAtlas(ctx *workflow.Context, clientDF *DataFederationServiceOp, projectID string, specPEs, atlasPEs []mdbv1.DataFederationPE) workflow.Result {
+	endpointsToCreate := set.Difference(specPEs, atlasPEs)
+	ctx.Log.Debugw("Data Federation PEs to Create", "endpoints", endpointsToCreate)
+	for _, endpoint := range endpointsToCreate {
+		endpointID := endpoint.Identifier().(string)
+		if _, _, err := clientDF.DeleteOnePrivateEndpoint(context.Background(), projectID, endpointID); err != nil {
```
helderjs, line 37: Are you iterating over EP to create but calling the deletion func

pkg/controller/atlasdatafederation/private_endpoint.go
```diff
@@ -0,0 +1,60 @@
...
+			return workflow.Terminate(workflow.Internal, err.Error())
+		}
+	}
+
+	endpointsToDelete := set.Difference(atlasPEs, specPEs)
+	ctx.Log.Debugw("Data Federation PEs to Delete", "endpoints", endpointsToDelete)
+	for _, item := range endpointsToDelete {
+		endpoint := *item.(*mdbv1.DataFederationPE)
+		if _, _, err := clientDF.CreateOnePrivateEndpoint(context.Background(), projectID, endpoint); err != nil {
```
helderjs, line 46: Are you iterating over EP to delete but calling the creation func

helderjs (changes requested):
Some initial suggestions

pkg/api/v1/datafederation_types.go
```diff
@@ -0,0 +1,208 @@
...
+	Region                   string   `json:"region,omitempty"`
+}
+
+type DataFederationPE struct {
+	EndpointID string `json:"endpointId,omitempty"`
+	Provider   string `json:"provider,omitempty"`
+	Type       string `json:"type,omitempty"`
+}
+
+func (pe DataFederationPE) Identifier() interface{} {
```
helderjs, line 105: Why does it need to return `interface{}` instead of string?

pkg/controller/atlasdatafederation/datafederation.go
```diff
@@ -0,0 +1,87 @@
...
+func dataFederationEqual(atlasSpec, operatorSpec mdbv1.DataFederationSpec, log *zap.SugaredLogger) (areEqual bool, diff string) {
+	mergedSpec, err := getMergedSpec(atlasSpec, operatorSpec)
+	if err != nil {
+		log.Errorf("failed to merge Data Federation specs: %s", err.Error())
+		return false, ""
+	}
+
+	d := cmp.Diff(atlasSpec, mergedSpec, cmpopts.EquateEmpty())
+	if d != "" {
+		fmt.Println("DEBUG DIFF", d)
```
helderjs, line 65: remove debug code please

pkg/api/v1/datafederation_types.go
helderjs: Saw later it's because of the `uitl.set` package, maybe worth to make the package generic?

igor-karpukhin: Yes, good idea. But better to do that in a separate PR.

pkg/controller/atlasdatafederation/connectionsecrets.go
```diff
@@ -0,0 +1,104 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/status"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/connectionsecret"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/workflow"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/stringutil"
+)
+
+func (r *AtlasDataFederationReconciler) ensureConnectionSecrets(ctx *workflow.Context, project *mdbv1.AtlasProject, df *mdbv1.AtlasDataFederation) workflow.Result {
```
helderjs, line 18: This is basically the same as https://github.com/mongodb/mongodb-atlas-kubernetes/blob/2ce7fee74e8426091f6b6aa8e2f89730af2e5dbc/pkg/controller/atlasdeployment/advanced_deployment.go#L397

What do you think about abstracting it, passing the connectionStrings/hostnames, moving it to `connectionsecret` package, and reuse in both places?

pkg/controller/atlasdatafederation/datafederation_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	err = c.Watch(&source.Kind{Type: &mdbv1.AtlasDataFederation{}}, &handler.EnqueueRequestForObject{}, r.GlobalPredicates...)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Delete implements a handler for the Delete event. (can't be done now with connection secrets)
+func (r *AtlasDataFederationReconciler) Delete(e event.DeleteEvent) error {
```
helderjs, line 239: Never understood why we need this. Can't a`deleteConnectionSecrets()` be called just before deleting the data federation (line 148)?

pkg/controller/atlasdatafederation/datafederation_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	customresource.UnsetFinalizer(df, customresource.FinalizerLabel)
+	if err = r.Client.Update(context, df); err != nil {
+		return fmt.Errorf("failed to remove deletion finalizer from %s: %w", df.GetName(), err)
+	}
+	return nil
+}
+
+func (r *AtlasDataFederationReconciler) deleteDataFederationFromAtlas(ctx context.Context, df *mdbv1.AtlasDataFederation, project *mdbv1.AtlasProject, log *zap.SugaredLogger) error {
+	log.Infof("Deleting DataFederation instance: %s from Atlas", df.Spec.Name)
+	connection, err := atlas.ReadConnection(log, r.Client, r.GlobalAPISecret, project.ConnectionSecretObjectKey())
```
helderjs, line 184: Why do we need to recreate the client if it's available in the caller and can be passed to this function?

test/int/datafederation_test.go
```diff
@@ -0,0 +1,186 @@
...
+						{
+							Name: "test-db-1",
+							Collections: []mdbv1.Collection{
+								{
+									Name: "test-collection-1",
+									DataSources: []mdbv1.DataSource{
+										{
+											StoreName: "http-test",
+											Urls: []string{
+												"https://data.cityofnewyork.us/api/views/vfnx-vebw/rows.csv",
```
helderjs, line 134: I think would be good to have the CSV offline, upload it to a S3 bucket and use it on the tests, we will neve now when this would be unavailable and break the test. Can be done a improvement.

test/int/datafederation_test.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			By("Removing Atlas Project " + createdProject.Status.ID)
+			Expect(k8sClient.Delete(context.Background(), createdProject)).To(Succeed())
+			Eventually(checkAtlasProjectRemoved(createdProject.Status.ID), 60, interval).Should(BeTrue())
+		}
+		removeControllersAndNamespace()
+	})
+
+	Describe("DataFederation can be created with stores and private endpoints", func() {
```
helderjs, line 103: Couldn't find anything related to the private endpoint, is it pending?

helderjs (changes requested):

pkg/controller/atlasdatafederation/connectionsecrets.go
igor-karpukhin: They are not the same. Yes, this can be extracted and reused, but the method would include a lot more parameters, like event recorder, k8s client, the actual object, etc.

pkg/controller/atlasdatafederation/datafederation_controller.go
igor-karpukhin: This is due to this: "EventHandlerWithDelete is an extension of EnqueueRequestForObject that will _not_ trigger a reconciliation for a Delete event. Instead, it will call an external controller's Delete() method and pass the event argument unchanged."

pkg/controller/atlasdatafederation/datafederation_controller.go
igor-karpukhin: Good point, fixed!

PR #971, opened 2023-05-29
Fixed integrations secret namespace

igor-karpukhin:
### All Submissions:

Fixed integrations secret namespace

### DIFF
--- a/pkg/api/v1/common/common.go
+++ b/pkg/api/v1/common/common.go
@@ -41,33 +41,34 @@ type LabelSpec struct {
 	Value string `json:"value"`
 }
 
-func (rn *ResourceRefNamespaced) GetObject(parentNamespace string) *client.ObjectKey {
+func (rn *ResourceRefNamespaced) GetObject(fallbackNamespace string) *client.ObjectKey {
 	if rn == nil {
 		return nil
 	}
 
-	ns := SelectNamespace(rn.Namespace, parentNamespace)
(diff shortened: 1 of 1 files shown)


pkg/api/v1/project/integration.go
```diff
@@ -82,7 +82,7 @@ func (i Integration) ToAtlas(c client.Client, defaultNS string) (result *mongodb
 			return
 		}
 
-		*target, err = passwordField.ReadPassword(c, defaultNS)
```
helderjs, line 85: I think you need to change the `ReadPassword()` behavior. It was supposed to fallback to `defaultNS` when the namespace of the requested resource it's not available, however, it's actually uses the passed namespace

helderjs (changes requested):

pkg/api/v1/project/integration.go
igor-karpukhin: Good point, thanks! Fixed

helderjs: There's still something strange. Look at `GetObject()`, it is supposed to already return the correct object key(falling back to parent namespace) but seems to not work correctly.

helderjs (changes requested):

pkg/api/v1/project/integration.go
igor-karpukhin: Yeah, looks it works the other way around. Let me fix this

PR #1003, opened 2023-06-22
CLOUDP-178751 - DatabaseUser deletion protection

helderjs:
… controller

### All Submissions:


cmd/manager/main.go
```diff
@@ -303,3 +314,46 @@ func initCustomZapLogger(level, encoding string) (*zap.Logger, error) {
...
+			objectDeletionSet = true
+		}
+
+		if f.Name == "subobject-deletion-protection" {
+			subObjectDeletionSet = true
+		}
+	})
+
+	if !objectDeletionSet {
+		objDeletion := os.Getenv("OBJECT_DELETION_PROTECTION")
```
josvazg, line 337: Nit
Just curious: should we be case insensitive here?
```suggestion
		objDeletion := strings.ToLower(os.Getenv("OBJECT_DELETION_PROTECTION"))
```

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -78,155 +76,207 @@ func (r *AtlasDatabaseUserReconciler) Reconcile(context context.Context, req ctr
...
 	}
 
-	ctx.SetConditionTrue(status.DatabaseUserReadyType)
-	ctx.SetConditionTrue(status.ReadyType)
-	return result.ReconcileResult(), nil
-}
+	if !owner {
+		result = workflow.Terminate(
+			workflow.AtlasDeletionProtection,
+			"unable to reconcile database user: it already exists in Atlas, it was not previously managed by the operator, de deletion protection is enabled.",
```
josvazg, line 144: Typo?
```suggestion
			"unable to reconcile database user: it already exists in Atlas, it was not previously managed by the operator, and the deletion protection is enabled.",
```

cmd/manager/main.go
```diff
@@ -303,3 +314,46 @@ func initCustomZapLogger(level, encoding string) (*zap.Logger, error) {
...
+func configureDeletionProtectionFlags(config *Config) {
+	if config == nil {
+		return
+	}
+
+	objectDeletionSet := false
+	subObjectDeletionSet := false
+
+	flag.Visit(func(f *flag.Flag) {
+		if f.Name == "object-deletion-protection" {
```
igor-karpukhin, line 327: Consider making names of the flags `const`. I see this is also used in the flag parsing method

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -78,155 +76,212 @@ func (r *AtlasDatabaseUserReconciler) Reconcile(context context.Context, req ctr
...
+	result = r.ensureDatabaseUser(workflowCtx, *project, *databaseUser)
+	if !result.IsOk() {
+		workflowCtx.SetConditionFromResult(status.DatabaseUserReadyType, result)
+
+		return result.ReconcileResult(), nil
 	}
 
-	return nil
-}
+	err = customresource.ManageFinalizer(ctx, r.Client, databaseUser, customresource.SetFinalizer)
```
igor-karpukhin, line 173: non-blocking: consider consistent naming: `EnsureFinalizer`

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -78,155 +76,212 @@ func (r *AtlasDatabaseUserReconciler) Reconcile(context context.Context, req ctr
...
+	return true, workflow.OK()
+}
+
+func (r *AtlasDatabaseUserReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("AtlasDatabaseUser").
+		For(&mdbv1.AtlasDatabaseUser{}, builder.WithPredicates(r.GlobalPredicates...)).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, watch.NewSecretHandler(r.WatchedResources)).
+		Complete(r)
+}
```
igor-karpukhin, line 248: Are you sure removing works fine in this case? I mean, won't it trigger the reconcile with result like "can not reconcile, the resource has been modified/deleted`?

igor-karpukhin (changes requested):
Looks good, massive job here 👍 However, let's not merge it to "main", but into a separate branch ("deletion-protection", maybe), so we won't break "main".

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
helderjs: As it's not a controller function but actually a helper, I think there's no inconsistency with naming

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
helderjs: Yes, All the flows has worked as expected

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -78,155 +76,212 @@ func (r *AtlasDatabaseUserReconciler) Reconcile(context context.Context, req ctr
...
-	return workflow.OK()
-}
 
-func (r *AtlasDatabaseUserReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	c, err := controller.New("AtlasDatabaseUser", mgr, controller.Options{Reconciler: r})
-	if err != nil {
-		return err
+	deletionRequest, result := r.handleDeletion(ctx, databaseUser, project, atlasClient, log)
+	if deletionRequest {
+		return result.ReconcileResult(), nil
```
josvazg, line 154: I am having difficulties following the logic of this reconcile function.
Why do we handle deletion here? I do not see where we are checking that a deletion is needed or requested.

pkg/controller/customresource/protection.go
```diff
@@ -0,0 +1,66 @@
...
+
+func IsOwner(resource mdbv1.AtlasCustomResource, protectionFlag bool, operatorChecker OperatorChecker, atlasChecker AtlasChecker) (bool, error) {
+	if !protectionFlag {
+		return true, nil
+	}
+
+	wasManaged, err := operatorChecker(resource)
+	if err != nil {
+		return false, err
+	}
```
josvazg, line 28: If it was managed we do not need to go to atlas right? we can just return true here.

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
helderjs: Here is where it's validated

https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1003/files#diff-b936aa2d7b65922a3ed27d3debce3bf9cb8848177e170dbf3294d4157ee22c12R202-R204

pkg/controller/customresource/protection.go
helderjs: Make sense

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: ok, I guess the confusion was due to the name of the method, `handleDeletion` makes you think that deletion is already decided.

igor-karpukhin (approved):

PR #1008, opened 2023-06-26
CLOUDP-167207: Validate GCP Service Account Key

josvazg:
### All Submissions:

### DIFF
--- a/pkg/controller/validate/validate.go
+++ b/pkg/controller/validate/validate.go
@@ -1,9 +1,14 @@
 package validate
 
 import (
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
+	"net/url"
 	"reflect"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -12,6 +17,20 @@ import (
 
 	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
 )
+
+type googleServiceAccountKey struct {
+	Type                    string `json:"type"`
+	ProjectID               string `json:"project_id"`
+	PrivateKeyID            string `json:"private_key_id"`
+	PrivateKey              string `json:"private_key"`  // Expects valid PEM key
+	ClientEmail             string `json:"client_email"` // Expects a valid email
+	ClientID                string `json:"client_id"`
+	AuthURI                 string `json:"auth_uri"`                    // Expects valid URL
+	TokenURI                string `json:"token_uri"`                   // Expects valid URL
+	AuthProviderX509CertURL string `json:"auth_provider_x509_cert_url"` // Expects valid URL
+	ClientX509CertURL       string `json:"client_x509_cert_url"`        // Expects valid URL
+	UniverseDomain          string `json:"universe_domain"`
+}
 
 func DeploymentSpec(deploymentSpec mdbv1.AtlasDeploymentSpec) error {
 	var err error
@@ -50,6 +69,10 @@ func DeploymentSpec(deploymentSpec mdbv1.AtlasDeploymentSpec) error {
 
 func Project(project *mdbv1.AtlasProject) error {
 	if err := projectCustomRoles(project.Spec.CustomRoles); err != nil {
+		return err
+	}
+
+	if err := encryptionAtRest(project.Spec.EncryptionAtRest); err != nil {
 		return err
 	}
 
@@ -194,5 +217,60 @@ func projectCustomRoles(customRoles []mdbv1.CustomRole) error {
 		customRolesMap[customRole.Name] = struct{}{}
 	}
 
+	return err
+}
+
+func encryptionAtRest(encryption *mdbv1.EncryptionAtRest) error {
+	if encryption != nil &&
+		encryption.GoogleCloudKms.Enabled != nil &&
+		*encryption.GoogleCloudKms.Enabled {
+		if encryption.GoogleCloudKms.ServiceAccountKey == "" {
+			return fmt.Errorf("missing Google Service Account Key but GCP KMS is enabled")
+		}
+		if err := gceServiceAccountKey(encryption.GoogleCloudKms.ServiceAccountKey); err != nil {
+			return fmt.Errorf("failed to validate Google Service Account Key: %w", err)
+		}
+	}
+	return nil
+}
+
+func unfilter(key string) string {
+	return strings.ReplaceAll(key, "\\\\n", "\\n")
+}
+
+func gceServiceAccountKey(key string) error {
+	emptyKey := googleServiceAccountKey{}
+	gceSAKey := googleServiceAccountKey{}
+	if err := json.Unmarshal(([]byte)(unfilter(key)), &gceSAKey); err != nil {
+		return fmt.Errorf("invalid service account key format: %w", err)
+	}
+	if emptyKey == gceSAKey {
+		return fmt.Errorf("invalid empty service account key")
+	}
+	for _, rawURL := range []string{gceSAKey.AuthURI,
+		gceSAKey.TokenURI,
+		gceSAKey.ClientX509CertURL,
+		gceSAKey.AuthProviderX509CertURL} {
+		if _, err := url.ParseRequestURI(rawURL); err != nil {
+			return fmt.Errorf("invalid URL address %q: %w", rawURL, err)
+		}
+	}
+	block, _ := pem.Decode([]byte(gceSAKey.PrivateKey))
+	if block == nil || !strings.HasSuffix(block.Type, "PRIVATE KEY") {
+		return fmt.Errorf("failed to decode PEM block containing a private key")
+	}
+
+	err := assertParsePrivateKey(block.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse PEM private key: %w", err)
+	}
+	return nil
+}
+
+func assertParsePrivateKey(key []byte) error {
+	_, err := x509.ParsePKCS1PrivateKey(key)
+	if err != nil && strings.Contains(err.Error(), "ParsePKCS8PrivateKey") {
+		_, err = x509.ParsePKCS8PrivateKey(key)
+	}
 	return err
 }
--- a/test/e2e/encryption_at_rest_test.go
+++ b/test/e2e/encryption_at_rest_test.go
@@ -1,16 +1,25 @@
 package e2e_test
 
 import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"go.mongodb.org/atlas/mongodbatlas"
+	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/types"
 
 	v1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/status"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/workflow"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/testutil"
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/toptr"
 	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/actions"
 	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/actions/cloud"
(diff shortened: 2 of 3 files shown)


pkg/controller/validate/validate.go
```diff
@@ -196,3 +234,46 @@ func projectCustomRoles(customRoles []mdbv1.CustomRole) error {
 
 	return err
 }
+
+func encryptionAtRest(project *mdbv1.AtlasProject) error {
+	if project.Spec.EncryptionAtRest != nil &&
+		project.Spec.EncryptionAtRest.GoogleCloudKms.Enabled != nil &&
+		*project.Spec.EncryptionAtRest.GoogleCloudKms.Enabled == true &&
+		project.Spec.EncryptionAtRest.GoogleCloudKms.ServiceAccountKey != "" {
```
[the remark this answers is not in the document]
josvazg: You mean return error when `Enabled = true` and `ServiceAccountKey == ""` ?

pkg/controller/validate/validate.go
```diff
@@ -13,6 +18,35 @@ import (
...
+	  "client_email": "SERVICE_ACCOUNT_EMAIL",
+	  "client_id": "CLIENT_ID",
+	  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
+	  "token_uri": "https://accounts.google.com/o/oauth2/token",
+	  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
+	  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/SERVICE_ACCOUNT_EMAIL"
+	  "universe_domain": "googleapis.com"
+	}
+*/
+type googleServiceAccountKey struct {
```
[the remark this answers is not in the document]
josvazg: If we do have that dependency I will check and use if it exists. If we do not have it, I prefer not to bring a new dependency on just for a single struct.

pkg/controller/validate/validate.go
```diff
@@ -196,3 +234,46 @@ func projectCustomRoles(customRoles []mdbv1.CustomRole) error {
...
+		return fmt.Errorf("invalid empty service account key")
+	}
+	for _, rawURL := range []string{gceSAKey.AuthURI,
+		gceSAKey.TokenURI,
+		gceSAKey.ClientX509CertURL,
+		gceSAKey.AuthProviderX509CertURL} {
+		if _, err := url.ParseRequestURI(rawURL); err != nil {
+			return fmt.Errorf("invalid URL address %q: %w", rawURL, err)
+		} else {
+			fmt.Println("url=", rawURL, "err=", err)
```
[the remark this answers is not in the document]
josvazg: Because I forgot to clean up traces!
Thanks for the heads up, will remove.

josvazg:
> Overall looks good, just have a few comments and 2 extra thoughts:
> 
> 1. Does it make sense to have an e2e test to see how an escaped string from an applied yaml would go through the validation?
> 2. Talking to @igor-karpukhin, we were wondering if makes sense to validate the entire service account object, once Atlas API will check it is really valid and can be used to connect to GCPP.

I am doing some manual testing at the moment before I decide how to proceed there. I am hesitant to add up more to our slow and flaky e2e tests, unless for a very good reason that has no better alternative.

I think the validation against the API was not the initial idea. We could do that, even before we deploy anything in the cloud. And that could be a fast e2e test as it is a single check and does not require a cloud resource deployment to take place beforehand.

pkg/controller/validate/validate.go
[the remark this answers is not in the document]
josvazg: fixed

pkg/controller/validate/validate.go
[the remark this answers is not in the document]
josvazg: I think it is done, feel free to reopen or comment if it is not.

pkg/controller/validate/validate.go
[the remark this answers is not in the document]
josvazg: I believe the package that would have that would be: `google.golang.org/api/iam/v1`
https://pkg.go.dev/google.golang.org/api/iam/v1#ServiceAccountKey

I do not see us already using it so, rather than importing it just for this I would rather copy the struct or the parts of it that would make sense.

josvazg:
> Overall looks good, just have a few comments and 2 extra thoughts:
> 
> 1. Does it make sense to have an e2e test to see how an escaped string from an applied yaml would go through the validation?
> 2. Talking to @igor-karpukhin, we were wondering if makes sense to validate the entire service account object, once Atlas API will check it is really valid and can be used to connect to GCPP.

In the end I added the same tests as e2e test as well, but the happy path case is using the format that we can recommend to end users when setting up the keys, just take the service account JSON file from Google and put it, as is, in the CRD [YAML as a multiline attribute](https://yaml-multiline.info/):

```yaml
apiVersion: atlas.mongodb.com/v1
kind: AtlasProject
metadata:
  name: my-project
spec:
  name: Test Atlas Operator Project
  encryptionAtRest:
    googleCloudKms:
      enabled: true
      keyVersionResourceID: projects/...
      serviceAccountKey: |
        {
          "type": "service_account",
          "project_id": "...",
          ...
        }
```
The only caveat is to **respect the indentation**, as seen in the example, so that the YAML format knows this file is a document for  `serviceAccountKey`

.github/workflows/test-e2e.yml
```diff
@@ -138,6 +138,7 @@ jobs:
             "deployment-ns",
             "deployment-wide",
             "encryption-at-rest",
+            "encryption-at-rest-gcp",
```
helderjs, line 141: Why separate it from other EaR tests?

pkg/controller/validate/validate.go
```diff
@@ -196,3 +219,57 @@ func projectCustomRoles(customRoles []mdbv1.CustomRole) error {
 
 	return err
 }
+
+func encryptionAtRest(project *mdbv1.AtlasProject) error {
```
helderjs, line 223: Would be better to pass only `EncryptionAtRest` struct? I think it would also make the evaluation be less verbose

pkg/controller/validate/validate.go
```diff
@@ -196,3 +219,57 @@ func projectCustomRoles(customRoles []mdbv1.CustomRole) error {
...
+	if project.Spec.EncryptionAtRest != nil &&
+		project.Spec.EncryptionAtRest.GoogleCloudKms.Enabled != nil &&
+		*project.Spec.EncryptionAtRest.GoogleCloudKms.Enabled {
+		if project.Spec.EncryptionAtRest.GoogleCloudKms.ServiceAccountKey == "" {
+			return fmt.Errorf("missing Google Service Account Key but GCP KMS is enabled")
+		} else if err := gceServiceAccountKey(project.Spec.EncryptionAtRest.GoogleCloudKms.ServiceAccountKey); err != nil {
```
helderjs, line 229:
```suggestion
		} 
		
		if err := gceServiceAccountKey(project.Spec.EncryptionAtRest.GoogleCloudKms.ServiceAccountKey); err != nil {
```

test/e2e/encryption_at_rest_test.go
```diff
@@ -294,3 +302,223 @@ var _ = Describe("Encryption at rest AWS", Label("encryption-at-rest"), func() {
...
+  name: my-project
+spec:
+  name: Test Atlas Operator Project
+  encryptionAtRest:
+    googleCloudKms:
+      enabled: true
+      keyVersionResourceID: %s
+      serviceAccountKey: %s`
+
+// composeGoogleEncryptionAtRestProjectYAML produces something like this YAML:
```
helderjs, line 405: Should we remove those commented block?

test/e2e/encryption_at_rest_test.go
```diff
@@ -294,3 +302,223 @@ var _ = Describe("Encryption at rest AWS", Label("encryption-at-rest"), func() {
...
+			condition, err := getProjectCondition(
+				testData.Context,
+				testData.K8SClient,
+				status.ValidationSucceeded,
+				testData.Resources.Namespace,
+				testData.Project.GetName(),
+			)
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(condition).ToNot(BeNil())
+		}).WithPolling(5 * time.Second).WithTimeout(40 * time.Second).Should(Succeed())
```
helderjs, line 348: Can't we use this helper function instead of duplicating it?
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/2592e1901e184503c8a1faa7f1d786b14743c7b1/test/e2e/actions/deploy/deploy_operator.go#L48

helderjs (changes requested):

igor-karpukhin (changes requested):
Looks good in general 👍 , but let's not separate this test from other "encryption-at-rest" tests

.github/workflows/test-e2e.yml
josvazg: Ok, separation was more due to flakiness than time though.

pkg/controller/validate/validate.go
josvazg: checking

josvazg: agreed, fixing in next update.

test/e2e/encryption_at_rest_test.go
josvazg: I would leave it, I think it is helpful to quickly visualize what the code is trying to output and this functions is unlikely to change over time.

pkg/controller/validate/validate.go
josvazg: will apply on my next update

pkg/controller/validate/validate.go
```diff
@@ -196,3 +219,57 @@ func projectCustomRoles(customRoles []mdbv1.CustomRole) error {
...
+	if block == nil || !strings.HasSuffix(block.Type, "PRIVATE KEY") {
+		return fmt.Errorf("failed to decode PEM block containing a private key")
+	}
+
+	err := assertParsePrivateKey(block.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse PEM private key: %w", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
josvazg: I think there si value on doing some cheap and fast local checks before going to the cloud. In fact, the Jira task wording was focused on local validation, AFAIU.

test/e2e/encryption_at_rest_test.go
```diff
@@ -294,3 +302,223 @@ var _ = Describe("Encryption at rest AWS", Label("encryption-at-rest"), func() {
...
+	ctx := context.Background()
+	go func(ctx context.Context) context.Context {
+		err := mgr.Start(ctx)
+		Expect(err).NotTo(HaveOccurred())
+		return ctx
+	}(ctx)
+	testData.ManagerContext = ctx
+}
+
+func getProjectCondition(ctx context.Context, k8sClient client.Client, statusType status.ConditionType, ns string, name string) (*status.Condition, error) {
```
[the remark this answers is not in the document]
josvazg: checking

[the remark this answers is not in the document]
josvazg: That does not work for me here. I would need an `actions.WaitForConditionsToAppear` kind of thing.

The reason is this is for checking for particular failures as soon as they appear. So the condition will appear false and I will check whether or not the error is as expected.

test/e2e/encryption_at_rest_test.go
josvazg: `CreateProject` cannot replace `createProjectWithValidationError` because it:
- Will not look for the expected error, and end up as success if that is what happened.
- Will timeout trying to get a successful project created, which we do not expect in these test cases.

josvazg: I do feel that the code for this helper can be simplified somehow, though.

test/e2e/encryption_at_rest_test.go
```diff
@@ -294,3 +302,223 @@ var _ = Describe("Encryption at rest AWS", Label("encryption-at-rest"), func() {
...
+
+	It("fails if the service account key is missing", func() {
+		testData.Project.Spec.EncryptionAtRest = &v1.EncryptionAtRest{
+			GoogleCloudKms: v1.GoogleCloudKms{
+				Enabled:           toptr.MakePtr(true),
+				ServiceAccountKey: "",
+			},
+		}
+
+		configureManager(testData)
```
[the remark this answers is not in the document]
josvazg: ok, coming on next update

pkg/controller/validate/validate.go
josvazg: done

.github/workflows/test-e2e.yml
josvazg: done

pkg/controller/validate/validate.go
josvazg: done

helderjs (changes requested):

test/e2e/encryption_at_rest_test.go
```diff
@@ -294,3 +302,221 @@ var _ = Describe("Encryption at rest AWS", Label("encryption-at-rest"), func() {
...
+		}).WithPolling(5 * time.Second).WithTimeout(40 * time.Second).Should(Succeed())
+		condition, err := getProjectCondition(
+			testData.Context,
+			testData.K8SClient,
+			status.ValidationSucceeded,
+			testData.Resources.Namespace,
+			testData.Project.GetName(),
+		)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(condition.Status)).To(Equal("False"))
```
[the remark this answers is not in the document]
josvazg: done

helderjs (approved):
Good work!

igor-karpukhin (approved):
Looks good 👍

PR #1014, opened 2023-06-27
CLOUDP-186801: disable depguard linter

josvazg:
### All Submissions:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,7 +2,7 @@ linters:
   disable-all: true
   enable:
     #- bodyclose
-    - depguard
+    #- depguard
     - dogsled
     - dupl
     - errcheck


helderjs:
@josvazg We should not disable the depguard linter, but update golangci-lint in the project and adjust its config.

josvazg:
> @josvazg We should not disable the depguard linter, but update golangci-lint in the project and adjust its config.

This change is to make the development environment `make lint` behaviour match the current CI state.

What is the value of having development fail for something the CI is not checking yet?
We can enable it back on the same PR that fixes the config and would make the CI use depguard.

helderjs:
@josvazg Why do you think CI is not checking it?
The latest depguard version has changed its default behavior, requiring to explicitly setting a config for it.
https://github.com/OpenPeeDeeP/depguard/issues/55
https://github.com/OpenPeeDeeP/depguard/issues/49

josvazg:
> 

It seemed so, because `make lint` in development fails with a lot of `depguard` failures. In any case, the point of this PR is to fix the development behavior by default, while the proper fix gets deployed to both dev and the CI.

josvazg:
If I understood correctly, I could alternatively just drop this PR and instead keep on using the same linter version as the CI to have the same results.
I will be closing this then.
Thanks!

PR #1026, opened 2023-07-05
CLOUDP-190400: Merge deletion protection bits

josvazg:
**No need to review, all code here has already been reviewed and merged** in a feature branch.

Feature branch for deletion protection.

After merging #1052 we can now safely merge this branch into `main` and keep adding bits of the deletion protection feature directly into `main`. The feature is disabled by default and can only be enabled on development build for some e2e tests that require it.

### All Submissions:


pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -99,125 +100,155 @@ func (r *AtlasProjectReconciler) Reconcile(context context.Context, req ctrl.Req
...
+				return workflow.Terminate(workflow.AtlasFinalizerNotSet, err.Error())
 			}
 		}
 	}
 
 	if !project.GetDeletionTimestamp().IsZero() {
 		if customresource.HaveFinalizer(project, customresource.FinalizerLabel) {
-			if customresource.ResourceShouldBeLeftInAtlas(project) {
-				log.Infof("Not removing the Atlas Project from Atlas as the '%s' annotation is set", customresource.ResourcePolicyAnnotation)
+			log.Infow("RESOURCE PROTECTED", r.ObjectDeletionProtection, customresource.IsResourceProtected(project, r.ObjectDeletionProtection))
```
igor-karpukhin, line 230: nit: Any reason for upper case info here?

josvazg: Not a special one, seemed to be the norm when PRs were open, seems now it has not been preserved.

I would suggest to clean it up afterwards, this should be a clean rebase of what was already reviewed and merged.

helderjs: This was a debug I put in the code, I'll remove it in the upcoming PRs

Dockerfile
```diff
@@ -27,9 +27,9 @@ ENV TARGET_OS=${TARGETOS}
 
 RUN make manager
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2
```
helderjs, line 30: @igor-karpukhin are you okay with this update?
It would not affect openshift certification process, right?

helderjs (approved):

Dockerfile
igor-karpukhin: @helderjs if it passed the upgrade test, then that's okay. We can always rebuild it with the previous one if it won't pass certification.

PR #1029, opened 2023-07-07
Support Atlas client transport customization

josvazg:
### All Submissions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ lint:
 $(TIMESTAMPS_DIR)/fmt: $(GO_SOURCES)
 	go fmt ./...
 	find . -name "*.go" -not -path "./vendor/*" -exec gofmt -w "{}" \;
-	find . -name "*.go" -not -path "./vendor/*" -exec goimports -l -w "{}" \;
+	find . -name "*.go" -not -path "./vendor/*" -exec goimports -local github.com/mongodb/mongodb-atlas-kubernetes -l -w "{}" \;
 	@mkdir -p $(TIMESTAMPS_DIR) && touch $@
 
 .PHONY: fmt
--- a/pkg/controller/atlas/client.go
+++ b/pkg/controller/atlas/client.go
@@ -15,11 +15,13 @@ import (
 
 // Client is the central place to create a client for Atlas using specified API keys and a server URL.
 // Note, that the default HTTP transport is reused globally by Go so all caching, keep-alive etc will be in action.
-func Client(atlasDomain string, connection Connection, log *zap.SugaredLogger) (mongodbatlas.Client, error) {
+func Client(atlasDomain string, connection Connection, log *zap.SugaredLogger, opts ...httputil.ClientOpt) (mongodbatlas.Client, error) {
 	withDigest := httputil.Digest(connection.PublicKey, connection.PrivateKey)
 	withLogging := httputil.LoggingTransport(log)
+	allOptions := []httputil.ClientOpt{withDigest, withLogging}
+	allOptions = append(allOptions, opts...)
 
-	httpClient, err := httputil.DecorateClient(basicClient(), withDigest, withLogging)
+	httpClient, err := httputil.DecorateClient(basicClient(), allOptions...)
 	if err != nil {
 		return mongodbatlas.Client{}, err
 	}
--- a/pkg/controller/atlas/client_test.go
+++ b/pkg/controller/atlas/client_test.go
@@ -36,7 +36,7 @@ func TestCustomTransport(t *testing.T) {
 	tt := &testTransport{used: false}
 
 	client, err := atlas.Client("https://cloud.mongodb.com", atlas.Connection{}, nil, httputil.CustomTransport(tt))
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	client.Projects.GetAllProjects(context.Background(), &mongodbatlas.ListOptions{})
 	assert.Equal(t, true, tt.used)
 }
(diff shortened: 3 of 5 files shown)


pkg/controller/atlas/client.go
```diff
@@ -13,13 +13,19 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/httputil"
 )
 
+type HTTPClientFn func() *http.Client
+
+// CustomHTTPClientFn allows to override the default basic http.Client,
+// for protocol logging or test mocking
+var CustomHTTPClientFn HTTPClientFn
```
helderjs, line 20: I'm not in favor of adding a global as a way to override a dependency. What do you think about user Functional Options to configure the client?
I understand it will be more work but I believe it's a better approach.

pkg/controller/atlas/client_test.go
```diff
@@ -17,3 +19,24 @@ func TestClientUserAgent(t *testing.T) {
...
+
+func (tcb *TestClientBuilder) NewClientFn() *http.Client {
+	tcb.used = true
+	return &http.Client{Transport: http.DefaultTransport}
+}
+
+func TestCustomClient(t *testing.T) {
+	tcb := &TestClientBuilder{used: false}
+	atlas.CustomHTTPClientFn = tcb.NewClientFn
+	defer func() {
```
helderjs, line 35: The presence of this in your test shows how risky is the current approach.

pkg/controller/atlas/logger.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	responseDump, err := httputil.DumpResponse(rsp, true)
+	if err != nil {
+		return err.Error()
+	}
+	return string(responseDump)
+}
+
+// AtlasLog is all is needed to implement logging
+type AtlasLog interface {
```
helderjs, line 35: As already mentioned and adding what Igor said, you could configure the client to use zap logger to do this job.
I see no need to maintain our custom logger even for test purposes.

helderjs (changes requested):

pkg/controller/atlas/client.go
josvazg: That is cleaner indeed, I will refactor.

pkg/controller/atlas/logger.go
josvazg: Will fix.

pkg/controller/atlas/client.go
josvazg: Done

pkg/controller/atlas/client_test.go
josvazg: fixed

Makefile
```diff
@@ -160,7 +160,7 @@ lint:
 $(TIMESTAMPS_DIR)/fmt: $(GO_SOURCES)
 	go fmt ./...
 	find . -name "*.go" -not -path "./vendor/*" -exec gofmt -w "{}" \;
-	find . -name "*.go" -not -path "./vendor/*" -exec goimports -l -w "{}" \;
+	find . -name "*.go" -not -path "./vendor/*" -exec goimports -local github.com/mongodb/mongodb-atlas-kubernetes -l -w "{}" \;
```
josvazg, line 163: This change is to fix the linter complain that you had not run goimports with `-local github.com/mongodb/mongodb-atlas-kubernetes`

igor-karpukhin: It's ok to have it in this PR. Thanks!

igor-karpukhin (approved):
Looks good 👍

pkg/controller/atlas/client_test.go
```diff
@@ -17,3 +21,22 @@ func TestClientUserAgent(t *testing.T) {
...
+
+func (tt *testTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	tt.used = true
+	return nil, nil
+}
+
+func TestCustomTransport(t *testing.T) {
+	tt := &testTransport{used: false}
+
+	client, err := atlas.Client("https://cloud.mongodb.com", atlas.Connection{}, nil, httputil.CustomTransport(tt))
```
igor-karpukhin, line 38: One nit: @josvazg maybe you meant "cloud-qa"?

josvazg: right, but it does not really matter as we are not accessing anything on the network in this test, the custom transport does nothing. We could even say here `"https://invented.company.co"`
want me to change it anyway?

pkg/controller/atlas/client_test.go
```diff
@@ -17,3 +21,22 @@ func TestClientUserAgent(t *testing.T) {
...
+	assert.NoError(t, err)
+	client.Projects.GetAllProjects(context.Background(), &mongodbatlas.ListOptions{})
+	assert.Equal(t, true, tt.used)
```
gssbzn, line 41:
```suggestion
	assert.True(t, tt.used)
```

pkg/controller/atlas/client_test.go
```diff
@@ -17,3 +21,22 @@ func TestClientUserAgent(t *testing.T) {
...
+func (tt *testTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	tt.used = true
+	return nil, nil
+}
+
+func TestCustomTransport(t *testing.T) {
+	tt := &testTransport{used: false}
+
+	client, err := atlas.Client("https://cloud.mongodb.com", atlas.Connection{}, nil, httputil.CustomTransport(tt))
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
josvazg: thx

helderjs (approved):
LGTM

PR #1035, opened 2023-07-11
CLOUDP-153003: Encryption at Rest Tests for GCP & Azure

roothorp:
### All Submissions:


test/e2e/actions/cloud/azure.go
```diff
@@ -392,6 +399,62 @@ func (a *AzureAction) deletePrivateEndpoint(ctx context.Context, endpointName st
...
+		Kty:     toptr.MakePtr(azkeys.KeyTypeRSA),
+	}
+
+	r, err := client.CreateKey(ctx, keyName, params, nil)
+	if err != nil {
+		return "", err
+	}
+
+	a.t.Cleanup(func() {
+		err = a.deleteKeyVault(ctx, keyName)
```
josvazg, line 423: We are not doing auto clean up for AWS right?
Can you explain why is that the case?

test/e2e/encryption_at_rest_test.go
```diff
@@ -31,11 +31,21 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/utils"
 )
 
+const (
+	ClientID         = "AZURE_CLIENT_ID"
+	KeyVaultName     = "ako-kms-test"
+	ClientSecret     = "AZURE_CLIENT_SECRET" //#nosec G101 -- False positive; this is the env var, not the secret itself
```
josvazg, line 37: [nitpick]
These 3 are Azure specific, I wonder if we want to use a prefix in the name to make it more explicit where used?

test/e2e/actions/cloud/gcp.go
```diff
@@ -459,6 +465,50 @@ func (a *GCPAction) deleteVPCPeering(ctx context.Context, vpcName, peerName stri
 	return nil
 }
 
+func (a *GCPAction) CreateKMS() (string, error) {
```
helderjs, line 468:
```suggestion
func (a *GCPAction) CreateKMS() (string, error) {
    a.t.Helper()
```

test/e2e/actions/cloud/gcp.go
```diff
@@ -459,6 +465,50 @@ func (a *GCPAction) deleteVPCPeering(ctx context.Context, vpcName, peerName stri
...
+		CryptoKeyVersionId: keyVersion,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return result.Name, nil
+}
+
+func (a *GCPAction) deleteKMS(ctx context.Context, keyName string) error {
```
helderjs, line 499:
```suggestion
func (a *GCPAction) deleteKMS(ctx context.Context, keyName string) error {
    a.t.Helper()
```

test/e2e/actions/cloud/azure.go
```diff
@@ -392,6 +399,62 @@ func (a *AzureAction) deletePrivateEndpoint(ctx context.Context, endpointName st
...
+
+	keyID := string(*r.KeyBundle.Key.KID)
+
+	return keyID, nil
+}
+
+func (a *AzureAction) deleteKeyVault(ctx context.Context, keyName string) error {
+	a.t.Helper()
+
+	client, err := azkeys.NewClient(AzureKeyVaultURI, a.credentials, nil)
```
helderjs, line 437: As I imagined, this is not the latest SDK API and it differs from the usage in other methods.
See the package `github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault`
You will be able to create a factory and use the `PollUntilDone()` from the returned operation.

test/e2e/actions/cloud/aws.go
```diff
@@ -25,13 +31,137 @@ type awsNetwork struct {
...
+		PolicyName: aws_sdk.String("default"),
+		Policy:     aws_sdk.String(policyString),
+	}
+
+	_, err = kmsClient.PutKeyPolicy(policyInput)
+	if err != nil {
+		return "", err
+	}
+
+	// return session.GetCustomerMasterKeyID(atlasAccountArn, assumedRoleArn)
```
helderjs, line 77: remove unnecessary comment

test/e2e/encryption_at_rest_test.go
```diff
@@ -155,15 +208,31 @@ func fillVaultforAzure(encAtRest *v1.EncryptionAtRest) {
...
 
 func fillKMSforGCP(encAtRest *v1.EncryptionAtRest) {
 	if (encAtRest.GoogleCloudKms == v1.GoogleCloudKms{}) {
 		return
 	}
 
-	// todo: fill in
+	t := GinkgoT()
+
+	gcpAction, err := cloud.NewGCPAction(t, cloud.GoogleProjectID)
```
helderjs, line 229:
```suggestion
	gcpAction, err := cloud.NewGCPAction(GinkgoT(), cloud.GoogleProjectID)
```

test/e2e/encryption_at_rest_test.go
```diff
@@ -155,15 +208,31 @@ func fillVaultforAzure(encAtRest *v1.EncryptionAtRest) {
 		return
 	}
 
-	// todo: fill in
+	t := GinkgoT()
+
+	azAction, err := cloud.NewAzureAction(t, os.Getenv(SubscriptionID), cloud.ResourceGroupName)
```
helderjs, line 213:
```suggestion
	azAction, err := cloud.NewAzureAction(GinkgoT(), os.Getenv(SubscriptionID), cloud.ResourceGroupName)
```

helderjs (changes requested):

test/e2e/api/aws/aws.go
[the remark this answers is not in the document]
roothorp: I think it's just gradual code sprawl over time; moving this was at the request of @helderjs, who I think wants to converge all of this cloud util stuff in actions/

test/e2e/actions/cloud/azure.go
roothorp: `armkeyvault` has [no functionality to delete keys](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault#KeysClient) (just create and get), as far as I can tell. I'm not sure why there is an additional module for keyvault like this, but it has delete functionality. Alternatively, I could use `armkeyvault` to create a new vault every time, which you _can_ delete and poll, but that might add some more work for figuring out permissions when creating the vault.

test/e2e/actions/cloud/azure.go
roothorp: I'm not sure, the AWS tests were already here. There is some setup done for AWS in the tests, but as you know we need to have already done some setup manually. Would be nice to eventually have everything happen in-test (infrastructure-wise) so tests don't need to be run specifically in our cloud accounts

igor-karpukhin (approved):
Looks good 👍  But let's wait for the CI to become green

helderjs (approved):

.github/workflows/test-e2e.yml
```diff
@@ -223,11 +223,11 @@ jobs:
           AWS_ACCOUNT_ARN_LIST: ${{ secrets.AWS_ACCOUNT_ARN_LIST }}
           AWS_KMS_KEY_ID: ${{ secrets.AWS_KMS_KEY_ID }}
```
josvazg, line 224: No need to update the AWS credentials?

helderjs: No, there's no problem with AWS account

test/e2e/encryption_at_rest_test.go
```diff
@@ -64,7 +74,13 @@ var _ = Describe("Encryption at REST test", Label("encryption-at-rest"), func()
 		func(test *model.TestDataProvider, encAtRest v1.EncryptionAtRest, roles []cloudaccess.Role) {
 			testData = test
 			actions.ProjectCreationFlow(test)
-			encryptionAtRestFlow(test, encAtRest, roles)
+
+			if roles != nil {
+				cloudAccessRolesFlow(test, roles)
+				encAtRest.AwsKms.RoleID = test.Project.Status.CloudProviderAccessRoles[0].IamAssumedRoleArn
+			}
```
josvazg, line 81: Is this the fix for encryption at rest flaky tests?

test/e2e/encryption_at_rest_test.go
```diff
@@ -90,20 +104,54 @@ var _ = Describe("Encryption at REST test", Label("encryption-at-rest"), func()
...
 	By("Create KMS", func() {
 		Expect(userData.K8SClient.Get(userData.Context, types.NamespacedName{Name: userData.Project.Name,
 			Namespace: userData.Resources.Namespace}, userData.Project)).Should(Succeed())
 
-		Expect(len(userData.Project.Status.CloudProviderAccessRoles)).NotTo(Equal(0))
-		aRole := userData.Project.Status.CloudProviderAccessRoles[0]
+		var aRole status.CloudProviderAccessRole
+		if len(userData.Project.Status.CloudProviderAccessRoles) > 0 {
+			aRole = userData.Project.Status.CloudProviderAccessRoles[0]
+		}
```
josvazg, line 154: Should we not fail already here in the else case?

test/e2e/encryption_at_rest_test.go
```diff
@@ -146,7 +194,8 @@ func fillKMSforAWS(encAtRest *v1.EncryptionAtRest, atlasAccountArn, assumedRoleA
 	}
 
 	Expect(encAtRest.AwsKms.Region).NotTo(Equal(""))
-	awsAction := cloud.NewAwsAction()
+	awsAction, err := cloud.NewAWSAction(GinkgoT())
+	Expect(err).ToNot(HaveOccurred())
```
josvazg, line 198: Was this failing silently before?

josvazg: Never mind, seems it changed signature

helderjs: No, this change was a refactor done in the test utility

test/e2e/encryption_at_rest_test.go
helderjs: This was the actual bug. For this test case, roleID was not being set and the operator was failing to create it.
Despite there being a piece of code in the operator to handle this, it depends on having the latest version of the resource with updated status. Locally this was easily achieved and the test was always passing while in the CI we always failing.

test/e2e/encryption_at_rest_test.go
```diff
@@ -198,7 +259,7 @@ func checkIfEncryptionsAreDisabled(projectID string) (areEmpty bool, err error)
 	return true, nil
 }
 
-var _ = Describe("Encryption at rest AWS", Label("encryption-at-rest"), func() {
+var _ = Describe("Encryption at rest AWS", Label("encryption-at-rest"), Ordered, func() {
```
josvazg, line 262: Did ordering influence flakiness?

josvazg (approved):
LGTM for the title of the PR

For the related fix on flaky tests, I have some questions, but non of them are blocking. Just curiosity or something to follow up on.

test/e2e/encryption_at_rest_test.go
helderjs: I believe it is like this because this applies only to the AWS config. As the cloud provider configuration was happening before and now I moved it to even earlier, it is not supposed to fail ever.

test/e2e/encryption_at_rest_test.go
helderjs: For this test case yes. We have this `var testData *model.TestDataProvider` global variable that is only set on the spec (`It`). The way Ginkgo behaves might lead to one spec overriding what was set by the other. The proper way to use global vars is to set them using `BeforeEach` so the framework will properly encapsulate it when creating the Test Tree.

helderjs: Making it ordered avoids one spec to override the data set by the other.

test/e2e/encryption_at_rest_test.go
josvazg: not always failing, but almost always right? seemed we got lucky sometimes.

josvazg:
Thanks for the replies?

PR #1041, opened 2023-07-18
CLOUDP-188104: Deployment Resource Tagging

cveticm:
### All Submissions:


pkg/api/v1/atlasdeployment_types.go
```diff
@@ -205,9 +214,19 @@ type ServerlessSpec struct {
...
+	// +kubebuilder:validation:MaxItems=50
+	// +optional
+	Tags []*TagSpec `json:"tags,omitempty"`
+}
+
+// ServerlessToAtlas converts the ServerlessSpec to native Atlas client Cluster format.
+func (s *ServerlessSpec) ServerlessToAtlas() (*mongodbatlas.Cluster, error) {
+	result := &mongodbatlas.Cluster{}
+	err := compat.JSONCopy(result, s)
+	return result, err
```
josvazg, line 229: nitpick, I'd rather be explicit here, even if it si verbose to map each key to its same or similar key value on the original. This way we see what fields we are talking about and, if we ever need to make some non-standard mapping going forward, the change is smaller.

Not a blocker at all

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -477,3 +482,24 @@ func (r *AtlasDeploymentReconciler) removeDeletionFinalizer(context context.Cont
 }
 
 type deploymentHandlerFunc func(ctx *workflow.Context, project *mdbv1.AtlasProject, deployment *mdbv1.AtlasDeployment, req reconcile.Request) (workflow.Result, error)
+
+// Parse through tags and verfiy that all keys are unique. Return error on duplicate key.
```
josvazg, line 486: This can be easily unit tested. Being part of the spec, I think it makes even more sense.

pkg/api/v1/atlasdeployment_types.go
josvazg: BTW, the explicit mapping would not fail, would it? no err needed

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -477,3 +482,24 @@ func (r *AtlasDeploymentReconciler) removeDeletionFinalizer(context context.Cont
...
+func uniqueKey(deploymentSpec *mdbv1.AtlasDeploymentSpec) (workflow.Result, error) {
```
josvazg, line 487: I think we do not need 2 results here. Error would suffice.

```suggestion
func uniqueKey(deploymentSpec *mdbv1.AtlasDeploymentSpec) error {
```

That will simplify the returns here and then on the calling spot you just handle the result as needed:

```go
	if err := uniqueKey(&deployment.Spec); err != nil {
		log.Errorw("failed to validate tags", "error", err)
		result := workflow.Terminate(workflow.Internal, err.Error())
		ctx.SetConditionFromResult(status.DeploymentReadyType, result)
	}
```

Not a big deal anyway, but the unit test will also be a bit easier.

cveticm: Noted, I'll push these changes with my next commit!

pkg/api/v1/atlasdeployment_types.go
cveticm: I went with this method since a similar method is already used elsewhere in the reconciler (`ToAtlas`) If you think the approach you suggested is better, I can look into it! I'm just not entirely sure I'm understanding the benefit of using explicit mapping over what's already there. Would you be able to explain cases we may need non-standard mapping?

pkg/controller/atlasdeployment/atlasdeployment_controller.go
josvazg: Another option is to use just `workflow.Result`, as is can always wrap an error inside as needed (the error message).

pkg/api/v1/atlasdeployment_types.go
josvazg: I noticed this is kind of a standard in the code later on. Please feel free to leave it as is.

josvazg: Why not simply `ToAtlas()` here? seems it is the default name on most other situations, and being a method os serverless already the prefix feels a bit redundant.

cveticm: It feels a bit redundant, yes, but Advanced and Serverless clusters have different structs (`mongodbatlas.AdvancedCluster` and `mongodbatlas.Cluster`) so two methods are needed to return the appropriate struct type

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -205,9 +214,19 @@ type ServerlessSpec struct {
...
 	Name string `json:"name"`
 	// Configuration for the provisioned hosts on which MongoDB runs. The available options are specific to the cloud service provider.
-	ProviderSettings *ProviderSettingsSpec `json:"providerSettings"`
-
+	ProviderSettings *ProviderSettingsSpec       `json:"providerSettings"`
 	PrivateEndpoints []ServerlessPrivateEndpoint `json:"privateEndpoints,omitempty"`
+	// Key-value pairs for resource tagging.
+	// +kubebuilder:validation:MaxItems=50
+	// +optional
+	Tags []*TagSpec `json:"tags"`
```
roothorp, line 222: should this be `omitempty` too or does serverless behave differently?

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/sethvargo/go-password v0.2.0
 	github.com/stretchr/testify v1.8.4
-	go.mongodb.org/atlas v0.31.0
+	go.mongodb.org/atlas v0.32.0
```
josvazg, line 25: uhm is this upgrade needed by this change? if not I would advise against including it here. We had a dependabot attempt that failed today !1058 and it is being included and fixed by !1041 from @roothorp already, if I am not mistaken.

But maybe it is required.

roothorp: This is #1041 😄

pkg/controller/atlasdeployment/serverless_deployment.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"reflect"
```
josvazg, line 7: I would usually avoid reflect on production code whenever possible...

pkg/controller/atlasdeployment/serverless_deployment.go
```diff
@@ -40,8 +46,26 @@ func ensureServerlessInstanceState(ctx *workflow.Context, project *mdbv1.AtlasPr
...
 	switch atlasDeployment.StateName {
 	case status.StateIDLE:
+		convertedDeployment, err := serverlessSpec.ServerlessToAtlas()
+		if err != nil {
+			return atlasDeployment, workflow.Terminate(workflow.Internal, err.Error())
+		}
+		if convertedDeployment.Tags == nil {
+			convertedDeployment.Tags = &[]*mongodbatlas.Tag{}
+		}
+		if !reflect.DeepEqual(convertedDeployment.Tags, atlasDeployment.Tags) {
```
josvazg, line 56: And turns out we do not want reflect here I think, as it makes this check incorrect:
- []TagSpec is an array, so order is important and DeepEqual will take it into account.
- But in reality, we do not care if x tags appear in the same or different order, the tag set is the same right?

I would make a custom function for the check and unit test it as per requirements.

pkg/api/v1/atlasdeployment_types.go
josvazg, line 222: An alternative way to model this with types would have been doing this here:

```suggestion
	Tags TagSpecSet `json:"tags"`
```

Where:
```
type TagSpecSet []*TagSpec

// Validate returns an error is TagSpecSet is not a proper set, keys are repeated.
func (tset *TagSpecSet) Validate() error {
...

// Equal returns true if the given TagSet holds the same key value pairs, regardless or ordering
func (tset *TagSpecSet) Equal(other *TagSpecSet) bool {
...
```

roothorp: looking at https://github.com/mongodb/go-client-mongodb-atlas/pull/509, assuming this is intentional

test/int/deployment_test.go
```diff
@@ -241,6 +243,9 @@ var _ = Describe("AtlasDeployment", Label("int", "AtlasDeployment"), func() {
 	performUpdate := func(timeout time.Duration) {
 		Expect(k8sClient.Update(context.Background(), createdDeployment)).To(Succeed())
 
+		j, _ := json.MarshalIndent(createdDeployment, "", " ")
+		GinkgoWriter.Println(">>>", string(j))
```
josvazg, line 247: Is this a left over from traces?

pkg/controller/atlasdeployment/serverless_deployment.go
helderjs: I'd second José here and discourage the usage of reflect in the code.

pkg/api/v1/atlasdeployment_types.go
helderjs: While I got the point here, I'd say calling a slice a Set is a bad name choice :smile:

pkg/controller/atlasdeployment/serverless_deployment.go
igor-karpukhin: @cveticm it looks like this has to be converted into a custom comparison function, as `reflect.DeepEqual()` takes into account the positions of array elements:
```
	t1 := []*TagSpec{&TagSpec{Key: "1", Value: "2"}, &TagSpec{Key: "2", Value: "3"}}
	t2 := []*TagSpec{&TagSpec{Key: "2", Value: "3"}, &TagSpec{Key: "1", Value: "2"}}
```
These two won't be equal for `DeepEqual`

pkg/controller/atlasdeployment/serverless_deployment.go
```diff
@@ -40,8 +46,26 @@ func ensureServerlessInstanceState(ctx *workflow.Context, project *mdbv1.AtlasPr
...
+			atlasDeployment, _, err = ctx.Client.ServerlessInstances.Update(context.Background(), project.Status.ID, serverlessSpec.Name, &mongodbatlas.ServerlessUpdateRequestParams{
+				// TODO: include ServerlessBackupOptions and TerminationProtectionEnabled
```
igor-karpukhin, line 58: Let's create a ticket add these fields

test/int/deployment_test.go
```diff
@@ -585,6 +590,32 @@ var _ = Describe("AtlasDeployment", Label("int", "AtlasDeployment"), func() {
...
 			})
 
+			By("Updating the Deployment tags", func() {
+				createdDeployment.Spec.DeploymentSpec.Tags = []*mdbv1.TagSpec{{Key: "int-test", Value: "true"}}
+				performUpdate(20 * time.Minute)
+				doDeploymentStatusChecks()
+				checkAtlasState(func(c *mongodbatlas.AdvancedCluster) {
+					for i, tag := range createdDeployment.Spec.DeploymentSpec.Tags {
+						Expect(reflect.DeepEqual(c.Tags[i].Key, tag.Key)).To(BeTrue())
+						Expect(reflect.DeepEqual(c.Tags[i].Value, tag.Value)).To(BeTrue())
```
igor-karpukhin, line 600:
```suggestion
						Expect(c.Tags[i].Key == tag.Key).To(BeTrue())
						Expect(c.Tags[i].Value == tag.Value).To(BeTrue())
```

test/int/deployment_test.go
```diff
@@ -1020,11 +1051,37 @@ var _ = Describe("AtlasDeployment", Label("int", "AtlasDeployment"), func() {
...
 
+			By("Updating the Instance tags", func() {
+				createdDeployment.Spec.ServerlessSpec.Tags = []*mdbv1.TagSpec{{Key: "int-test", Value: "true"}}
+				performUpdate(20 * time.Minute)
 				doServerlessDeploymentStatusChecks()
+				atlasDeployment, _, _ := atlasClient.ServerlessInstances.Get(context.Background(), createdProject.Status.ID, createdDeployment.Spec.ServerlessSpec.Name)
+				if createdDeployment != nil {
+					for i, tag := range createdDeployment.Spec.ServerlessSpec.Tags {
+						Expect(reflect.DeepEqual((*atlasDeployment.Tags)[i].Key, tag.Key)).To(BeTrue())
+						Expect(reflect.DeepEqual((*atlasDeployment.Tags)[i].Value, tag.Value)).To(BeTrue())
```
igor-karpukhin, line 1068: Same thing here with the reflection

igor-karpukhin (changes requested):
Looks good in general! But let's replace `DeepEqual` comparison for Tags

go.mod
cveticm: update is required. You can see the changes needed in this PR [https://github.com/mongodb/go-client-mongodb-atlas/pull/509](url)

pkg/controller/atlasdeployment/serverless_deployment.go
cveticm: Created [CLOUDP-193173](https://jira.mongodb.org/browse/CLOUDP-193173)

pkg/controller/atlasdeployment/serverless_deployment.go
```diff
@@ -54,3 +77,16 @@ func ensureServerlessInstanceState(ctx *workflow.Context, project *mdbv1.AtlasPr
 		return atlasDeployment, workflow.Terminate(workflow.Internal, fmt.Sprintf("unknown deployment state %q", atlasDeployment.StateName))
 	}
 }
+
+func isEqual(a *mongodbatlas.Cluster, c *mongodbatlas.Cluster) bool {
```
josvazg, line 81: I'd rename as `isEqualTags()` because we are only checking the tags, not the cluster objects in full.

Alternatively, it can be left as isEqual() but just passing 2 arrays of tags to be compared.

pkg/controller/atlasdeployment/serverless_deployment.go
```diff
@@ -54,3 +77,16 @@ func ensureServerlessInstanceState(ctx *workflow.Context, project *mdbv1.AtlasPr
...
+	// TODO: include ServerlessBackupOptions and TerminationProtectionEnabled
+	if len(*a.Tags) == len(*c.Tags) {
+		for i, aTags := range *a.Tags {
```
josvazg, line 84: This will find differences if the arrays contain the same key-value pairs but in different order.
One way to avoid this is to order both arrays before comparing.
Another is to put all keys into a map from the first array and check all key pairs from the second array are in the map built before.

josvazg (commented):
I am confused, I was assuming that tags order did not matter. Can you confirm this point?

pkg/controller/atlasdeployment/serverless_deployment.go
```diff
@@ -54,3 +77,15 @@ func ensureServerlessInstanceState(ctx *workflow.Context, project *mdbv1.AtlasPr
...
+	if len(a) == len(c) {
+		for i, aTags := range a {
+			if aTags.Key != (c)[i].Key || aTags.Value != (c)[i].Value {
+				return false
+			}
+		}
+		return true
+	}
+	return false
+}
```
josvazg, line 91: I see we decided on making the order matter. This is fine, except we now need to discard the following pathologic case: Say we have the pair `{A=a,B=b}` in Atlas, but the k8s spec says `{B=b,A=a}`

1. This detects a change, because the order does not match.
2. A reconcile does patch the tags, but as the tags are present, Atlas makes no real changes on its side, **and keeps the old order**.
3. Next reconcile, we start again from 1 and we are in a loop reconciling unordered tags that do not need to be reconciled.

Do we have any test confirming or discarding that this can happen with Atlas tags?

josvazg:
LGTM, except I would like to confirm whether or not that pathological case of trying to reorder tags that never get reordered cannot happen.

pkg/controller/atlasdeployment/serverless_deployment.go
josvazg: Thanks for double checking. No more objections then.

josvazg (approved):
LGTM

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
+)
+
+func TestUniqueKey(t *testing.T) {
+	//var deploymentSpec *mdbv1.AtlasDeploymentSpec
```
roothorp, line 12: Can this be removed?

roothorp (approved):
LGTM - one tiny comment to remove (?)

pkg/controller/atlasdeployment/serverless_deployment.go
```diff
@@ -40,8 +45,26 @@ func ensureServerlessInstanceState(ctx *workflow.Context, project *mdbv1.AtlasPr
...
 	switch atlasDeployment.StateName {
 	case status.StateIDLE:
+		convertedDeployment, err := serverlessSpec.ToAtlas()
+		if err != nil {
+			return atlasDeployment, workflow.Terminate(workflow.Internal, err.Error())
+		}
+		if convertedDeployment.Tags == nil {
+			convertedDeployment.Tags = &[]*mongodbatlas.Tag{}
+		}
+		if !isTagsEqual(*(atlasDeployment.Tags), *(convertedDeployment.Tags)) {
```
igor-karpukhin, line 55: Are we sure here that `atlasDeployment.Tags` never becomes nil?

pkg/controller/atlasdeployment/serverless_deployment.go
```diff
@@ -54,3 +77,15 @@ func ensureServerlessInstanceState(ctx *workflow.Context, project *mdbv1.AtlasPr
 		return atlasDeployment, workflow.Terminate(workflow.Internal, fmt.Sprintf("unknown deployment state %q", atlasDeployment.StateName))
 	}
 }
+
+func isTagsEqual(a []*mongodbatlas.Tag, c []*mongodbatlas.Tag) bool {
+	if len(a) == len(c) {
+		for i, aTags := range a {
+			if aTags.Key != (c)[i].Key || aTags.Value != (c)[i].Value {
```
igor-karpukhin, line 84: nit: redundant parentheses for `(c)`

igor-karpukhin (approved):
Looks good in general 👍

pkg/controller/atlasdeployment/serverless_deployment.go
cveticm: I believe the behaviour is that  `atlasDeployment.Tags` cannot go to nil after an initial value is fed to it, be it a full or empty array of pointers.
Since `performUpdate` passes after removing tags [here](https://github.com/mongodb/mongodb-atlas-kubernetes/blob/CLOUDP-188104/deployment-resource-tagging/test/int/deployment_test.go#L1101-L1103) that means AKO returns to idle so `atlasDeployment.Tags` has remained an empty array instead of converting to nil.

igor-karpukhin (approved):

PR #1051, opened 2023-07-25
CLOUDP-175080: Project resources deletion protection

helderjs:
Includes:
* IP Access List
* Cloud Provider Access
* Network Peering
* ~Alert Configuration~ Excluded from deletion protection
* Integrations

### All Submissions:


cmd/manager/main.go
```diff
@@ -52,7 +50,14 @@ import (
...
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/util/kube"
-	// +kubebuilder:scaffold:imports
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/version"
+)
+
+const (
+	objectDeletionProtectionFlag       = "object-deletion-protection"
+	subobjectDeletionProtectionFlag    = "subobject-deletion-protection"
+	objectDeletionProtectionDefault    = true
+	subobjectDeletionProtectionDefault = true
```
josvazg, line 60: These should be false instead, for now, right?

cmd/manager/main.go
```diff
@@ -144,28 +147,32 @@ func main() {
...
+		Client:                      mgr.GetClient(),
+		Log:                         logger.Named("controllers").Named("AtlasProject").Sugar(),
+		Scheme:                      mgr.GetScheme(),
+		AtlasDomain:                 config.AtlasDomain,
+		ResourceWatcher:             watch.NewResourceWatcher(),
+		GlobalAPISecret:             config.GlobalAPISecret,
+		GlobalPredicates:            globalPredicates,
+		EventRecorder:               mgr.GetEventRecorderFor("AtlasProject"),
+		ObjectDeletionProtection:    config.ObjectDeletionProtection,
+		SubObjectDeletionProtection: config.SubObjectDeletionProtection,
```
josvazg, line 159: project and user was already implemented, why where these changes still needed here? I guess I am missing something.

josvazg:
I see more changes than I expected. As if this branch was off `main` instead. What am I missing?

Makefile
```diff
@@ -129,7 +129,7 @@ run: generate fmt vet manifests ## Run against the configured Kubernetes cluster
 	OPERATOR_POD_NAME=$(OPERATOR_POD_NAME) \
 	OPERATOR_NAMESPACE=$(OPERATOR_NAMESPACE) \
 	go run ./cmd/manager/main.go --atlas-domain=$(ATLAS_DOMAIN) \
-	--global-api-secret-name=$(ATLAS_KEY_SECRET_NAME)
+	--global-api-secret-name=$(ATLAS_KEY_SECRET_NAME) --log-level=debug
```
josvazg, line 132: Do we want this debug log level all the time here or only during current development?

cmd/manager/main.go
```diff
@@ -240,6 +240,10 @@ func parseConfiguration() Config {
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.StringVar(&config.LogLevel, "log-level", "info", "Log level. Available values: debug | info | warn | error | dpanic | panic | fatal")
 	flag.StringVar(&config.LogEncoder, "log-encoder", "json", "Log encoder. Available values: json | console")
+	flag.BoolVar(&config.ObjectDeletionProtection, objectDeletionProtectionFlag, true, "Defines the operator will not delete Atlas resource "+
+		"when a Custom Resource is deleted")
+	flag.BoolVar(&config.SubObjectDeletionProtection, subobjectDeletionProtectionFlag, true, "Defines that the operator will not overwrite "+
+		"(and consequently delete) subresources that were not previously created by the operator")
```
josvazg, line 246: We must remember to remove these CLI flags again, and the associated variables. It is due to the rebase. Same happened to my PR.

I will remind in Slack as well.

josvazg (changes requested):

helm-charts
josvazg, line 1: The CLI flags are not there, the submodule might need to go back to `main` branch for now I think.

pkg/api/v1/zz_generated.deepcopy.go
```diff
@@ -14,10 +14,9 @@ a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
 package v1
 
 import (
-	"k8s.io/apimachinery/pkg/runtime"
-
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/common"
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/project"
+	"k8s.io/apimachinery/pkg/runtime"
```
josvazg, line 19: This is a formatting issue between the code generator and our enforced import formatting.

I will send a small fix today.

pkg/controller/atlasproject/alert_configurations.go
```diff
@@ -317,3 +337,35 @@ func isAlertConfigSpecEqualToAtlas(logger *zap.SugaredLogger, alertConfigSpec md
...
+		return true, nil
+	}
+
+	latestConfig := &mdbv1.AtlasProjectSpec{}
+	latestConfigString, ok := project.Annotations[customresource.AnnotationLastAppliedConfiguration]
+	if ok {
+		if err := json.Unmarshal([]byte(latestConfigString), latestConfig); err != nil {
+			return false, err
+		}
+	}
```
josvazg, line 352: So for Alter Configs to be managed, not only a config must already exist, but also we should not find certain differences?
Can you elaborate a bit more on this?

pkg/controller/atlasproject/alert_configurations.go
```diff
@@ -317,3 +337,35 @@ func isAlertConfigSpecEqualToAtlas(logger *zap.SugaredLogger, alertConfigSpec md
...
+
+	diff := sortAlertConfigs(logger, latestConfig.AlertConfigurations, list)
+	if len(diff.Create) == 0 && len(diff.Delete) == 0 {
+		return true, nil
+	}
+
+	diff = sortAlertConfigs(logger, project.Spec.AlertConfigurations, list)
+
+	return len(diff.Create) == 0 || len(diff.Delete) == 0, nil
```
josvazg, line 370: nitpick: these sub-comparisons might self document a bit better if they were more to their own function with a name that explains what we are checking for on each case.

pkg/controller/atlasproject/alert_configurations_test.go
```diff
@@ -0,0 +1,406 @@
...
+	return nil, nil, nil
+}
+
+func (c *alertConfigurationClient) Update(_ context.Context, _ string, _ string, _ *mongodbatlas.AlertConfiguration) (*mongodbatlas.AlertConfiguration, *mongodbatlas.Response, error) {
+	return nil, nil, nil
+}
+
+func (c *alertConfigurationClient) Delete(_ context.Context, _ string, _ string) (*mongodbatlas.Response, error) {
+	return nil, nil
+}
```
josvazg, line 54: I think this mock code must go to its own package, or at least a separate file.

Note we will use these mocks from many tests. Not sure if we want mocks on all packages or a package with all mocks.

Makefile
helderjs: I don't see a case where someone would start the manager manually other than in development, and in that case, we usually are interested in debug-level messages.

pkg/controller/atlasproject/alert_configurations_test.go
helderjs: Later, they will all be extracted to (probably) their own package

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -160,35 +160,40 @@ func (r *AtlasProjectReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-	if result := r.ensureDeletionFinalizer(ctx, workflowCtx, atlasClient, project); !result.IsOk() {
+	if result = r.ensureDeletionFinalizer(ctx, workflowCtx, atlasClient, project); !result.IsOk() {
 		setCondition(workflowCtx, status.ProjectReadyType, result)
 		return result.ReconcileResult(), nil
 	}
 
+	if project.ID() == "" {
+		err = customresource.ApplyLastConfigApplied(ctx, project, r.Client)
+		if err != nil {
```
josvazg, line 186: Why are we applying last config here? seems like an early return?

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -379,6 +399,10 @@ func managedByAtlas(ctx context.Context, atlasClient mongodbatlas.Client) custom
 			return false, err
 		}
 
+		if project.Spec.Name == atlasProject.Name {
```
josvazg, line 402: So if the name matches we take ownership?

pkg/controller/atlasproject/alert_configurations.go
helderjs: Most of the resources can't be identified by a single key, many are composed an/or different compositions when they are different types of resources being manipulated.
Some of them had a comparison function built already and in this case, I decided to reuse them to match the same reconciliation logic, for other, I have to come up with the comparison logic

pkg/controller/atlasproject/alert_configurations_test.go
josvazg: I would already put them on their own file, even if we keep them in the same package. When we move them later it will be clearer what is being done.

pkg/controller/atlasproject/atlasproject_controller.go
helderjs: In order to properly check subresources, the last config annotation must be present, if a new CR is added we first sync the project itself and re-queue to sync the subresources

pkg/controller/atlasproject/atlasproject_controller.go
helderjs: We use the name to match a project if for any reason we fetch a project by its ID and they don't match the name, something is wrong and we should not proceed.

pkg/controller/atlasproject/cloud_provider_access.go
```diff
@@ -247,3 +269,64 @@ func sortAccessRoles(ctx context.Context, accessClient mongodbatlas.CloudProvide
...
+		return true, nil
+	}
+
+	latestConfig := &v1.AtlasProjectSpec{}
+	latestConfigString, ok := akoProject.Annotations[customresource.AnnotationLastAppliedConfiguration]
+	if ok {
+		if err := json.Unmarshal([]byte(latestConfigString), latestConfig); err != nil {
+			return false, err
+		}
+	}
```
josvazg, line 284: This part of the check seems generic, shared with all the other subresources right?

The only thing missing from the generic function would be the marshalling, which we can do if the argument is being passed, or on a new generic function. Then we would avoid code repetition in many places.

pkg/controller/atlasproject/cloud_provider_access_test.go
```diff
@@ -0,0 +1,300 @@
...
+	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
+
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/atlas/mongodbatlas"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/customresource"
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/workflow"
+)
+
+type cloudProviderAccessClient struct {
```
josvazg, line 17: again, would move the base mock code to its own file, even in same package.

pkg/controller/atlasproject/alert_configurations.go
helderjs: This function could be broken into 3 or 4 pieces following the same argument, I believe right now it still has a single responsibility, and doesn't make sense to have micro functions being reused by it.

pkg/controller/atlasproject/cloud_provider_access_test.go
helderjs: As mentioned, will do it as a separated step.

test/e2e/actions/deploy/deploy_operator.go
```diff
@@ -28,14 +28,13 @@ func MultiNamespaceOperator(data *model.TestDataProvider, watchNamespace []strin
 		for _, ns := range watchNamespace {
 			watchNamespaceMap[ns] = true
 		}
-		mgr, err := k8s.RunOperator(&k8s.Config{
+		mgr, err := k8s.BuildManager(&k8s.Config{
```
josvazg, line 31: we were running and now we are building? can you elaborate?

test/e2e/cli/helm/helm.go
```diff
@@ -107,6 +107,8 @@ func InstallOperatorWideSubmodule(input model.UserInputs) {
 		"atlas-operator-"+input.Project.GetProjectName(),
 		config.AtlasOperatorHelmChartPath,
 		"--set-string", fmt.Sprintf("atlasURI=%s", config.AtlasHost),
+		"--set", "objectDeletionProtection=false",
+		"--set", "subobjectDeletionProtection=false",
```
josvazg, line 111: I beleve all this changes need to go away after the rebase, those flags do not exist anymore.

test/e2e/k8s/operator.go
```diff
@@ -33,7 +34,7 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/watch"
 )
 
-func RunOperator(initCfg *Config) (manager.Manager, error) {
+func BuildManager(initCfg *Config) (manager.Manager, error) {
```
josvazg, line 37: Oh I see, we are renaming here. But still, I do not see the difference that explains the semantic change.

Are we building stuff that is not being run yet, or are we also launching it?
I guess we are only building and the semantic of the name before was misleading?

test/e2e/k8s/operator.go
```diff
@@ -205,31 +202,71 @@ func mergeConfiguration(initCfg *Config) *Config {
...
-			EncodeTime:  zapcore.ISO8601TimeEncoder,
-		},
+func WithGlobalKey(key client.ObjectKey) ManagerConfig {
+	return func(config *Config) {
+		config.GlobalAPISecret = key
 	}
-	return cfg.Build()
+}
+
+func RunManager(withConfigs ...ManagerConfig) (ManagerStart, error) {
```
josvazg, line 252: aha! not sure where this run was happening before though

pkg/controller/atlasproject/alert_configurations_test.go
helderjs: This task is quite big and has been running for a long time. I have already created a ticket to follow up on the mocks.

pkg/controller/atlasproject/cloud_provider_access.go
helderjs: It seemed a good idea at the beginning but after trying it, I just replaced this code for an evaluation of the result of it.
In this case, I'm not sure it makes sense.
Given the extraction of this snippet:
```go
func shouldReconcile(
    ctx context.Context,
    atlasClient mongodbatlas.Client,
    protected bool,
    akoProject *v1.AtlasProject,
) (bool, *v1.AtlasProjectSpec, error) {
	if !protected {
		return true, nil, nil
	}

	latestConfig := &v1.AtlasProjectSpec{}
	latestConfigString, ok := akoProject.Annotations[customresource.AnnotationLastAppliedConfiguration]
	if ok {
		if err := json.Unmarshal([]byte(latestConfigString), latestConfig); err != nil {
			return false, nil, err
		}
	}
	
	return false, latestConfig, nil
}
```
I'd have to replace the existing code with this in all functions
```suggestion
func canCloudProviderAccessReconcile(ctx context.Context, atlasClient mongodbatlas.Client, protected bool, akoProject *v1.AtlasProject) (bool, error) {
	proceed, spec, err := shouldReconcile(ctx, atlasClient, protected, akoProject)
	if err != nil {
	    return false, err
	}
	
	if proceed {
	    return true
	}
	
	....
```

I don't see real gain in this change. Do you have another approach to suggest?

josvazg: Yeah, you are right, I do not see the gain now. Lets leave it as is.

test/e2e/k8s/operator.go
josvazg: got it now, the run before was not a run but a build, so the name change makes sense. The manager is being run just the same as before.

pkg/api/v1/zz_generated.deepcopy.go
josvazg: See #1065

pkg/controller/atlasproject/alert_configurations.go
```diff
@@ -317,3 +337,35 @@ func isAlertConfigSpecEqualToAtlas(logger *zap.SugaredLogger, alertConfigSpec md
 
 	return true
 }
+
+func canAlertConfigurationReconcile(ctx context.Context, atlasClient mongodbatlas.Client, protected bool, project *mdbv1.AtlasProject, logger *zap.SugaredLogger) (bool, error) {
```
josvazg, line 341: nitpick

```suggestion
func canReconcileAlertConfiguration(ctx context.Context, atlasClient mongodbatlas.Client, protected bool, project *mdbv1.AtlasProject, logger *zap.SugaredLogger) (bool, error) {
```

pkg/controller/atlasproject/alert_configurations.go
josvazg, line 370: I think we want both creations AND deletions to not be there to assume control of the alert configurations, don't we?

```suggestion
	return len(diff.Create) == 0 && len(diff.Delete) == 0, nil
```

Or, alternatively:
```suggestion
	return len(diff.Create) != 0 || len(diff.Delete) != 0, nil
```

pkg/controller/atlasproject/alert_configurations_test.go
```diff
@@ -0,0 +1,406 @@
...
+		workflowCtx := &workflow.Context{
+			Client: atlasClient,
+		}
+		reconciler := &AtlasProjectReconciler{}
+		result := reconciler.ensureAlertConfigurations(context.TODO(), workflowCtx, akoProject, true)
+
+		require.Equal(t, workflow.Terminate(workflow.Internal, "unable to resolve ownership for deletion protection: failed to retrieve data"), result)
+	})
+
+	t.Run("should failed to reconcile when unable to synchronize with Atlas", func(t *testing.T) {
```
josvazg, line 328: "unable to synchronize with Atlas" means the atlas state and the k8s config differ, right?

pkg/controller/atlasproject/alert_configurations.go
josvazg: Otherwise, this is tricky enough to place a comment explaining the logic

test/e2e/k8s/operator.go
josvazg: this function was only creating the struct, so it is just building, not running.

pkg/controller/atlasproject/alert_configurations_test.go
helderjs: yes, that's correct

pkg/controller/atlasproject/alert_configurations.go
helderjs: Good catch! I'll fix it

pkg/controller/atlasproject/alert_configurations.go
helderjs: I'll decline just because all other functions follow the same structure

.github/workflows/test-e2e.yml
```diff
@@ -229,6 +230,7 @@ jobs:
           AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
           GCP_SA_CRED: ${{ secrets.GCP_SA_CRED_NEW_TEST }}
           DATADOG_KEY: ${{ secrets.DATADOG_KEY }}
+          PAGER_DUTY_SERVICE_KEY: ${{ secrets.PAGER_DUTY_SERVICE_KEY }}
```
josvazg, line 233: Where is this new credential used?

josvazg (approved):
LGTM with just a minor non blocking question

.github/workflows/test-e2e.yml
helderjs: In the e2e test for deletion protection

igor-karpukhin (approved):
Looks good 👍

PR #1052, opened 2023-07-28
CLOUDP-190400: Disable and hide deletion protection

josvazg:
We cannot release it yet, not even disabled by default, as it is still needs to be implemented on several Custom Resources.

### All Submissions:


josvazg:
I merge into `deletion-protection-main` first, so that the PR can be reviewed properly for this particular change. But after this is merged, it is safe to `deletion-protection-main` into `main`.

helderjs (approved):

roothorp (approved):
LGTM

cmd/manager/main.go
```diff
@@ -319,45 +315,10 @@ func initCustomZapLogger(level, encoding string) (*zap.Logger, error) {
...
-		case "true":
-			config.SubObjectDeletionProtection = true
-		case "false":
-			config.SubObjectDeletionProtection = false
-		default:
-			config.SubObjectDeletionProtection = subobjectDeletionProtectionDefault
-		}
-	}
+	config.ObjectDeletionProtection = objectDeletionProtectionDefault
+	config.SubObjectDeletionProtection = subobjectDeletionProtectionDefault
```
helderjs, line 323: What would we do to enable it for the E2E test? We need to find a way, right?

helderjs (changes requested):

cmd/manager/main.go
josvazg: Are not all our protection related tests in integration tests?
If not, could they be moved there instead?

If we need to enable e2e test to test for protection we are giving a way to experiment with an feature which is not completed yet.

josvazg: I have undone all changes we made for e2e test in this branch before

helderjs: I can't test integration with cloud providers in the integration test

josvazg: The only test that we had was on the helm chart. I think we can post-post e2e testing until final release of the full feature. Or at least till all pieces are in place to allow to set protected on via CLI flags, which is not yet the case.

helderjs: Well, while I having been added unit tests and kept most of the tests as integration tests. It's not reliable to go without e2e.
As Atlas depends on some valid entry to reach the proper state, we can't mock such behavior.

josvazg: We are not going anywhere, this feature is incomplete and this change is to make it impossible to partially enable it in production. Being impossible in production means precisely that e2e test cannot test it and should see the system behaving just as it was behaving before. Right?

The only thing that would require e2e testing is, if we believe we need any **additional e2e test to ensure that protection is set of OFF as expected after this change**.

Can you think of any regression e2e test we might be missing to ensure the default protection disabled behaviour keeps working as it should after these changes?

I would assume we had to have that covered already, but if there is any reinforcement we need we can add it now.

Once we can enable/disable in all components, we can add the CLI flags, and then it will make sense to have e2e tests that use such flags on one way or another. In fact, upon changing those flags I would expect some e2e tests to break when protected is set to ON.

helderjs: If doing this at the cost of losing E2E test in the development process of deletion protection, we should not do it.
At the moment, the only way to move forward safely is with our E2E and I can't give that up
@igor-karpukhin do you have an opinion here?

josvazg: I insist, we are NOT giving anything up here.

The E2E test job after this is change to test that everything keeps working as before, that is protection changes had NOT changes anything in production. Again, if we are missing any test to ensure that, we can add them now.

But we should NOT add e2e test for a feature that does not yet exit in production. We can add those later, when it does make sense as we have the flags that enable the feature.

josvazg: Ok, after syncing offline, I think I get it now.

There are some behaviours we need to validate e2e for now, as our codebase stands today. Changes up to here used unit and int tests only, but upcoming ones rely heavily on e2e test that were finding issues to be corrected not surfaced otherwise.

Having the e2e tests delayed on the process, would create a hassle. We could run those tests now on development, but not commit them to main until later. But that would require still keeping a separate branch for those.

So what we can do is to use a couple of undocumented and hidden env vars that would enable the protected behaviour so we can test it in our CI in e2e tests. As a safety measure, ensuring that production code cannot enable the partial feature, we can allow the **hidden env vars ONLY when running non release versions**. That way it should not happen in production, someone would have explicitly go out of their way to enable the feature before it was intended to be completed.

WDYT?

igor-karpukhin: Sorry for the late response. I'd prefer not to give up on E2E testing of course. Why can't we leave this configurable so that E2E tests can configure the operator?

josvazg: Yes, that is what we propose in the end. The feature cannot be activated on production releases, but using environment variables it can be tested on development builds. I am sending that change next.

josvazg: e2e testing of deletion protection is now enabled ONLy for non release versions using env vars flagged "unsupported".

cmd/manager/main_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/version"
+)
+
+const (
+	nonReleaseVersion = "1.8.0-30-g81233c6-dirty"
+	releaseVersion    = "1.9.0-certified"
+)
+
+func TestDeletionProtectionEnable(t *testing.T) {
+	testCases := []struct {
```
helderjs, line 19: Suggestion: when we have that many options to setup a test, let's avoid table testing, IMO it's very hard to see what's actually being tested.

test/e2e/helm_chart_test.go
```diff
@@ -298,14 +297,11 @@ func deleteDeploymentAndOperator(data *model.TestDataProvider) {
 	By("Check project, deployment does not exist", func() {
 		helm.Uninstall(data.Resources.Deployments[0].Spec.GetDeploymentName(), data.Resources.Namespace)
 		Eventually(
-			func(g Gomega) {
-				if atlasClient.IsProjectExists(g, data.Resources.ProjectID) {
-					_, err := atlasClient.Client.Projects.Delete(context.TODO(), data.Resources.ProjectID)
-					g.Expect(err).To(BeNil())
-				}
+			func(g Gomega) bool {
+				return atlasClient.IsProjectExists(g, data.Resources.ProjectID)
```
helderjs, line 301: I think we can simply:

```suggestion
			func(g Gomega) bool {
				atlasClient.IsProjectExists(g, data.Resources.ProjectID)
```

test/e2e/helm_chart_test.go
```diff
@@ -298,14 +297,11 @@ func deleteDeploymentAndOperator(data *model.TestDataProvider) {
...
 			},
 			"7m", "20s",
-		).Should(Succeed(), "Project and deployment should be deleted from Atlas")
+		).Should(BeFalse(), "Project and deployment should be deleted from Atlas")
```
helderjs, line 304: An revert this change:
```suggestion
		).Should(Succeed(), "Project and deployment should be deleted from Atlas")
```

helderjs (approved):
Looks good if CI is happy, just added some suggestions

igor-karpukhin (approved):
Looks good!

PR #1053, opened 2023-07-28
CLOUDP-188083: Autorecover from duplicate project in tests

josvazg:
### All Submissions:

### DIFF
--- a/test/e2e/actions/deploy/deploy_operator.go
+++ b/test/e2e/actions/deploy/deploy_operator.go
@@ -14,7 +14,9 @@ import (
 
 	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/status"
 	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/actions/kube"
+	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/api/atlas"
 	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/config"
+	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/fixtest"
 	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/k8s"
 	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/model"
 	"github.com/mongodb/mongodb-atlas-kubernetes/test/e2e/utils"
@@ -50,9 +52,16 @@ func CreateProject(testData *model.TestDataProvider) {
 		testData.Project.Namespace = testData.Resources.Namespace
 	}
 	By(fmt.Sprintf("Deploy Project %s", testData.Project.GetName()), func() {
+		aClient := atlas.GetClientOrFail()
 		err := testData.K8SClient.Create(testData.Context, testData.Project)
 		Expect(err).ShouldNot(HaveOccurred(), "Project %s was not created", testData.Project.GetName())
 		Eventually(func(g Gomega) {
+			// We reported Atlas creating duplicates of a project with the same name
+			// See https://jira.mongodb.org/browse/CLOUDP-187749
+			// this fix in our tests allows them to automatically fix this issue
+			// and thus avoid a flaky failure when this duplicates happens
+			g.Expect(fixtest.EnsureNoDuplicates(aClient.Client, fixtest.ZapLoggerFrom(GinkgoWriter), testData.Project.Spec.Name)).ToNot(HaveOccurred())
+
 			condition, _ := k8s.GetProjectStatusCondition(
 				testData.Context,
 				testData.K8SClient,
--- a/test/e2e/fixtest/remove_duplicates.go
+++ b/test/e2e/fixtest/remove_duplicates.go
@@ -1,0 +1,77 @@
+package fixtest
+
+import (
+	"context"
+	"io"
+	"sort"
+
+	"go.mongodb.org/atlas/mongodbatlas"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// ZapLoggerFrom builds a zap.SugaredLogger from an IO Writer
+func ZapLoggerFrom(w io.Writer) *zap.SugaredLogger {
+	zcore := zapcore.NewCore(
+		zapcore.NewJSONEncoder(zap.NewDevelopmentEncoderConfig()),
+		zapcore.Lock(zapcore.AddSync(w)),
+		zap.NewAtomicLevel(),
+	)
+	return zap.New(zcore).Sugar()
+}
+
+// EnsureNoDuplicates removes projects with same name but different ID.
+// Atlas sometimes creates duplicate projects, we need our tests to defend
+// against that to avoid flaky tests
+func EnsureNoDuplicates(client *mongodbatlas.Client, logger *zap.SugaredLogger, projectName string) error {
+	found, err := listProjectsByName(client, projectName)
+	if err != nil || len(found) <= 1 {
+		return err
+	}
+	logger.Warnf("Found more than one project with name %q", projectName)
+	keep, rest := selectProject(found)
+	logger.Warnf("Will keep project ID %s as %s and remove the rest %v", keep.ID, projectName, ids(rest))
+	return removeProjects(client, rest)
+}
+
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -92,7 +92,7 @@ all: manager ## Build all binaries
 
 .PHONY: unit-test
 unit-test:
-	go test -race -cover ./pkg/...
+	go test -race -cover ./pkg/... ./test/helper/...
```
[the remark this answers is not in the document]
josvazg: let me check how the httptest in the stdlib is doing it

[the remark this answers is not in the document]
josvazg: Maybe we could figure out a way to not include the test code in the operator module and and find a proper home for both the tests of the tests and the mock code to be used by all tests.

test/helper/remove_duplicates_test.go
```diff
@@ -0,0 +1,105 @@
+package helper_test
```
igor-karpukhin, line 1: I'd move this to the "test" folder, not to the "helper"

test/helper/fake_projects.go
```diff
@@ -0,0 +1,103 @@
+package helper
```
igor-karpukhin, line 1: How about to create a directory with mocks? E.g. `test/mock/projects.go`

igor-karpukhin (changes requested):
Looks good in general, but let's organize the code a bit better

Makefile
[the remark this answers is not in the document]
josvazg: http://peter.bourgon.org/go-in-production/#testing-and-validation
I think we should move to build tags for the code under the `test/` folder. As a way to separate from code compile in the production binary or part of the importable module.
In fact I was using that on my previous project and, once setup, is totally transparent. It should even reduce compilation time a bit for the released code.

I can work on that on a separate PR. For this one I will just move code to proper places.

For this particular I will try the packagetest approach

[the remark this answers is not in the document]
josvazg: I tried to move it to `pkg/util/fixtest` lets see if it looks better like this.

test/helper/remove_duplicates_test.go
josvazg: moved it elsewhere 🤞

test/helper/fake_projects.go
josvazg: Moved to `pks/utils/atlastest` as it will need to be accessed by unit tests under ´pkg/´ as well.

pkg/util/atlastest/project.go
```diff
@@ -0,0 +1,103 @@
+package atlastest
+
+import (
+	"context"
+
+	"go.mongodb.org/atlas/mongodbatlas"
+)
+
+type ProjectsServiceMock struct {
+	GetAllProjectsFn func(context.Context, *mongodbatlas.ListOptions) (*mongodbatlas.Projects, *mongodbatlas.Response, error)
```
helderjs, line 10: The mocked func doesn't need to have the same signature as the interface, we usually don't care about context in our mock, for example.

```suggestion
	GetAllProjectsFn func() (*mongodbatlas.Projects, *mongodbatlas.Response, error)
```

pkg/util/atlastest/project.go
```diff
@@ -0,0 +1,103 @@
...
+	DeleteFn         func(context.Context, string) (*mongodbatlas.Response, error)
```
helderjs, line 11:
```suggestion
	DeleteFn         func(projectID string) (*mongodbatlas.Response, error)
```

pkg/util/fixtest/remove_duplicates.go
```diff
@@ -0,0 +1,65 @@
...
+	log.Printf("Found more than one project with name %q", projectName)
+	keep, rest := selectProject(found)
+	log.Printf("Will keep project ID %s as %s and remove the rest", keep.ID, projectName)
+	return removeProjects(client, rest)
+}
+
+func listProjectsByName(client *mongodbatlas.Client, projectName string, pageSize int) ([]*mongodbatlas.Project, error) {
+	projects, _, err := client.Projects.GetAllProjects(
+		context.Background(),
+		&mongodbatlas.ListOptions{
```
helderjs, line 32: I think this is redundant. Atlas has already a limit of projects per Org and this function doesn't enable page navigation. 
```suggestion
		&mongodbatlas.ListOptions{},
```

pkg/util/fixtest/remove_duplicates.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"context"
+	"log"
+	"sort"
+
+	"go.mongodb.org/atlas/mongodbatlas"
+)
+
+type Logger interface {
```
helderjs, line 11: While I understand this is a utility function, why we can't simply use Zap instead of defining a new "random" interface?

helderjs (changes requested):
Have you decided on which strategy to use to not ship this in the manager binary?

pkg/util/atlastest/project.go
josvazg: Right!
I will simplify that, makes sense.

pkg/util/fixtest/remove_duplicates.go
josvazg: ok will make it zap aware and remove this interface

josvazg:
> Have you decided on which strategy to use to not ship this in the manager binary?

I only see 2 options:

1. Move code back to the `test/` folder, but then we must unit test that folder too, as done before. Later on I will add build flags for the test folder so it will not be part of the module by default.
1. Use a build flag here directly now.

WDYT? any other ideas?

pkg/util/atlastest/project.go
josvazg: coming on next change

pkg/util/fixtest/remove_duplicates.go
josvazg: Done, we are now using a `*zap.SugaredLogger` here instead.

I am not sure I like it. The test using this code would want to dump to a GinkgoWriter, and connecting the zap logger to such writer was not obvious at all.

pkg/util/fixtest/remove_duplicates.go
josvazg: done in next change

josvazg:
>

josvazg:
> > Have you decided on which strategy to use to not ship this in the manager binary?
> 
> I only see 2 options:
> 
> 1. Move code back to the `test/` folder, but then we must unit test that folder too, as done before. Later on I will add build flags for the test folder so it will not be part of the module by default.
> 2. Use a build flag here directly now.
> 
> WDYT? any other ideas?

I think the best option is this:

1. Move the atlas mocks under test: `pkg/util/atlastest` -> `test/mocks/atlastest`
2. Move the fixtest also under test: `pkg/util/fixtest` -> `test/fixtest`
3. Turn the fixtest test into a ginkgo test to be run with the e2e tests.

WDYT?

josvazg:
Related [CLOUDP-192976](https://jira.mongodb.org/browse/CLOUDP-192976) Use buildtags for test code, avoid it in production

josvazg:
Caught a successful project de-duplication happening on a local test:
```
STEP: Create secret in the data prefix name @ 08/01/23 17:19:21.896
 STEP: Deploy Project my-project @ 08/01/23 17:19:21.908
 {"L":"WARN","T":"2023-08-01T17:19:32.731+0200","M":"Found more than one project with name \"encryption-at-rest-aws-Test Atlas Operator Project-c66205\""}
 {"L":"WARN","T":"2023-08-01T17:19:32.731+0200","M":"Will keep project ID 64c9227a4cb482626e2cf8a6 as encryption-at-rest-aws-Test Atlas Operator Project-c66205 and remove the rest [64c9227a90262a68f0dca4a8]"}
 STEP: Add cloud access role (AWS only) @ 08/01/23 17:19:53.578
 STEP: Create AWS role @ 08/01/23 17:19:53.578
...
Ran 1 of 57 Specs in 134.366 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 56 Skipped
PASS
```

igor-karpukhin (approved):

PR #1063, opened 2023-08-03
Pick random networks to fix flaky test

josvazg:
### All Submissions:


helderjs:
The reason we don't do that before is that we would easily reach the limit of VPCs per region, especially in AWS.
We can slice the subnet into more pieces and give one part to each test, this would require better creation of IPs in the GCP process too.
In the moment, creating a lot o VPCs will just bring the Ops guys to us, again.

josvazg:
I am closing, I cannot test locally. I think still need to setup some creds properly for local testing.

PR #1066, opened 2023-08-04
CLOUDP-170078: Add testing guide

josvazg:
### All Submissions:

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,0 +1,162 @@
+# Testing Guidelines
+
+## High level advice
+
+_Prefer reliable, fast and focused tests over end to end (e2e) testing_.
+
+### Start with a spec
+
+Before starting the implementation, **focus on more detailed behavior the operator code must have**. Having a clear written down specification of the change helps. Use this spec to inform which tests need to be satisfied.
+
+### Exploratory testing as needed
+
+For changes that involve dependencies, such as Atlas, a Cloud provider API or Kubernetes, it might be the case that just checking their API documentation might not be enough.
+
+It is possible that a test is needed to learn how the interaction between the operator and the dependency should play out in the context of the given specification.
+
+But **do not assume this test will become the final e2e test** for the change.
+
+We will refer to this testing phase as _exploratory testing_.
+
+### Prefer TDD
+
+**When the detailed behavior on part of the operator is clear, you can start writing the unit or integration test that covers it**. Unit tests are preferable whenever possible, as they will allow faster and more reliable development feedback loops. 
+
+**Keep your tests focused on a single check**. Note a single _exploratory test_ scenario will usually imply several steps or interactions to be tested individually.
+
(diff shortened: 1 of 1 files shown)


helderjs (approved):

docs/testing.md
```diff
@@ -0,0 +1,136 @@
...
+- Under `test-support/atlastest` you should find all mocked services already being unit tested using this approach.
+- And from them, you can also [find examples of how they are being used in unit tests](#sample-usage-mock).
+- If a service is mocked but the method you need is not configurable in the mock struct, it means it has not been used before in a test.
+- You can add service method implementations following the same structure as the ones already in place. [Add a new function to be set in the struct and called in the method, and just remember to remove the context first parameter](#sample-projects-mock), ignored in this mocks.
+- If a service is not mocked yet, create a new entry for it at `test-support/atlastest` following the same conventions as already present service mocks there.
+- You can use the [impl tool](https://github.com/josharian/impl) to generate the skeleton implementation. Alternativaly, you can try using your IDE such for filling the implemntation with “unimplemented” method calls so that the service interface is satisfied for the compiler. After that, just fill in the methods needed for the test and ignore the rest.
```
josvazg, line 69:
```suggestion
- You can use the [impl tool](https://github.com/josharian/impl) to generate the skeleton implementation. Alternativaly, you can try using your IDE for filling the implementation with “unimplemented” method calls so that the service interface is satisfied for the compiler. After that, just fill in the methods needed for the test and ignore the rest.
```

docs/testing.md
```diff
@@ -0,0 +1,156 @@
...
+
+It might even be the case now that the initial exploratory testing code must be converted to an e2e test. Note this should not be done lightly, as e2e tests add costs to the development process and CI pipeline feedback. If unit tests and/or integration tests cover the change or feature properly and the e2e feels redundant, it is preferably to not include it. On the other hand, if the e2e might catch issues that could have been missed otherwise, or need to detect dependency behavior changes, you should include the e2e, for those specific reasons, but not as the primary means to cover the change or feature testing.
+
+### <a name="prefer-external-tests"></a>Prefer external tests
+
+When writing tests for a feature or change, it is best if those can exercise the code as an external consumer, without any knowledge of or access to our code internal implementation. This makes for tests on behavior rather than implementation details. In Go this usually means prefer `package xxx_test` rather than `package xxx` directly.
```
igor-karpukhin, line 41: This may make unit tests more complicated and harder to write. I agree we should test the behaviour, but I wouldn't ban using the same package name

igor-karpukhin (approved):
Looks good 👍

docs/testing.md
josvazg: This is not a ban. It is a reminder of a best practice. Our code right now breaks this rule more often than it should, most probably due to writing the test after the implementation. But it is good to remember, when possible, it is best to test from the outside using the intended external interfaces.

See https://jdkaplan.dev/thinkin-logs/2021-10-07/ although not proposing about internal packages, at least not yet.

In other words, when there is no other way, just use `package xxx`, when creating new features, try TDD (write the test first) and use the external interface, so you should be able to test from `package xxx_test`.

cveticm (approved):
LGTM 👍

helderjs (approved):

PR #1068, opened 2023-08-07
CLOUDP-192976: Add int and e2e build tags

josvazg:
### All Submissions:

* [X] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [ ] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if there is one).
* [ ] Update docs/release-notes/release-notes.md if your changes should be included in the release notes for the next release.

PR #1075, opened 2023-08-09
CLOUDP-192976: Unit test everything but int and e2e

josvazg:
This is an alternative to #1068 avoiding the use of build tags to achieve the same main goal.

It is a simpler change and we keep the project build tag free using explicit env vars to activate the **int** and **e2e** tests instead.

Uses 2 tricks:
* Use `go list ./...` and remove the `test/int` and `test/e2e` tests from the unit test set explicitly.
* Allow `go test ./...` to run unmolested by requiring a env var to explicitly enable the int or e2e tests. 
  * Doing `AKO_INT_TEST=1` enables **int** tests.
  * `AKO_E2E_TEST=1` enables **e2e** tests.

### All Submissions:


josvazg:
I might send an update for this, I think we can even get `go test ./...` to work by skipping the e2e and int test unless they are passed some env var.

josvazg:
Note if we take this route I need to make some changes to the wording at #1066

Makefile
```diff
@@ -105,7 +107,7 @@ int-test: export ATLAS_PRIVATE_KEY=$(shell grep "ATLAS_PRIVATE_KEY" .actrc | cut
 int-test: export GINKGO_EDITOR_INTEGRATION="true"
 int-test: generate manifests ## Run integration tests. Sample with labels: `make int-test label=AtlasProject` or `make int-test label='AtlasDeployment && !slow'`
 	mkdir -p $(ENVTEST_ASSETS_DIR)
-	test -f $(ENVTEST_ASSETS_DIR)/setup-envtest.sh || curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.8.0/hack/setup-envtest.sh
+	test -v -f $(ENVTEST_ASSETS_DIR)/setup-envtest.sh || curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.8.0/hack/setup-envtest.sh
```
josvazg, line 110: not needed

josvazg:
Will wait for #1079 and rebase after it lands

test/helper/enable.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"os"
+	"strings"
+)
+
+func Enabled(envvar string) bool {
+	value := strings.ToLower(os.Getenv(envvar))
+	return value == "1"
+}
```
igor-karpukhin, line 11: nit: I think this can be just checked if the ENV is set or not set

igor-karpukhin (approved):
Looks good in general 👍 One nit regarding ENV variables verification

test/helper/enable.go
josvazg: I'd rather be more explicit here. What if you have `MYVAR=Off` or something, enabled would be confusing.

Still we can revisit once we do not need to run e2e testing by default for these kind of changes.

PR #1092, opened 2023-08-16
CLOUDP-195419: Improve Cloud Provider Access reconciliation flow

helderjs:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasproject/cloud_provider_access_test.go
+++ b/pkg/controller/atlasproject/cloud_provider_access_test.go
@@ -3,17 +3,15 @@ package atlasproject
 import (
 	"context"
 	"errors"
-	"github.com/stretchr/testify/assert"
 	"testing"
-
-	"go.uber.org/zap/zaptest"
-
-	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1/status"
 
 	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/pkg/api/v1"
 
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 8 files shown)


igor-karpukhin (approved):
Great work! 👍

pkg/controller/atlasproject/cloud_provider_access_test.go
```diff
@@ -26,30 +31,953 @@ func (c *cloudProviderAccessClient) GetRole(_ context.Context, _ string, _ strin
...
+					return &mongodbatlas.Response{}, nil
+				},
+			},
+		}
+		workflowCtx := &workflow.Context{
+			Client: atlasClient,
+		}
+
+		result, err := syncCloudProviderAccess(context.TODO(), workflowCtx, "projectID", cpas)
+		assert.NoError(t, err)
```
josvazg, line 135: nit

Should this be `require.NoError(...)` instead? Or is the nil error also part of the final status expected in a way?

josvazg (commented):
LGTM

Great and extensive unit tests!
This builds up more confidence that our code actually does what it "says on the tin".

josvazg (approved):

pkg/controller/atlasproject/cloud_provider_access_test.go
helderjs: As `require` fails the test immediately it's a bit annoying not knowing what went wrong in the first run. Given that, I think makes more sense to use it in places where we know, "if this doesn't work, everything else won't work and we should stop asap"

roothorp (approved):
😍

PR #1082, opened 2023-08-14
Release backup schedule and policy finalizers

josvazg:
At deployment removal time, backup schedules need to be released, if this is the latest deployment that was referencing them.

Same things transitively for policies: If the schedule was released and it was the only one referencing a policy, the policy finalizer must also be released.

Apart from the original issue, the lack of cleanup at deployment deletion time fixed by the new **cleanupBindings** call, there was another bug. The **policy status schedules ids** was being populated with references to policies instead of the schedule reference. That has been fixed as well and included in the new unit tests.

### All Submissions:

* [X] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [ ] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if there is one).
* [ ] Update docs/release-notes/release-notes.md if your changes should be included in the release notes for the next release.

PR #1093, opened 2023-08-16
CLOUPD-193335: Deletion Protection Serverless Private Endpoints

josvazg:
### All Submissions:

* ~~Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if there is one).~~
* ~~Update docs/release-notes/release-notes.md if your changes should be included in the release notes for the next release.~~

### DIFF
--- a/pkg/controller/atlasdeployment/atlasdeployment_controller.go
+++ b/pkg/controller/atlasdeployment/atlasdeployment_controller.go
@@ -153,7 +153,19 @@ func (r *AtlasDeploymentReconciler) Reconcile(context context.Context, req ctrl.
 	// Allow users to specify M0/M2/M5 deployments without providing TENANT for Normal and Serverless deployments
 	r.verifyNonTenantCase(deployment)
 
-	if result := r.checkDeploymentIsManaged(workflowCtx, context, log, project, deployment); !result.IsOk() {
+	// convertedDeployment is either serverless or advanced, deployment must be kept unchanged
+	// convertedDeployment is always a separate copy, to avoid changes on it to go back to k8s
+	convertedDeployment := deployment.DeepCopy()
+	if deployment.IsLegacyDeployment() {
+		if err := ConvertLegacyDeployment(&convertedDeployment.Spec); err != nil {
+			result = workflow.Terminate(workflow.Internal, err.Error())
+			log.Errorw("failed to convert legacy deployment", "error", err)
+			return result.ReconcileResult(), nil
+		}
+		convertedDeployment.Spec.DeploymentSpec = nil
+	}
+
+	if result := r.checkDeploymentIsManaged(workflowCtx, context, log, project, convertedDeployment); !result.IsOk() {
 		return result.ReconcileResult(), nil
 	}
 
@@ -162,44 +174,46 @@ func (r *AtlasDeploymentReconciler) Reconcile(context context.Context, req ctrl.
 		return result.ReconcileResult(), nil
 	}
 
-	err = customresource.ApplyLastConfigApplied(context, deployment, r.Client)
-	if err != nil {
-		result = workflow.Terminate(workflow.Internal, err.Error())
-		workflowCtx.SetConditionFromResult(status.DeploymentReadyType, result)
-		log.Error(result.GetMessage())
-
-		return result.ReconcileResult(), nil
-	}
-
-	if deployment.IsLegacyDeployment() {
-		if err := ConvertLegacyDeployment(&deployment.Spec); err != nil {
-			result = workflow.Terminate(workflow.Internal, err.Error())
-			log.Errorw("failed to convert legacy deployment", "error", err)
-			return result.ReconcileResult(), nil
-		}
-		deployment.Spec.DeploymentSpec = nil
-	}
-
-	if err := uniqueKey(&deployment.Spec); err != nil {
+	if err := uniqueKey(&convertedDeployment.Spec); err != nil {
 		log.Errorw("failed to validate tags", "error", err)
 		result := workflow.Terminate(workflow.Internal, err.Error())
 		workflowCtx.SetConditionFromResult(status.DeploymentReadyType, result)
 		return result.ReconcileResult(), nil
 	}
 
-	handleDeployment := r.selectDeploymentHandler(deployment)
-	if result, _ := handleDeployment(workflowCtx, project, deployment, req); !result.IsOk() {
+	handleDeployment := r.selectDeploymentHandler(convertedDeployment)
+	if result, _ := handleDeployment(context, workflowCtx, project, convertedDeployment, req); !result.IsOk() {
 		workflowCtx.SetConditionFromResult(status.DeploymentReadyType, result)
-		return result.ReconcileResult(), nil
+		return r.registerConfigAndReturn(workflowCtx, context, log, deployment, result), nil
 	}
 
-	if !deployment.IsServerless() {
-		if result := r.handleAdvancedOptions(workflowCtx, project, deployment); !result.IsOk() {
+	if !convertedDeployment.IsServerless() {
+		if result := r.handleAdvancedOptions(workflowCtx, project, convertedDeployment); !result.IsOk() {
 			workflowCtx.SetConditionFromResult(status.DeploymentReadyType, result)
-			return result.ReconcileResult(), nil
+			return r.registerConfigAndReturn(workflowCtx, context, log, deployment, result), nil
 		}
 	}
-	return workflow.OK().ReconcileResult(), nil
+
+	return r.registerConfigAndReturn(workflowCtx, context, log, deployment, workflow.OK()), nil
+}
+
+func (r *AtlasDeploymentReconciler) registerConfigAndReturn(
+	workflowCtx *workflow.Context,
+	context context.Context,
+	log *zap.SugaredLogger,
+	deployment *mdbv1.AtlasDeployment, // this must be the original non converted deployment
+	result workflow.Result) ctrl.Result {
+	if result.IsOk() || result.IsInProgress() {
+		err := customresource.ApplyLastConfigApplied(context, deployment, r.Client)
+		if err != nil {
+			alternateResult := workflow.Terminate(workflow.Internal, err.Error())
+			workflowCtx.SetConditionFromResult(status.DeploymentReadyType, alternateResult)
+			log.Error(result.GetMessage())
+
+			return result.ReconcileResult()
+		}
+	}
+	return result.ReconcileResult()
 }
 
 func (r *AtlasDeploymentReconciler) verifyNonTenantCase(deployment *mdbv1.AtlasDeployment) {
@@ -231,19 +245,16 @@ func (r *AtlasDeploymentReconciler) checkDeploymentIsManaged(
 	project *mdbv1.AtlasProject,
 	deployment *mdbv1.AtlasDeployment,
 ) workflow.Result {
-	advancedDeployment := deployment
 	if deployment.IsLegacyDeployment() {
-		advancedDeployment = deployment.DeepCopy()
-		if err := ConvertLegacyDeployment(&advancedDeployment.Spec); err != nil {
-			result := workflow.Terminate(workflow.Internal, err.Error())
-			log.Errorw("failed to temporary convert legacy deployment", "error", err)
-			return result
-		}
-		advancedDeployment.Spec.DeploymentSpec = nil
+		result := workflow.Terminate(workflow.Internal, "ownership check expected a converted deployment, not a legacy one")
+		workflowCtx.SetConditionFromResult(status.DatabaseUserReadyType, result)
+		log.Error(result.GetMessage())
+
+		return result
 	}
 
 	owner, err := customresource.IsOwner(
-		advancedDeployment,
+		deployment,
 		r.ObjectDeletionProtection,
 		customresource.IsResourceManagedByOperator,
 		managedByAtlas(context, workflowCtx.Client, project.ID(), log),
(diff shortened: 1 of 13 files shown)


josvazg:
Still a draft as it should go after bugfix #1082

josvazg:
debugging

josvazg:
Even after a rebase the issues persist. Seems `main` is not affected.
Debugging the database user integration test failure first.

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -188,18 +179,38 @@ func (r *AtlasDeploymentReconciler) Reconcile(context context.Context, req ctrl.
...
+	return r.registerConfigAndReturn(workflowCtx, context, log, deployment, workflow.OK()), nil
+}
+
+func (r *AtlasDeploymentReconciler) registerConfigAndReturn(
+	workflowCtx *workflow.Context,
+	context context.Context,
+	log *zap.SugaredLogger,
+	deployment *mdbv1.AtlasDeployment,
+	result workflow.Result) ctrl.Result {
+	if result.IsOk() || result.IsInProgress() {
```
josvazg, line 203: cc @igor-karpukhin 
I think this was the issue before I was ONLY saving the last config if all went fine till the end of the reconcile method and was going to return OK. But there are times when the `handleDeployment` or `handleAdvancedOptions` might need to exit early but still it is not due to an abrupt failure or interruption of the reconcile loop, but just because the action is still "in progress" and this is has to be revisited later.

My reasoning here is that if the exit at those points is either `OK` or `In Progress` it means that the operator has already taken ownership of the Atlas counterpart and therefore we need to save the current config as applied, even if it is still being applied (in progress).

I hope to see a lot more green in the CI tests after this fix.

josvazg:
Debugging some panics in CI tests.

josvazg:
I think there are more usages of `.Spec.DeploymentSpec.` leading to flaky CI failures. And there seems to be a persistent helm upgrade issue. Still do not understand how my changes are related to that, but will keep on debugging and sending fixes in this branch.

josvazg:
@igor-karpukhin gave me the right hint here.
Somehow my code must be saving the converted deployment to Kubernetes, changing the original spec. We do the conversion here:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/CLOUDP-193335/deletion-protection-spe/pkg/controller/atlasdeployment/atlasdeployment_controller.go#L165
```go
if deployment.IsLegacyDeployment() {
		if err := ConvertLegacyDeployment(&deployment.Spec); err != nil {
			result = workflow.Terminate(workflow.Internal, err.Error())
			log.Errorw("failed to convert legacy deployment", "error", err)
		}
		deployment.Spec.DeploymentSpec = nil
	}
```

But the key point is that **conversion must never be seen in Kubernetes**. It is ONLY used in memory for convenience so we always handle only serverless deployments and advanced. aAs old deployments get converted to their advanced equivalent.

My code somewhere must be saving the converted spec to Kubernetes, so the rest of the test code and even Helm fail as they assume a default configuration which is missing, as it was converted to advanced. This conversion is wrong, it should not have happened.

josvazg:
Seems all issues, including the helm error and the panics were due to `ApplyLastConfigApplied` adding the annotation and updating the **converted deployment in Kubernetes, instead of the original** deployment without conversion. This update was not expected and all tests break accordingly.

The fix is to actually make sure `ApplyLastConfigApplied` is called with the original (non converted) deployment object.

josvazg:
2 errors remain that might not be flaky:
```
Summarizing 2 Failures:
  [FAIL] AtlasDeployment Create the advanced deployment with enabled autoscaling [It] Should Succeed [int, AtlasDeployment, deployment-non-backups]
  /home/runner/work/mongodb-atlas-kubernetes/mongodb-atlas-kubernetes/test/int/deployment_test.go:209
  [FAIL] AtlasDeployment Create the advanced deployment & change the InstanceSize [It] Should Succeed [int, AtlasDeployment, deployment-non-backups]
  /home/runner/work/mongodb-atlas-kubernetes/mongodb-atlas-kubernetes/test/int/deployment_test.go:868
```
Debugging those now.

josvazg:
> 2 errors remain that might not be flaky:
> 
> ```
> Summarizing 2 Failures:
>   [FAIL] AtlasDeployment Create the advanced deployment with enabled autoscaling [It] Should Succeed [int, AtlasDeployment, deployment-non-backups]
>   /home/runner/work/mongodb-atlas-kubernetes/mongodb-atlas-kubernetes/test/int/deployment_test.go:209
>   [FAIL] AtlasDeployment Create the advanced deployment & change the InstanceSize [It] Should Succeed [int, AtlasDeployment, deployment-non-backups]
>   /home/runner/work/mongodb-atlas-kubernetes/mongodb-atlas-kubernetes/test/int/deployment_test.go:868
> ```
> 
> Debugging those now.

The problem was due to `handleAutoscaling` making changes in the Kubernetes spec for advanced deployment here:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/pkg/controller/atlasdeployment/advanced_deployment.go#L192
```go
func handleAutoscaling(ctx *workflow.Context, desiredDeployment *mdbv1.AdvancedDeploymentSpec, currentDeployment *mongodbatlas.AdvancedCluster) error {
	...
	syncInstanceSize := func(s *mdbv1.Specs, as *mdbv1.AdvancedAutoScalingSpec) error {
		if s != nil {
			...

			if isInstanceSizeTheSame(currentDeployment, size) {
				size = ""
			}

			s.InstanceSize = size
```

Those changes were being saved to the Kubernetes state when updating the deployment after (re)setting the previous config here:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/pkg/controller/customresource/protection.go#L50
```go
func ApplyLastConfigApplied(ctx context.Context, resource mdbv1.AtlasCustomResource, k8sClient client.Client) error {
...
	return k8sClient.Update(ctx, resource, &client.UpdateOptions{})
}
```

This was only happening for advanced cluster, because I was NOT making a copy of the Kubernetes deployment structure unless the conversion was actually needed.

The fix was to simply always make the converted deployment variable a separate copy, so that any changes in the reconcile loop will not be reflected back into the Kubernetes etcd status.

A separate matter is whether or not those ephemeral changes to the deployment struct make actual sense, but that is out of the scope of this PR.

josvazg:
Current persistent failures are in encryption-at-rest a well know flaky test that, AFAIK, is totally unrelated to the changes here:

`[FAIL] Encryption at REST test Encryption at rest for AWS, GCP, Azure [It] Test[encryption-at-rest-aws]: Can add Encryption at Rest to AWS project [encryption-at-rest, encryption-at-rest-aws]`

Will be debugging it, as it is blocking me.

josvazg:
> 



> Current persistent failures are in encryption-at-rest a well know flaky test that, AFAIK, is totally unrelated to the changes here:
> 
> `[FAIL] Encryption at REST test Encryption at rest for AWS, GCP, Azure [It] Test[encryption-at-rest-aws]: Can add Encryption at Rest to AWS project [encryption-at-rest, encryption-at-rest-aws]`
> 
> Will be debugging it now.

Created a follow up issue for that.

Now the test passed, we are ready for review.

igor-karpukhin:
@josvazg this is done because of the way we operate with autoscaling. For example, if you set instanceSize to be M10 and enabled autoscaling from M10 to M40, later, after cluster is autoscaled to M30, if you try to modify the `AtlasDeployment` resource, the instanceSize is going to be M10, and we send this M10 to Atlas. Which is wrong. That's why if instance size is equal to the one in Atlas and autoscaling is enabled, we remove this field before sending PATCH request.

josvazg:
> @josvazg this is done because of the way we operate with autoscaling. For example, if you set instanceSize to be M10 and enabled autoscaling from M10 to M40, later, after cluster is autoscaled to M30, if you try to modify the `AtlasDeployment` resource, the instanceSize is going to be M10, and we send this M10 to Atlas. Which is wrong. That's why if instance size is equal to the one in Atlas and autoscaling is enabled, we remove this field before sending PATCH request.

Thanks for the explanation. Then it means that ephemeral changes to the k8s struct is like an intermediate representation to ease syncing with Atlas. Makes sense.

josvazg:
Rebased

pkg/controller/atlasdeployment/serverless_private_endpoint.go
```diff
@@ -57,6 +80,82 @@ func ensureServerlessPrivateEndpoints(service *workflow.Context, groupID string,
...
+		return false, err
+	}
+
+	if len(existingPE) == 0 {
+		return true, nil
+	}
+
+	logger := service.Log
+	prevCfg := prevPEConfig(latestConfig)
+	if matchingPEs(logger, deployment.Spec.ServerlessSpec.PrivateEndpoints, existingPE) ||
```
helderjs, line 108: I think relying simply on the name might lead to mistakes. Uniqueness is not enforced on the Atlas side once the name is stored in the "comment" field and the duplication check uses the name and assumes the user will not play around with items in CR definition (move them around, also can happen with auto-generated data). I'd suggest compare the relevant fields `endpointID` and `ipAddress` (for azure)

helderjs (changes requested):

pkg/controller/atlasdeployment/serverless_private_endpoint.go
josvazg: I worry that might make this more brittle. We also do not want reconciliations to stop because we missed some intermediate change.

Will implement it as suggested, but we should sync on "sub-object grip" for this feature. We do not have a way to grip a sub object once it has been started to be managed, like we do with full blown CRs. I think that is problematic.

josvazg: Reverted to just check the name/comment, as discussed elsewhere.

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -153,53 +153,67 @@ func (r *AtlasDeploymentReconciler) Reconcile(context context.Context, req ctrl.
...
 		return result.ReconcileResult(), nil
 	}
 
-	handleDeployment := r.selectDeploymentHandler(deployment)
-	if result, _ := handleDeployment(workflowCtx, project, deployment, req); !result.IsOk() {
+	handleDeployment := r.selectDeploymentHandler(convertedDeployment)
+	if result, _ := handleDeployment(context, workflowCtx, project, convertedDeployment, req); !result.IsOk() {
 		workflowCtx.SetConditionFromResult(status.DeploymentReadyType, result)
-		return result.ReconcileResult(), nil
+		return r.registerConfigAndReturn(workflowCtx, context, log, deployment, result), nil
```
helderjs, line 187: Setting the config for non-successful operations may cause trouble in the next reconciliation once we will be comparing to a "wrong" configuration. I think we should separate the intermediate state "in progress"(creating/updating) from actual errors when saving the config. The same applies to serverless.

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -242,7 +242,7 @@ func (r *AtlasProjectReconciler) ensureDeletionFinalizer(ctx context.Context, wo
 	if !project.GetDeletionTimestamp().IsZero() {
 		if customresource.HaveFinalizer(project, customresource.FinalizerLabel) {
 			if customresource.IsResourceProtected(project, r.ObjectDeletionProtection) {
-				log.Info("Not removing Atlas database user from Atlas as per configuration")
+				log.Info("Not removing Project from Atlas as per configuration")
```
helderjs, line 245: This change is already in the main, is your branch updated?

pkg/controller/atlasproject/maintenancewindow.go
```diff
@@ -233,23 +233,10 @@ func isMaintenanceWindowConfigEqual(akoMWindow project.MaintenanceWindow, atlasM
...
-
-	if akoMWindow.AutoDefer != *atlasMWindow.AutoDeferOnceEnabled {
-		return false
-	}
-
-	return true
+	return akoMWindow.DayOfWeek == atlasMWindow.DayOfWeek &&
+		akoMWindow.HourOfDay == *atlasMWindow.HourOfDay &&
+		akoMWindow.StartASAP == *atlasMWindow.StartASAP &&
+		akoMWindow.AutoDefer == *atlasMWindow.AutoDeferOnceEnabled
```
helderjs, line 239: This change is already in the main, is your branch updated?

helderjs (changes requested):
I noticed something strange with your rebase. Changes already in the main branch are showing as new in your PR, I pointed out 2 but there are more, even entire files.

josvazg:
> I noticed something strange with your rebase. Changes already in the main branch are showing as new in your PR, I pointed out 2 but there are more, even entire files.

Thanks for the heads up, checking...

pkg/controller/atlasdeployment/atlasdeployment_controller.go
josvazg: It is separated:
```go
func (r *AtlasDeploymentReconciler) registerConfigAndReturn(
...
	if result.IsOk() || result.IsInProgress() {
		err := customresource.ApplyLastConfigApplied(context, deployment, r.Client)
...
```

Errors will not save the config, OK and in progress will.

pkg/controller/atlasproject/atlasproject_controller.go
josvazg: will double check, this is probably because I wanted to revert o my prior changes before yesterdays fix.

pkg/controller/atlasproject/maintenancewindow.go
josvazg: will check it as well.

helderjs (approved):

pkg/controller/atlasproject/atlasproject_controller.go
josvazg: fixed

pkg/controller/atlasproject/maintenancewindow.go
josvazg: fixed

PR #1084, opened 2023-08-14
CLOUDP-194416: Isolate CI unit tests to always run

josvazg:
Both `lint` and `unit-test`
- Run now as part of the `Test` workflow.
- Will **always** run on `push`, `PR` and workflow dispatch. Nothing can stop them.
- Are required to run any cloud bound tests: `integration' and `e2e` tests. 

Additionally, `e2e` and `integration` tests will NOT run on draft by default, so this also completes [CLOUDP-192963: Atlas CI/CD: Do not run CI/CD pipeline for draft PRs](https://jira.mongodb.org/browse/CLOUDP-192963)

### All Submissions:

* [X] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [ ] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if there is one).
* [ ] Update docs/release-notes/release-notes.md if your changes should be included in the release notes for the next release.

PR #1095, opened 2023-08-17
CLOUDP-194416: Revamp main CI test workflows

josvazg:
- Isolate **Unit Tests** and make them run always. From #1084 
- Separate expensive **Cloud Tests** and make them run in a single queue ONLY when production code was changed.
  - Unit test changes do not count as a production code change and are self-covered by unit testing.  
- Fix **goimports** and cancellations in `openshift-upgrade-test` #1094 

### All Submissions:


.github/workflows/test.yml
```diff
@@ -7,6 +7,7 @@ on:
     paths-ignore:
       - 'docs/**'
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review, converted_to_draft]
```
igor-karpukhin, line 10: The last element: `converted_to_draft`. Does that mean it will run tests for Draft PRs?

josvazg: It means it will re-evaluate the workflow when you make it a draft. Then it will disable cloud tests, due to the check we have below.

igor-karpukhin (approved):
Looks good 👍

helderjs (approved):

PR #1096, opened 2023-08-17
CLOUDP-195522: Default cluster bool to false

roothorp:
### All Submissions:


josvazg (approved):
LGTM
Unit tests about it would be nice

roothorp:
@josvazg is this sort of thing OK? I'll admit I'm probably not as familiar with unit testing as I should be - this is just based off what @igor-karpukhin has done for the DataFederation CRD.

josvazg:
> @josvazg is this sort of thing OK? I'll admit I'm probably not as familiar with unit testing as I should be - this is just based off what @igor-karpukhin has done for the DataFederation CRD.
> 
> If so, I can add the equivalent test for the other direction (from Atlas) tomorrow and we should be good to merge :)

LGTM, just one nit pick. As you only have 1 test case you could drop the test case table completely and its loop. Replace it with a `t.Run("title"...` so that your test is still titled. Alternatively you could come with more useful tests cases to add so that the table and loop are actually useful.

igor-karpukhin (approved):
Looks good 👍

helderjs (approved):

PR #1099, opened 2023-08-21
CLOUDP-196136: Use UBI micro base image instead of minimal

roothorp:
Use Red Hat's Micro Universal Base Image, rather than Minimal. This cuts the compressed size (download from Dockerhub) from ~90mb to ~30mb, and the uncompressed (on disk) from ~260mb to ~77mb. 

This should also reduce security exposure through simply having less packages in the image; [micro has 20](https://catalog.redhat.com/software/containers/ubi9/ubi-micro/615bdf943f6014fa45ae1b58?container-tabs=packages) compared to [minimal's 102](https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5?container-tabs=packages).

### All Submissions:

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -7,6 +7,7 @@ on:
         type: boolean
         required: false
         default: false
+  workflow_dispatch:
 
 jobs:
   prepare-e2e:
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,18 +27,7 @@ ENV TARGET_OS=${TARGETOS}
 
 RUN make manager
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2
-
-RUN microdnf install -y yum &&\
-  yum -y update &&\
-  yum -y upgrade &&\
-  yum clean all &&\
-  dnf remove python3-pip -y &&\
-  microdnf clean all
-
-#FROM registry.access.redhat.com/ubi8/ubi
-#
-#RUN dnf -y update-minimal --security --sec-severity=Important --sec-severity=Critical
+FROM registry.access.redhat.com/ubi9/ubi-micro:9.2
 
 LABEL name="MongoDB Atlas Operator" \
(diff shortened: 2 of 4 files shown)


josvazg:
LGMT if it passes tests.
Note that right now, draft will not run cloud tests.

josvazg:
It did not run the tests because it does not change production code. We can force with a tag.

josvazg:
Looking good now!

.github/workflows/test-e2e.yml
```diff
@@ -7,6 +7,7 @@ on:
         type: boolean
         required: false
         default: false
+  workflow_dispatch:
```
josvazg, line 10: @igor-karpukhin we will need to add this line here and in `int-test.yml` so that we can run the tests manually for your PR.

Dockerfile
```diff
@@ -56,6 +46,7 @@ LABEL name="MongoDB Atlas Operator" \
 WORKDIR /
 COPY --from=builder /workspace/bin/manager .
 COPY hack/licenses licenses
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
```
roothorp, line 49: The micro image doesn't include CA certs (nor a package manager to retrieve them) so we get them from the builder. Both RHEL (UBI) and Debian (golang) use Mozilla's, so it shouldn't make a difference.

Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN go install golang.org/x/tools/cmd/goimports@latest
 COPY cmd/manager/main.go cmd/manager/main.go
 COPY .git/ .git/
 COPY pkg/ pkg/
+COPY internal/ internal/
```
roothorp, line 17: This is needed for the manager to compile; it contains the new mocks @helderjs added.

helderjs (approved):

PR #1100, opened 2023-08-21
CLOUDP-175080: Deletion protection for project resources (final)

helderjs:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasproject/auditing.go
+++ b/pkg/controller/atlasproject/auditing.go
@@ -2,7 +2,13 @@ package atlasproject
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"reflect"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/pkg/controller/customresource"
 
 	"go.mongodb.org/atlas/mongodbatlas"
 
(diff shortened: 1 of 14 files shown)


igor-karpukhin (approved):
Massive work here! Good job! 👍

pkg/controller/atlasproject/auditing.go
```diff
@@ -59,12 +81,24 @@ func auditingInSync(atlas *mongodbatlas.Auditing, spec *v1.Auditing) bool {
...
+	if !isAuditingEmpty(spec) {
+		specAsAtlas = spec.ToAtlas()
+	}
+
+	if isAuditingEmpty(atlas) {
+		atlas = &mongodbatlas.Auditing{
+			AuditAuthorizationSuccess: toptr.MakePtr(false),
+			Enabled:                   toptr.MakePtr(false),
+		}
 	}
```
josvazg, line 98: I do not understand this change, can you elaborate?

helderjs: While unlikely to happen, this is a protection once the response from the client is a pointer. Given that, if for any reason it returns no error and nil as a response, we assume default values.

pkg/controller/atlasproject/maintenancewindow.go
```diff
@@ -169,3 +193,65 @@ func toggleAutoDeferInAtlas(client mongodbatlas.Client, projectID string) workfl
...
+
+	if akoMWindow.StartASAP != *atlasMWindow.StartASAP {
+		return false
+	}
+
+	if akoMWindow.AutoDefer != *atlasMWindow.AutoDeferOnceEnabled {
+		return false
+	}
+
+	return true
```
josvazg, line 252: Why not this instead?

```suggestion
	return akoMWindow.DayOfWeek == atlasMWindow.DayOfWeek &&
		akoMWindow.HourOfDay == *atlasMWindow.HourOfDay &&
		akoMWindow.StartASAP != *atlasMWindow.StartASAP &&
		akoMWindow.AutoDefer != *atlasMWindow.AutoDeferOnceEnabled
```

For me it is easier to read as it separates the pointer value fixes above from the actual equals check.

pkg/controller/atlasproject/maintenancewindow_test.go
```diff
@@ -1,13 +1,50 @@
...
+type maintenanceWindowClient struct {
+	GetFunc func(projectID string) (*mongodbatlas.MaintenanceWindow, *mongodbatlas.Response, error)
+}
+
+func (c *maintenanceWindowClient) Get(_ context.Context, projectID string) (*mongodbatlas.MaintenanceWindow, *mongodbatlas.Response, error) {
+	return c.GetFunc(projectID)
+}
+
+func (c *maintenanceWindowClient) Update(_ context.Context, _ string, _ *mongodbatlas.MaintenanceWindow) (*mongodbatlas.Response, error) {
+	return nil, nil
```
josvazg, line 33: You can use [impl](https://github.com/josharian/impl) which will by default panic unimplemented here. This is what we recommend in (docs/testing.go)[https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/docs/testing.md]

Not a big deal, but the more similar these mocks are, the less refactor they will need when we move them to their own package.

helderjs: Here I think it's not necessary to apply this once there's no one else consuming these mocks. Next week when I'm back I'll be addressing the ticket to move all the mocks to its own pkg. Also, I will provide "full-featured" mocks together.

pkg/controller/atlasproject/teams.go
```diff
@@ -243,3 +267,85 @@ func hasTeamRolesChanged(current []string, desired []v1.TeamRole) bool {
...
+		if atlasAssignedTeam != nil {
+			atlasAssignedTeamsInfo = append(
+				atlasAssignedTeamsInfo,
+				assignedTeamInfo{
+					ID:    atlasAssignedTeam.TeamID,
+					Roles: atlasAssignedTeam.RoleNames,
+				},
+			)
+		}
+	}
```
josvazg, line 309: nitpick
This seems to be a format conversion. Moving to its own function could made this more readable.

josvazg, line 309: nitpick
This seems to be a format conversion. Moving to its own function could made this more readable.

```suggestion
	atlasAssignedTeamsInfo := atlasAssignedTeamsToK8s(atlasAssignedTeams.Results)
```

josvazg (approved):
LGTM

All my comments are non blocking nitpicks

PR #1120, opened 2023-09-05
CLOUDP-197433: Fix Encryption at Rest flakiness

helderjs:
### All Submissions:


roothorp (approved):
Seems to have worked! My only concern here would be any quotas on how many keys we have lying around if we're forced to let them sit for a week before they're deleted (and we're creating a new one for each test) but [it looks like the quota is 100k](https://docs.aws.amazon.com/kms/latest/developerguide/resource-limits.html) so we should be ok! 

Also nice to have one less env var required for tests :)

LGTM

igor-karpukhin (approved):
Good job 👍

PR #1124, opened 2023-09-07
CLOUDP-195050: Atlas Federated Authentication configuration

igor-karpukhin:
### All Submissions:

* [x] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [ ] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if there is one).
* [ ] Update docs/release-notes/release-notes.md if your changes should be included in the release notes for the next release.

PR #1135, opened 2023-09-20
CLOUDP-199666: Support Atlas for government

helderjs:
### All Submissions:

### DIFF
--- a/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
+++ b/config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
@@ -887,6 +887,17 @@ spec:
                       type: string
                   type: object
                 type: array
+              regionUsageRestrictions:
+                default: NONE
+                description: RegionUsageRestrictions designate the project's AWS region
+                  when using Atlas for Government. This parameter should not be used
+                  with commercial Atlas. In Atlas for Government, not setting this
+                  field (defaulting to NONE) means the project is restricted to COMMERCIAL_FEDRAMP_REGIONS_ONLY
+                enum:
+                - NONE
+                - GOV_REGIONS_ONLY
+                - COMMERCIAL_FEDRAMP_REGIONS_ONLY
+                type: string
               settings:
                 description: Settings allow to set Project Settings for the project
                 properties:
--- a/pkg/api/v1/atlasproject_types.go
+++ b/pkg/api/v1/atlasproject_types.go
@@ -46,6 +46,14 @@ type AtlasProjectSpec struct {
 	// Name is the name of the Project that is created in Atlas by the Operator if it doesn't exist yet.
 	Name string `json:"name"`
 
+	// RegionUsageRestrictions designate the project's AWS region when using Atlas for Government.
+	// This parameter should not be used with commercial Atlas.
+	// In Atlas for Government, not setting this field (defaulting to NONE) means the project is restricted to COMMERCIAL_FEDRAMP_REGIONS_ONLY
+	// +kubebuilder:validation:Enum=NONE;GOV_REGIONS_ONLY;COMMERCIAL_FEDRAMP_REGIONS_ONLY
+	// +kubebuilder:default:=NONE
+	// +optional
+	RegionUsageRestrictions string `json:"regionUsageRestrictions,omitempty"`
+
 	// ConnectionSecret is the name of the Kubernetes Secret which contains the information about the way to connect to
 	// Atlas (organization ID, API keys). The default Operator connection configuration will be used if not provided.
 	// +optional
--- a/pkg/controller/customresource/customresource.go
+++ b/pkg/controller/customresource/customresource.go
@@ -2,6 +2,8 @@ package customresource
 
 import (
 	"context"
+	"net/url"
+	"strings"
 
 	"fmt"
 
@@ -24,11 +26,11 @@ const (
 	ReconciliationPolicyAnnotation = "mongodb.com/atlas-reconciliation-policy"
 	ResourceVersion                = "app.kubernetes.io/version"
 	ResourceVersionOverride        = "mongodb.com/atlas-resource-version-policy"
-
-	ResourcePolicyKeep       = "keep"
-	ResourcePolicyDelete     = "delete"
-	ReconciliationPolicySkip = "skip"
-	ResourceVersionAllow     = "allow"
+	ResourcePolicyKeep             = "keep"
+	ResourcePolicyDelete           = "delete"
+	ReconciliationPolicySkip       = "skip"
+	ResourceVersionAllow           = "allow"
+	govAtlasDomain                 = "mongodbgov.com"
 )
 
 // PrepareResource queries the Custom Resource 'request.NamespacedName' and populates the 'resource' pointer.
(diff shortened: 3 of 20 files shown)


config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
```diff
@@ -887,6 +887,17 @@ spec:
                       type: string
                   type: object
                 type: array
+              regionUsageRestrictions:
+                default: NONE
+                description: RegionUsageRestrictions designate the project's AWS region
+                  when using Atlas fo Government. This parameter should not be used
+                  with commercial Atlas In Atlas for Government, not setting this
+                  field (defaulting to NONE) means the project is restricted to COMMERCIAL_FEDRAMP_REGIONS_ONLY
```
josvazg, line 895: > RegionUsageRestrictions designate the project's AWS region when using Atlas fo Government.

It does not designate a particular AWS region, right? I guess it limits the choice of AWS regions.

> This parameter should not be used with commercial Atlas In Atlas for Government

Strange, one would assume setting this field to something other than NONE would be precisely when the Atlas cloud setup is government related.

> not setting this field (defaulting to NONE) means the project is restricted to COMMERCIAL_FEDRAMP_REGIONS_ONLY

Are we saying here NONE becomes same as COMMERCIAL_FEDRAMP_REGIONS_ONLY??

For me the bare enum seems intuitive, but after reading the description I believe my intuitions might be all wrong.


Can you elaborate more about this setting?

config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
```diff
@@ -887,6 +887,17 @@ spec:
                       type: string
                   type: object
                 type: array
+              regionUsageRestrictions:
+                default: NONE
+                description: RegionUsageRestrictions designate the project's AWS region
+                  when using Atlas fo Government. This parameter should not be used
```
josvazg, line 893:
```suggestion
                  when using Atlas for Government. This parameter should not be used
```

pkg/api/v1/atlasproject_types.go
```diff
@@ -46,6 +46,14 @@ type AtlasProjectSpec struct {
 	// Name is the name of the Project that is created in Atlas by the Operator if it doesn't exist yet.
 	Name string `json:"name"`
 
+	// RegionUsageRestrictions designate the project's AWS region when using Atlas fo Government.
```
josvazg, line 49:
```suggestion
	// RegionUsageRestrictions designate the project's AWS region when using Atlas for Government.
```

pkg/controller/customresource/customresource.go
```diff
@@ -145,3 +148,31 @@ func SetAnnotation(resource mdbv1.AtlasCustomResource, key, value string) {
...
+func IsGov(domain string) bool {
+	domainURL, err := url.Parse(domain)
+	if err != nil {
+		return false
+	}
+
+	return strings.HasSuffix(domainURL.Hostname(), govAtlasDomain)
+}
+
+func IsResourceSupportedByGov(resource mdbv1.AtlasCustomResource, domain string) bool {
```
josvazg, line 161: Maybe we can use a more generic name here:
```suggestion
func IsResourceSupported(resource mdbv1.AtlasCustomResource, domain string) bool {
```

config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
josvazg, line 895:
```suggestion
                  with commercial Atlas.
                  In Atlas for Government, not setting this field (defaulting to NONE) is equivalent to COMMERCIAL_FEDRAMP_REGIONS_ONLY
```

pkg/controller/customresource/customresource.go
helderjs: I prefer to keep it as it is because the checks are about  cloud Gov

config/crd/bases/atlas.mongodb.com_atlasprojects.yaml
helderjs: I prefer to make it clear in the description because no setting means `COMMERCIAL_FEDRAMP_REGIONS_ONLY`, setting it to `NONE` also means `COMMERCIAL_FEDRAMP_REGIONS_ONLY`.

pkg/controller/customresource/customresource.go
josvazg: My point is that this naming is misleading, almost "a lie". For instance, I pass a resource that is NOT supported by Gov and I get `true` because I am NOT in a Gov region, but the resource is still not supported by Gov.

This would be clearer and truthful:
```go
func IsResourceSupportedInDomain(resource mdbv1.AtlasCustomResource, domain string) bool {
	if IsGov(domain) {
		return isResourceSupportedByGov(resource)
	}
	// TODO add any other special env checks needed in teh future
	return true
}

func isResourceSupportedByGov(resource mdbv1.AtlasCustomResource) bool {
	switch atlasResource := resource.(type) {
	case *mdbv1.AtlasProject, *mdbv1.AtlasTeam, *mdbv1.AtlasBackupSchedule, *mdbv1.AtlasBackupPolicy, *mdbv1.AtlasDatabaseUser:
		return true
	case *mdbv1.AtlasDataFederation:
		return false
	case *mdbv1.AtlasDeployment:
		if atlasResource.Spec.ServerlessSpec == nil {
			return true
		}
	}
	return false
}
```

test/e2e/k8s/operator.go
```diff
@@ -150,6 +152,22 @@ func BuildManager(initCfg *Config) (manager.Manager, error) {
...
+		AtlasDomain:                 config.AtlasDomain,
+		GlobalAPISecret:             config.GlobalAPISecret,
+		EventRecorder:               mgr.GetEventRecorderFor("AtlasDataFederation"),
+		GlobalPredicates:            globalPredicates,
+		ObjectDeletionProtection:    config.ObjectDeletionProtection,
+		SubObjectDeletionProtection: config.SubObjectDeletionProtection,
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "AtlasDataFederation")
+		return nil, err
+	}
```
josvazg, line 169: I guess this is here because of concurrency with PR #1124 **CLOUDP-195050: Atlas Federated Authentication configuration**

If it requires those changes then it should be rebased after that one.

josvazg (commented):
LGTM except for that misleading naming issue

Not sure: Does this PR depend on other (#1124 ?) to merge first?

josvazg (approved):
In case there are no deps

test/e2e/k8s/operator.go
helderjs: It has nothing to do with Federated Authentication. This adds the Data Federation controller that was missing in the e2e setup. Related tests using Data FEderation CR would never be handled because of this, and actually, the existing Data Federation e2e test has never run.

pkg/controller/customresource/customresource.go
helderjs: Splitting a simple conditional into 2 functions seems overkill as it tries to solve a problem we don't have, as we don't expect to have a feature flag behavior in the future. However, I'm okay with renaming the existing function to `IsResourceSupportedInDomain` if you believe it's more meaningful

josvazg:
👍

igor-karpukhin (approved):
Good job! 👍

PR #1168, opened 2023-10-12
CLOUDP-204494: Improve snapshot distribution management

helderjs:
### All Submissions:

### DIFF
--- a/pkg/api/v1/atlasdeployment_types.go
+++ b/pkg/api/v1/atlasdeployment_types.go
@@ -617,6 +617,14 @@ func (c *AtlasDeployment) IsAdvancedDeployment() bool {
 	return c.Spec.AdvancedDeploymentSpec != nil
 }
 
+func (c *AtlasDeployment) GetReplicationSetID() string {
+	if len(c.Status.ReplicaSets) > 0 {
+		return c.Status.ReplicaSets[0].ID
+	}
+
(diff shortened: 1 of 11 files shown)


pkg/api/v1/atlasdeployment_types.go
```diff
@@ -617,6 +617,14 @@ func (c *AtlasDeployment) IsAdvancedDeployment() bool {
 	return c.Spec.AdvancedDeploymentSpec != nil
 }
 
+func (c *AtlasDeployment) GetReplicationSetID() string {
+	if len(c.Status.ReplicaSets) > 0 {
+		return c.Status.ReplicaSets[0].ID
```
josvazg, line 622: Getting always the first is always enough and correct?

josvazg (approved):
LGTM

pkg/api/v1/atlasdeployment_types.go
helderjs: Currently yes, it matches the Atlas behavior.
While the API seems like a cluster can have multiple replica-set, it is only an architectural design of how a cluster is configured, being a replica-set a group of mongod instances. The sharded architecture in another way can have each shard as replica-set but the API still represents it as a single replica-set

igor-karpukhin (approved):

PR #1179, opened 2023-10-17
CLOUDP-179084: Push daily builds also to Quay.io

josvazg:
Update releases to rebuild and push to all tags and to all repos with a [single action step](https://docs.docker.com/build/ci/github-actions/push-multi-registries/).

It also contains a **new docker build process that leverages native Go cross compilation** to reduce the **CI job time from over 30 minutes to about 5 minutes**.

✅  [Daily Build](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/6571312683) 

### All Submissions:


.github/workflows/rebuild-released-images.yaml
```diff
@@ -63,22 +64,13 @@ jobs:
...
-      - name: "Set up Docker Buildx"
-        uses: docker/setup-buildx-action@v3
-        with:
-          platforms: ${{ fromJSON(needs.read-versions.outputs.matrix).platforms }}
-      - name: Login to docker registry
-        uses: docker/login-action@v3
-        with:
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Upgrade build Makefile & Dockerfile
```
helderjs, line 70: Was this added to be able to test it?

josvazg: Not really, this is needed because of the new build code. As the rebuild uses the old released commit, it needs to patch the `Makefile` and add the new `fast.Dockerfile` to be able to build properly using the new way. Once the last version has moved to the new way we would be able to remove this.

igor-karpukhin (approved):
Great job! 👍

josvazg:
SBOM included

.github/workflows/rebuild-released-images.yaml
josvazg: Not really. This is required as the daily build executes in old code which does not have the newer Makefile and fast.Dockerfile that is used by this new build process.

PR #1181, opened 2023-10-18
CLOUDP-205248: Build once on release

josvazg:
✅  [Create Release (test)](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/6574565626/job/17859870298)
✅  [E2E tests](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/6574580645)
✅  [Integration tests](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/6575123825)

The build-and-push sub-action will:
- **Fail** if both push to docker and quay are unset.
- Set the main build **tag** to push to Docker OR Quay.
- If both flags are set it means it pushed to docker and also must push to quay. If both were unset it would have already failed. If docker was unset then it would have already push to quay. If quay is unset, no need to push to quay.
- If the verified version (`*-certified`) is set AND push to quay is also set, push verified version to quay.

### All Submissions:

### DIFF
--- a/.github/actions/build-push-image/action.yaml
+++ b/.github/actions/build-push-image/action.yaml
@@ -11,6 +11,10 @@ inputs:
   version:
     description: The version of the operator will be built
     required: true
+  certified_version:
+    description: The certified version of the operator will be built
+    default: ''
+    required: false
   repository:
     description: The name of repository to build image
     required: true
@@ -58,20 +62,10 @@ runs:
         ref: ${{github.event.pull_request.head.sha}}
         repository: ${{github.event.pull_request.head.repo.full_name}}
         submodules: true
-    - name: "Set up QEMU"
-      uses: docker/setup-qemu-action@v2
-      with:
-        platforms: ${{ inputs.platforms }}
     - name: "Set up Docker Buildx"
       uses: docker/setup-buildx-action@v2
       with:
         platforms: ${{ inputs.platforms }}
-    - name: Login to docker registry
-      if: ${{ inputs.push_to_docker == 'true' }}
-      uses: docker/login-action@v2
-      with:
-        username: ${{ inputs.docker_username }}
-        password: ${{ inputs.docker_password }}
     - name: Set up Go
       uses: actions/setup-go@v3
       with:
@@ -88,18 +82,25 @@ runs:
       shell: bash
       run: |
         go mod download
-    - name: Build and Push Operator to Docker Registry
+    - name: Build all platforms
+      shell: bash
+      run: make all-platforms
+    - name: Fail if send input flags are both unset
+      if: ${{ inputs.push_to_docker != 'true' && inputs.push_to_quay != 'true' }}
+      shell: bash
+      run: exit 1
+    - name: Compute main build tag
+      id: tag
+      shell: bash
+      run: |
+        main_tag=${{ inputs.push_to_docker == 'true' && format('{0}:{1}', inputs.repository, inputs.version) || format('quay.io/{0}:{1}', inputs.repository, inputs.version)}}
+        echo "main_tag=$main_tag" >> $GITHUB_OUTPUT
+    - name: Login to docker registry
       if: ${{ inputs.push_to_docker == 'true' }}
-      uses: docker/build-push-action@v3
+      uses: docker/login-action@v2
       with:
-        context: .
-        file: ${{ inputs.file }}
-        build-args: VERSION=${{ inputs.version }}
-        platforms: ${{ inputs.platforms }}
-        tags: ${{ inputs.repository }}:${{ inputs.version }}
-        cache-from: type=gha
-        cache-to: type=gha,mode=max
-        push: true
+        username: ${{ inputs.docker_username }}
+        password: ${{ inputs.docker_password }}
     - name: Login to quay.io registry
       if: ${{ inputs.push_to_quay == 'true' }}
       uses: docker/login-action@v2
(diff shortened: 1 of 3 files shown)


.github/actions/build-push-image/action.yaml
```diff
@@ -44,6 +48,11 @@ inputs:
     default: "false"
 runs:
   using: "composite"
+  env:
+    TAG1: "${{ inputs.repository }}:${{ inputs.version }}"
+    TAG2: "${{ inputs.push_to_quay && format('quay.io/{0}:{1}', inputs.repository, inputs.version) || '' }}"
+    TAG3: "${{ inputs.certified_version != '' && format('{0}:{1}', inputs.repository, inputs.certified_version) || '' }}"
+    TAG4: "${{ inputs.certified_version != '' && inputs.push_to_quay && format('quay.io/{0}:{1}', inputs.repository, inputs.certified_version) || '' }}"
```
igor-karpukhin, line 55: I think it missing the "-certified" ending for the certified image tag. E.g. "1.9.0-certified". So that it won't be overwritten by daily builds

josvazg: Are you sure?

This input is filled like this in the invoking workflow:
`certified_version: ${{ steps.tag.outputs.certified_version }}`

Anyway this code is going to disappear in the next update from here, but will be used similarly later. I will double check we have the expected '-certified' suffix in place.

igor-karpukhin: The release process to quay.io for the certified operator bundle should use `<version>-certified` tag format. See https://quay.io/repository/mongodb/mongodb-atlas-kubernetes-operator?tab=tags Otherwise, the daily build would rewrite the <version> and it can not be pulled later by the OLM, since for certification we use SHA256 of an image instead of a tag

josvazg: I am confused:

`.github/workflows/release-post-merge.yml`
Line 97: `certified_version: ${{ steps.tag.outputs.certified_version }}`
This is set at line 67:
`certified_version="${version}-certified"`

So the "<version>-certified" is propagated to the build-and-push sub action and used there.

What am I missing?

josvazg: Test sample:
https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/6572673633/job/17854215674

```shell
...
PUSH: docker push quay.io/mongodb/mongodb-atlas-kubernetes-operator-prerelease:1.9.1-rc0
...
PUSH: docker push mongodb/mongodb-atlas-kubernetes-operator-prerelease:1.9.1-rc0-certified
...
PUSH: docker push quay.io/mongodb/mongodb-atlas-kubernetes-operator-prerelease:1.9.1-rc0-certified
...
```

.github/actions/build-push-image/action.yaml
```diff
@@ -88,34 +82,65 @@ runs:
...
         build-args: VERSION=${{ inputs.version }}
         platforms: ${{ inputs.platforms }}
-        tags: quay.io/${{ inputs.repository }}:${{ inputs.version }}
         cache-from: type=gha
         cache-to: type=gha,mode=max
+        sbom: true
         push: true
+        tags: ${{ steps.tag.outputs.main_tag }}
+    - name: Push To Quay
+      if: ${{ inputs.push_to_quay == 'true' && inputs.push_to_docker == 'true' }}
```
helderjs, line 124: Why does push to Quay depend on push to docker?

.github/actions/build-push-image/action.yaml
```diff
@@ -88,34 +82,65 @@ runs:
...
+      shell: bash
+      run: |
+        docker pull ${{ steps.tag.outputs.main_tag }}
+        docker tag ${{ steps.tag.outputs.main_tag }} quay.io/${{ inputs.repository }}:${{ inputs.version }}
+        echo "PUSH: docker push quay.io/${{ inputs.repository }}:${{ inputs.version }}"
+        docker push quay.io/${{ inputs.repository }}:${{ inputs.version }}
+    - name: Push Certified to Docker Registry
```
helderjs, line 131: Certified image would never be pulled from docker, it doesn't make sense to actually push to it. I'd remove this step

.github/actions/build-push-image/action.yaml
```diff
@@ -88,34 +82,65 @@ runs:
...
         go mod download
-    - name: Build and Push Operator to Docker Registry
+    - name: Build all platforms
+      shell: bash
+      run: make all-platforms
+    - name: Fail if send input flags are both unset
+      if: ${{ inputs.push_to_docker != 'true' && inputs.push_to_quay != 'true' }}
+      shell: bash
+      run: exit 1
+    - name: Compute main build tag
```
helderjs, line 92: I don't get why you added this step if when pushing you're prefixing quay.io anyway

.github/workflows/release-post-merge.yml
```diff
@@ -12,14 +12,36 @@ on:
       version:
         description: "Release version (Be sure `Release-branch` is successful):"
         required: true
-
+      image_repo:
```
helderjs, line 15: Once approved, should we set the default to the one actually used when releasing?

.github/workflows/release-post-merge.yml
```diff
@@ -12,14 +12,36 @@ on:
...
+        type: choice
+        description: "Target image repository for built images"
+        default: mongodb/mongodb-atlas-kubernetes-operator-prerelease
+        required: true
+        options:
+        - mongodb/mongodb-atlas-kubernetes-operator-prerelease
+        - mongodb/mongodb-atlas-kubernetes-operator
+      release_helm:
```
helderjs, line 23: Once approved, should we cleanup those control flags? or at least set "prod" default values

helderjs (changes requested):
It would be good to have the CI running completely once other parts might be using the changed action, e2e test for sure use it to build an image.

josvazg:
> It would be good to have the CI running completely once other parts might be using the changed action, e2e test for sure use it to build an image.

Will kick off a e2e test suite manually

.github/actions/build-push-image/action.yaml
josvazg: We have both input flags to push to docker AND / OR quay. In top of that we know that build and push must push to at lease some repo, but depending on the inputs, we do not know which one it will be. The table of truth looks like this:

- Both flags false => **Fail the job!**
- Push to docker / NO quay => build and push will push to docker and this must be skipped 
- Push to quay / NO docker => build and push will push to quay, so this must be skipped also
- Both Push docker & quay => build and push will push to docker, and this must push to quay now

.github/actions/build-push-image/action.yaml
josvazg: will fix

.github/actions/build-push-image/action.yaml
josvazg: No, this tag job decides if we are pushing to docker OR quay in the build and push, depending on the input flags. See the table of truth in the other comment.

.github/workflows/release-post-merge.yml
josvazg: These defaults only apply for inputs, for prod it will fallback to fixed values, see env section.

.github/workflows/release-post-merge.yml
josvazg: See env section on how default values get computed.

.github/workflows/release-post-merge.yml
```diff
@@ -12,14 +12,36 @@ on:
...
   create-release:
     name: Create Release
     if: ${{ (github.event.pull_request.merged == true && startsWith(github.head_ref, 'release/')) || github.event.inputs.version != '' }}
     runs-on: ubuntu-latest
     env:
-      IMAGE_REPOSITORY: mongodb/mongodb-atlas-kubernetes-operator
+      IMAGE_REPOSITORY: ${{ github.event.inputs.image_repo || 'mongodb/mongodb-atlas-kubernetes-operator' }}
+      RELEASE_HELM: ${{ github.event.inputs.release_helm || 'true' }}
+      CERTIFY: ${{ github.event.inputs.certify || 'true' }}
+      RELEASE_TO_GITHUB: ${{ github.event.inputs.release_to_github || 'true' }}
```
josvazg, line 44: This will fallback to the latest value on the expression for prod calls.

josvazg:
> > It would be good to have the CI running completely once other parts might be using the changed action, e2e test for sure use it to build an image.
> 
> Will kick off a e2e test suite manually

E2e test linked on top

.github/actions/build-push-image/action.yaml
josvazg: Done

helderjs (approved):

PR #1185, opened 2023-10-23
CLOUDP-202467: Cleanup unused CI bits

josvazg:
Removing unused workflows and custom actions surfaced when inventorying our credentials usage from the CI.

### All Submissions:


.github/workflows/release-openshift.yaml
helderjs, line 1: This workflow only depends on a valid GH token to do its job and save us from all manual jobs of release Openshit bundles. I'd like to keep it until we figure if we can have such configuration

josvazg: got it, will restore this one.

josvazg: Done

helderjs (approved):

PR #1187, opened 2023-10-23
task: replace github.com/pborman/uuid with github.com/google/uuid

gssbzn:
### All Submissions:
I'm trying to reduce the amount of 3rd party libs the operator adds to the CLI, this as an easy one, unifying all uuid libs to the google one which is the newer one


josvazg (approved):
LGTM

helderjs (approved):

PR #1203, opened 2023-10-30
CLOUDP-205479: Force secretRef fields

roothorp:
### All Submissions:

### DIFF
--- a/.github/actions/gen-install-scripts/entrypoint.sh
+++ b/.github/actions/gen-install-scripts/entrypoint.sh
@@ -14,7 +14,7 @@ mkdir -p "${crds_dir}"
 mkdir -p "${openshift}"
 
 # Generate configuration and save it to `all-in-one`
-controller-gen crd:crdVersions=v1 rbac:roleName=manager-role webhook paths="./pkg/api/..." output:crd:artifacts:config=config/crd/bases
+controller-gen crd:crdVersions=v1,ignoreUnexportedFields=true rbac:roleName=manager-role webhook paths="./pkg/api/..." output:crd:artifacts:config=config/crd/bases
 cd config/manager && kustomize edit set image controller="${INPUT_IMAGE_URL}"
 cd -
 ./scripts/split_roles_yaml.sh
--- a/pkg/api/v1/alert_configurations.go
+++ b/pkg/api/v1/alert_configurations.go
@@ -47,8 +47,7 @@ func (in *AlertConfiguration) ToAtlas() (*mongodbatlas.AlertConfiguration, error
 	}
 
 	for _, n := range in.Notifications {
-		notification := &mongodbatlas.Notification{}
-		err := compat.JSONCopy(&notification, n)
+		notification, err := n.ToAtlas()
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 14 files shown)


.github/actions/gen-install-scripts/entrypoint.sh
```diff
@@ -14,7 +14,7 @@ mkdir -p "${crds_dir}"
 mkdir -p "${openshift}"
 
 # Generate configuration and save it to `all-in-one`
-controller-gen crd:crdVersions=v1 rbac:roleName=manager-role webhook paths="./pkg/api/..." output:crd:artifacts:config=config/crd/bases
+controller-gen crd:crdVersions=v1,ignoreUnexportedFields=true rbac:roleName=manager-role webhook paths="./pkg/api/..." output:crd:artifacts:config=config/crd/bases
```
josvazg, line 17: Should we document or comment somewhere we need this in order to allow for private struct types to be in the Go code but not show up in the CRD? It might be self-explanatory though

josvazg:
Seems a unit test might be broken?

pkg/api/v1/common/common.go
```diff
@@ -51,6 +52,13 @@ func (rn *ResourceRefNamespaced) GetObject(parentNamespace string) *client.Objec
 	return &key
 }
 
+func (rn *ResourceRefNamespaced) GetNamespacedName() types.NamespacedName {
```
helderjs, line 55: Why not use `ResourceRefNamespaced.GetObject()` instead?

pkg/api/v1/atlasproject_types.go
```diff
@@ -125,28 +125,10 @@ const hiddenField = "*** redacted ***"
...
-	// cleanup encryption at EncryptionAtRest
-	if printable.EncryptionAtRest != nil {
-		printable.EncryptionAtRest.AwsKms.AccessKeyID = hiddenField
-		printable.EncryptionAtRest.AwsKms.CustomerMasterKeyID = hiddenField
-		printable.EncryptionAtRest.AwsKms.SecretAccessKey = hiddenField
-		printable.EncryptionAtRest.AwsKms.RoleID = hiddenField
-		printable.EncryptionAtRest.AzureKeyVault.Secret = hiddenField
-		printable.EncryptionAtRest.GoogleCloudKms.ServiceAccountKey = hiddenField
-	}
-
```
igor-karpukhin, line 137: I know you made some of the fields unexported, but aren't they still going to be printed?

helderjs (approved):

pkg/api/v1/alert_configurations.go
```diff
@@ -137,14 +136,14 @@ func (in *Threshold) ToAtlas() (*mongodbatlas.Threshold, error) {
...
+	apiToken string
+	// Secret containing a Slack API token or Bot token. Populated for the SLACK notifications type. If the token later becomes invalid, Atlas sends an email to the project owner and eventually removes the token.
 	// +optional
 	APITokenRef common.ResourceRefNamespaced `json:"apiTokenRef,omitempty"`
 	// Slack channel name. Populated for the SLACK notifications type.
-	ChannelName string `json:"channelName,omitempty"`
-	// Datadog API Key. Found in the Datadog dashboard. Populated for the DATADOG notifications type.
-	DatadogAPIKey string `json:"datadogApiKey,omitempty"`
+	ChannelName   string `json:"channelName,omitempty"`
+	datadogAPIKey string
```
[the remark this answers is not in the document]
roothorp: I've made a ticket to keep track of this, definitely something I'd like to clean up a bit in future!

pkg/api/v1/atlasproject_types.go
roothorp: I'm fairly sure that [Zap does not print unexported fields](https://github.com/uber-go/zap/issues/1097#issuecomment-1124393884) - I think because it's built around JSON marshalling. Just to be safe, I've set the printable fields to hiddenField as it was before.

pkg/api/v1/common/common.go
roothorp: Done - but seems weird that GetObject takes a namespace param when the object it's operating on already has one?

igor-karpukhin (approved):
Good job @roothorp! 👍

pkg/api/v1/alert_configurations.go
```diff
@@ -234,6 +259,19 @@ func (in *Notification) IsEqual(notification mongodbatlas.Notification) bool {
 	return true
 }
 
+func (in *Notification) ToAtlas() (*mongodbatlas.Notification, error) {
+	result := &mongodbatlas.Notification{}
+	err := compat.JSONCopy(result, in)
```
s-urbaniak, line 264: as a follow-up fix:
```
if err != nil {
  return nil, err
```

pkg/api/v1/alert_configurations.go
```diff
@@ -234,6 +259,19 @@ func (in *Notification) IsEqual(notification mongodbatlas.Notification) bool {
...
+	result.APIToken = in.apiToken
+	result.DatadogAPIKey = in.datadogAPIKey
+	result.FlowdockAPIToken = in.flowdockAPIToken
+	result.OpsGenieAPIKey = in.opsGenieAPIKey
+	result.ServiceKey = in.serviceKey
+	result.VictorOpsAPIKey = in.victorOpsAPIKey
+	result.VictorOpsRoutingKey = in.victorOpsRoutingKey
+	return result, err
```
s-urbaniak, line 272: as a follow-up fix:
```
return nil, err
```

PR #1210, opened 2023-11-02
CLOUDP-208614: Document PR fixes for release

josvazg:
### All Submissions:

### DIFF
--- a/docs/dev/release.md
+++ b/docs/dev/release.md
@@ -167,9 +167,9 @@ from the `mongodb-fork` repository to https://github.com/redhat-openshift-ecosys
 Before posting the PR there are manual changes you need to make:
 
 1. Ensure to add the `quay.io/` prefix in all Operator image references.
-1. Add missing a `com.redhat.openshift.versions: "v4.8"` line at the end of `metadata/annotations.yaml`.
+1. Add a missing `com.redhat.openshift.versions: "v4.8"` line at the end of `metadata/annotations.yaml`.
 1. Ensure all image references, including `containerImage`, do NOT use the version *tag*. They **should only use the SHA of the AMD image**, NEVER the multi arch SHA.
-1. Add missing `spec.relatedImages` section in certified pinning all the images per architecture.
+1. Add the missing`spec.relatedImages` section in `manifests/mongodb-atlas-kubernetes.clusterserviceversion.yaml`, to pin all the images per architecture.
 
 For some reason, the certified OpenShift metadata does not use the multi arch image reference at all, and only understand direct architecture image references.
 
--- a/docs/dev/release.md
+++ b/docs/dev/release.md
@@ -98,14 +98,20 @@ pointing to the directory where `operatorhub-operator` repository was cloned in 
 
 For this PR the sources are copied from the `community-operators` folder instead of the one where the `mongodb-atlas-kubernetes` resides.
 
-Invoke with <version> like `1.0.0`:
+Invoke with <version> like `1.0.0` (never use the `v` prefix here, just the plain SEMVER version `x.y.z`):
+
 ```
 ./scripts/release-redhat.sh <version>
 ```
 
+Before posting the PR there is a manual change you need to make:
+
+* Ensure to add the `quay.io/` prefix in all Operator image references.
+
+You can see an [example fixed PR here on Community Operators for version 1.9.1](https://github.com/k8s-operatorhub/community-operators/pull/3457).
+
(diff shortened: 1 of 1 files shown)


docs/dev/release.md
```diff
@@ -98,14 +98,20 @@ pointing to the directory where `operatorhub-operator` repository was cloned in
 
 For this PR the sources are copied from the `community-operators` folder instead of the one where the `mongodb-atlas-kubernetes` resides.
 
-Invoke with <version> like `1.0.0`:
+Invoke with <version> like `1.0.0` (never use the `v` prefix here, just the plain SEMVER version `x.y.z`):
```
s-urbaniak, line 101: curiosity question: why no `v` prefix? (formatting?)

docs/dev/release.md
```diff
@@ -154,8 +164,18 @@ If script successfully finishes, you should be able to see new tag (e.g. 1.2.0)
 Then go the GitHub and create a PR
 from the `mongodb-fork` repository to https://github.com/redhat-openshift-ecosystem/certified-operators (`origin`).
 
+Before posting the PR there are manual changes you need to make:
+
+1. Ensure to add the `quay.io/` prefix in all Operator image references.
+1. Add missing a `com.redhat.openshift.versions: "v4.8"` line at the end of `metadata/annotations.yaml`.
```
s-urbaniak, line 170:
```suggestion
1. Add a missing `com.redhat.openshift.versions: "v4.8"` line at the end of `metadata/annotations.yaml`.
```

s-urbaniak: curiosity question: why OpenShift 4.8?

docs/dev/release.md
```diff
@@ -154,8 +164,18 @@ If script successfully finishes, you should be able to see new tag (e.g. 1.2.0)
...
+1. Ensure all image references, including `containerImage`, do NOT use the version *tag*. They **should only use the SHA of the AMD image**, NEVER the multi arch SHA.
```
s-urbaniak, line 171: curiosity question: what is the reasoning behind only hashing the AMD image?

docs/dev/release.md
```diff
@@ -154,8 +164,18 @@ If script successfully finishes, you should be able to see new tag (e.g. 1.2.0)
...
+1. Add missing `spec.relatedImages` section in certified pinning all the images per architecture.
```
s-urbaniak, line 172: What is "certified pinning" and what content shoidl `relatedImages` have?

docs/dev/release.md
josvazg: I made that mistake myself in a PR for version 1.9.1 and that breaks the diff. The v prefix is used at some places (maybe git tags/branches) but not in others (like docker tags) so the code will put it for you where applicable, but if you added it, you break the process as it is designed now.

docs/dev/release.md
josvazg: Not sure on why that one version in particular, maybe @igor-karpukhin can shed some light here

docs/dev/release.md
josvazg: We do not only hash the AMD image, but this certified PR only knows about actual direct images, not multi arch references. So when an image needs to be specified we chose the AMD one by default.

The per arch images are specified in the `spec.relatedImages` as I will explain later.

docs/dev/release.md
josvazg: Yeah that wording is not very clear I will revisit it...

This is an example of ´relatedImages´:

´´´yaml
spec:
  relatedImages:
    - name: mongodb-atlas-kubernetes-operator-arm64
      image: quay.io/mongodb/mongodb-atlas-kubernetes-operator@sha256:f13541f1d63c9f9b959cad82823fcc9aeb9ac50f94339f0ce606affbccf1c4d0
    - name: mongodb-atlas-kubernetes-operator-amd64
      image: quay.io/mongodb/mongodb-atlas-kubernetes-operator@sha256:ec4cb2fc9e32c88a7372be51971a5e24a1c87251f68269423e553e2febf9d4eb
´´´

Basically certified does not seem to understand multi-arch images, so you need to specify each of the hashes of each of the architectures you support here. Elsewhere, when a single image ref is required, we tend to use the AMD one by default.

josvazg, line 172:
```suggestion
1. Add the missing`spec.relatedImages` section in `manifests/mongodb-atlas-kubernetes.clusterserviceversion.yaml`, to pin all the images per architecture.
```

helderjs (approved):

PR #1212, opened 2023-11-03
Explicitly check licenses in CI

josvazg:
Convenience improvement to avoid having to run the license check on every `make unit-test` by moving that to an explicit CI check. The check also makes sure we do not forget to update the `licenses.csv` file as it will fail if it dectects it is stale.

### All Submissions:


josvazg:
> @josvazg Would it be better to run this only once during the release process? And not on every `make unit-test` run

Wait, something is wrong. make unit-test is now always running the CSV, even if code has not changed. This is not expected.

josvazg:
Fixed, now the CSV creation it runs on unit tests ONLY when `go.mod` changes.

That should be a good and hopefully reasonable balance between ensuring that the CSV gets updated before the PR is submitted and not being too intrusive when running unit tests locally.

Makefile
```diff
@@ -96,25 +96,28 @@ help: ## Show this help screen
...
-	$(GO_LICENSES) csv --include_tests $(BASE_GO_PACKAGE)/... > $@
+	GOOS=linux GOARCH=amd64 $(GO_LICENSES) csv --include_tests $(BASE_GO_PACKAGE)/... > $@
 
 .PHONY: check-licenses
 check-licenses: licenses.csv  ## Check licenses
 	@echo "Checking licenses not to be: $(DISALLOWED_LICENSES)"
 	@echo "============================================"
-	$(GO_LICENSES) check --include_tests $(BASE_GO_PACKAGE)/... \
+	GOOS=linux GOARCH=amd64 $(GO_LICENSES) check --include_tests $(BASE_GO_PACKAGE)/... \
 	--disallowed_types $(DISALLOWED_LICENSES)
```
slaskawi, line 115: The `DISALLOWED_LICENSES` seems not defined. Is that intentional?

Makefile
```diff
@@ -96,25 +96,28 @@ help: ## Show this help screen
...
 	@echo "--------------------"
 	@echo "Licenses check: PASS"
 
 .PHONY: unit-test
-unit-test: check-licenses
+unit-test: licenses.csv
```
slaskawi, line 120: Would that pre-requirement be always fulfilled? The `licenses.csv` seems to be committed to the repo.

igor-karpukhin: I think it's better to run this check on the CI and if the licence file is different from the one in the PR we mark this check as "failed" and block the PR from being merged

josvazg: Thanks @igor-karpukhin , that makes sense I will fix it and make the check in the CI

@slaskawi I will use the CI, as Igor suggests, to ensure we do not forget to run make licenses.csv and not run on unit tests.

Makefile
josvazg: will double check, thanks for the heads up

josvazg: It is defined in line 84

.github/workflows/test-forked.yml
```diff
@@ -21,6 +21,11 @@ jobs:
     with:
       forked: true
 
+  check-licenses:
+    uses: ./.github/workflows/check-licenses.yml
+    with:
+      forked: false
```
helderjs, line 27: I think this should be true
```suggestion
      forked: true
```

Makefile
```diff
@@ -96,25 +96,35 @@ help: ## Show this help screen
...
+	@mkdir -p $(TIMESTAMPS_DIR) && touch $@
+
+go-licenses: $(TIMESTAMPS_DIR)/go-licenses-installed
 
-licenses.csv: $(GO_SOURCES) go.mod ## Track licenses in a CSV file
+licenses.csv: $(TIMESTAMPS_DIR)/go-licenses-installed go.mod ## Track licenses in a CSV file
 	@echo "Tracking licenses into file $@"
 	@echo "========================================"
-	$(GO_LICENSES) csv --include_tests $(BASE_GO_PACKAGE)/... > $@
+	GOOS=linux GOARCH=amd64 $(GO_LICENSES) csv --include_tests $(BASE_GO_PACKAGE)/... > $@
```
helderjs, line 108: out of curiosity, why do you need to explicitly set OS and ARCH?

helderjs (changes requested):

Makefile
josvazg: We want to create the licenses for the same ARCH and OS always, so they do not differ if they were computed on your Mac or in the CI.

Unfortunately, this does not really fix all the cases. Right now I run `make licenses.csv` at my computer and the CI finds differences when it repeats that check.

Makefile
josvazg: Unfortunately, we cannot rely on the CI for reminding us to update the licenses .CSV because it reports differences with the same command just run from your Mac.

Not sure how to proceed now.

josvazg: The check is now on the file dates, that should be enough as a reminder.
I am going to add a comment to explain why we are not comparing generated files instead.

slaskawi: I'm OK with this approach.

Makefile
slaskawi: Perfect! Thanks!

slaskawi (approved):

.github/workflows/test-forked.yml
josvazg: agreed
```suggestion
      forked: true
```

helderjs (approved):

igor-karpukhin (approved):

PR #1221, opened 2023-11-08
CLOUDP-209691: Added OIDC and AWS IAM auth fields for AtlasDatabaseUser

igor-karpukhin:
### All Submissions:

Added OIDC and AWS IAM auth fields for AtlasDatabaseUser

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -44,6 +44,8 @@ import (
 
 	mdbv1 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/featureflags"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlasdatabaseuser"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlasdatafederation"
@@ -167,15 +169,16 @@ func main() {
 	}
 
 	if err = (&atlasdatabaseuser.AtlasDatabaseUserReconciler{
-		ResourceWatcher:             watch.NewResourceWatcher(),
-		Client:                      mgr.GetClient(),
-		Log:                         logger.Named("controllers").Named("AtlasDatabaseUser").Sugar(),
-		Scheme:                      mgr.GetScheme(),
-		EventRecorder:               mgr.GetEventRecorderFor("AtlasDatabaseUser"),
-		GlobalPredicates:            globalPredicates,
-		AtlasProvider:               atlasProvider,
-		ObjectDeletionProtection:    config.ObjectDeletionProtection,
-		SubObjectDeletionProtection: config.SubObjectDeletionProtection,
+		ResourceWatcher:               watch.NewResourceWatcher(),
+		Client:                        mgr.GetClient(),
+		Log:                           logger.Named("controllers").Named("AtlasDatabaseUser").Sugar(),
+		Scheme:                        mgr.GetScheme(),
+		EventRecorder:                 mgr.GetEventRecorderFor("AtlasDatabaseUser"),
+		AtlasProvider:                 atlasProvider,
+		GlobalPredicates:              globalPredicates,
+		ObjectDeletionProtection:      config.ObjectDeletionProtection,
+		SubObjectDeletionProtection:   config.SubObjectDeletionProtection,
+		FeaturePreviewOIDCAuthEnabled: config.FeatureFlags.IsFeaturePresent(featureflags.FeatureOIDC),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "AtlasDatabaseUser")
 		os.Exit(1)
@@ -241,6 +244,7 @@ type Config struct {
 	LogEncoder                  string
 	ObjectDeletionProtection    bool
 	SubObjectDeletionProtection bool
+	FeatureFlags                *featureflags.FeatureFlags
 }
 
 // ParseConfiguration fills the 'OperatorConfig' from the flags passed to the program
@@ -286,6 +290,7 @@ func parseConfiguration() Config {
 
 	configureDeletionProtection(&config)
 
+	config.FeatureFlags = featureflags.NewFeatureFlags(os.Environ)
 	return config
 }
 
--- a/internal/featureflags/featureflag.go
+++ b/internal/featureflags/featureflag.go
@@ -1,0 +1,50 @@
+package featureflags
+
+import (
+	"strings"
+)
+
+const (
+	featurePrefix    = "FEATURE_"
+	featureSeparator = "="
+
+	//nolint:stylecheck
+	FeatureOIDC = "FEATURE_PREVIEW_OIDC_DB_ACCESS"
+)
+
+type FeatureFlags struct {
+	features map[string]string
+}
+
+type EnvLister func() []string
+
+// NewFeatureFlags creates a new instance of FeatureFlags and reads feature flags from the ENV
+func NewFeatureFlags(envVarsLister EnvLister) *FeatureFlags {
+	instance := &FeatureFlags{}
+	envs := envVarsLister()
+	result := map[string]string{}
+	for _, e := range envs {
+		if strings.HasPrefix(e, featurePrefix) {
+			keyVal := strings.SplitN(e, featureSeparator, 2)
+			if len(keyVal) == 2 {
+				result[keyVal[0]] = keyVal[1]
+			}
+			result[e] = keyVal[0]
+		}
+	}
+	instance.features = result
+	return instance
+}
+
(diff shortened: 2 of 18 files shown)


pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -76,9 +76,32 @@ type AtlasDatabaseUserSpec struct {
 	// PasswordSecret is a reference to the Secret keeping the user password.
 	PasswordSecret *common.ResourceRef `json:"passwordSecretRef,omitempty"`
 
-	// Username is a username for authenticating to MongoDB.
+	// Username is a username for authenticating to MongoDB
+	// Human-readable label that represents the user that authenticates to MongoDB. The format of this label depends on the method of authentication:
+	// In case of AWS IAM: the value should be AWS ARN for the IAM User/Role;
+	// In case of OIDC: the value should be the Identity Provider ID;
+	// In case of Plain text auth: the value can be anything
```
josvazg, line 83: Is this actually true? I do not believe usernames allow for any arbitrary characters, they are usually letters and numbers, with some extra charts like - or _ or / allowed, but for instance, without spaces. Should we point to mongodb username format restrictions here?

igor-karpukhin: Looks like the only constraint there is the length: https://github.com/10gen/mms/blob/44ef23b3e8125a5d295be60120512d194c515474/server/src/main/com/xgen/svc/mms/api/view/atlas/ApiAtlasDatabaseUserView.java#L65-L72

igor-karpukhin:
Tests will be added today/tomorrow

pkg/api/v1/atlasdatabaseuser_types.go
s-urbaniak: does it make sense to xref that location here as well?

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -76,9 +76,32 @@ type AtlasDatabaseUserSpec struct {
...
+	// +kubebuilder:validation:MaxLength:=1024
 	Username string `json:"username"`
 
+	// Human-readable label that indicates whether the new database Username
+	// with OIDC federated authentication.
+	// To create a federated authentication user, specify the value
+	//of IDP_GROUP for this field
```
s-urbaniak, line 90: nit: missing whitespace

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -76,9 +76,32 @@ type AtlasDatabaseUserSpec struct {
...
+	// +kubebuilder:default:=NONE
+	// +kubebuilder:validation:Enum:=NONE;IDP_GROUP
+	// +optional
```
s-urbaniak, line 93: What is the expected behavior if the field is left empty? (add to godoc)

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -76,9 +76,32 @@ type AtlasDatabaseUserSpec struct {
...
+	OIDCAuthType string `json:"oidcAuthType,omitempty"`
+
+	// Human-readable label that indicates whether the new database
+	// user authenticates with the Amazon Web Services (AWS)
+	// Identity and Access Management (IAM) credentials associated with
+	// the user or the user's role
+	// +kubebuilder:default:=NONE
+	// +kubebuilder:validation:Enum:=NONE;USER;ROLE
+	// +optional
```
s-urbaniak, line 102: What is the expected behavior if the field is left empty? (add to godoc)

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -194,6 +204,32 @@ func (r *AtlasDatabaseUserReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	return nil
+}
+
+// TODO: Remove after the OIDC feature becomes stable
+func handleOIDCPreview(featureFlags *featureflags.FeatureFlags, dbuser *mdbv1.AtlasDatabaseUser) error {
+	if dbuser == nil {
+		return nil
+	}
+
+	if !featureFlags.IsFeaturePresent(featureflags.FeatureOIDC) && dbuser.Spec.OIDCAuthType != "" {
```
s-urbaniak, line 226: suggestion: let's handle feature flags centrally, more near to `main` and just dependency-inject a `oidcEnabled bool` field into the reconciler struct to keep logic in one place.

s-urbaniak: (this would reduce cross-package dependencies as well)

s-urbaniak:
@igor-karpukhin I see no explicit logic for handling these new fields on the Atlas side of things. Presumable this is because these fields are transparently copied via `JSONCopy` to Atlas fields?

internal/featureflags/featureflag.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	featurePrefix    = "FEATURE_"
+	featureSeparator = "="
+	//nolint:stylecheck
+	FeatureOIDC = "FEATURE_PREVIEW_OIDC_DB_ACCESS"
```
josvazg, line 12: Is this the only known feature flag so far?
If so. How do we know we will need this feature flags going forward for more values?

josvazg: Actually this could be what we need if we go for a more controlled process for releasing features. Instead of a commit releasing a new feature or behaviour directly, a flag might control whether some of those are active and maybe even depend on what version value it is being built, the default feature flag value is different. One single codebase can release different operator versions without branching.

josvazg:
LGTM but waiting for other comments to be revisited and the tests to pass

igor-karpukhin:
> @igor-karpukhin I see no explicit logic for handling these new fields on the Atlas side of things. Presumable this is because these fields are transparently copied via `JSONCopy` to Atlas fields?

That's right. These values are just sent as is, without any processing.

internal/featureflags/featureflag.go
igor-karpukhin: Yes, this is the only feature flag so far. I implemented then as env variables as it is easier to configure for the k8s `Deployment` and/or `Pod` resource. We can also add flag processing if needed (e.g. `--features=featureA,featureB`.

pkg/api/v1/atlasdatabaseuser_types.go
igor-karpukhin: Thanks, fixed!

pkg/api/v1/atlasdatabaseuser_types.go
igor-karpukhin: It should default to `NONE`. Anyway, even if it is missing, this parameter is optional to for the API call

pkg/api/v1/atlasdatabaseuser_types.go
igor-karpukhin: Same as above, It should default to `NONE`. Even if missing, this parameter is optional to for the API call

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
igor-karpukhin: Good point! Fixed

pkg/api/v1/atlasdatabaseuser_types.go
s-urbaniak: makes total sense, thanks for the clarification!

s-urbaniak (approved):
lgtm modulo failing tests, great work! 🎉

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -43,18 +44,24 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/workflow"
 )
 
+//nolint:stylecheck
+const FeatureOIDC = "FEATURE_PREVIEW_OIDC_DB_ACCESS"
```
s-urbaniak, line 48: non-blocking thought: I wonder if we rather want to gather all features in a central `features.go` file rather than sprinkling them in controller logic implementation files 🤔

igor-karpukhin: Fixed :)

cmd/manager/main.go
```diff
@@ -62,6 +64,7 @@ const (
 	subobjectDeletionProtectionEnvVar  = "SUBOBJECT_DELETION_PROTECTION"
 	objectDeletionProtectionDefault    = true
 	subobjectDeletionProtectionDefault = true
+	featurePreviewFlagOIDC             = "FEATURE_PREVIEW_OIDC_DB_ACCESS"
```
s-urbaniak, line 67: i believe this is unused?

igor-karpukhin: Yes, thanks! Removed

s-urbaniak (approved):
LGTM, love it, thank you for the patience! 🙏

josvazg:
Seems it needs a rebase

josvazg (approved):
LGTM, just needs a rebase it seems

josvazg (approved):
Still LGTM 👍

PR #1222, opened 2023-11-08
CLOUDP-210137: Fix major upgrade test & release

josvazg:
The new file `major-version` marks our intent to release a  new version of such major version. This specially important when the previous version was on an older major version. Using this explicit versioning we can:
- Skip tests that cannot work across major versions, like `helm update`.
- Avoid us from releasing by mistake an old major version using code intended for the newer major version.

✅ [e2e test helm update](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/6802231764/job/18516743993?pr=1222)

### All Submissions:


major-version
```diff
@@ -0,0 +1,2 @@
+2
+
```
s-urbaniak, line 2: superfluous newline

josvazg:
```suggestion
```

igor-karpukhin (approved):

helderjs:
While I don't see any problem merging this, I think this requires further discussion on how to approach `new releases` vs `a master branch ahead of it`.
For example, the 1.9.1 release was done from a different branch that never returned to the master. With that, the release of 2.0 was showing changes that were already released 1.9.1 and generating wrong data on the Openshift CSV

josvazg:
> While I don't see any problem merging this, I think this requires further discussion on how to approach `new releases` vs `a master branch ahead of it`. For example, the 1.9.1 release was done from a different branch that never returned to the master. With that, the release of 2.0 was showing changes that were already released 1.9.1 and generating wrong data on the Openshift CSV

I agree. We should not have to rely on a branch to re-release an older patch version. But I do not have a good solution yet.

This is a fail safe we can use until we get a better approach IMHO.

helderjs (approved):

PR #1237, opened 2023-11-14
Remove go.work.sum as well

josvazg:
We have recently [removed `go.work`](https://github.com/mongodb/mongodb-atlas-kubernetes/commit/f21d7c2c12cd9e78720bf59d82407dae2695d627), we should have also removed `go.work.sum` with it. Here it is the fix.

### All Submissions:


igor-karpukhin (approved):

roothorp:
Could we also add these files to .gitignore?

josvazg:
> Could we also add these files to .gitignore?

Why? do you want the option to use them locally in an untracked way?

PR #1249, opened 2023-11-21
Upgrade log/zapr from 1.2.4 to 1.3.0 manually

josvazg:
Replaces #1246 

The linter was acting up until I upgraded it.

I also noticed some complain in the Makefile that got fixed this a simple change: `--field-separator=" "` => `-F' '`

### All Submissions:


.github/workflows/lint.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
         uses: golangci/golangci-lint-action@v3.7.0
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
```
roothorp, line 28: Is this the case? We seem to be specifying the patch version anyway.

roothorp (approved):
LGTM, wondering about the comment in the lint workflow though.

.github/workflows/lint.yaml
josvazg: You are right. I had not noticed that. I can try and see if 1.54 would work, and we could leave that.

Security wise we should go the other way around, not just lock to the version, but to the sha of the action.

.github/workflows/lint.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
...
-          version: v1.53.3
+          version: v1.54.2
```
josvazg, line 29:
```suggestion
          version: v1.54
```

.github/workflows/lint.yaml
josvazg: seems to work just fine

helderjs (approved):

PR #1254, opened 2023-11-22
CLOUDP-212063: Fix k8s project delete and update race conditions

josvazg:
This fix includes several race condition fixes on Kubernetes object deletion and update.

It also includes a few project renames on the Kubernetes side that seem to fix some flakiness. I believe some tests were interfering with each other on the same project test set.

### All Submissions:

### DIFF
--- a/test/int/project_test.go
+++ b/test/int/project_test.go
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"sync"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/version"
 
 	"time"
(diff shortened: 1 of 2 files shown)


helderjs (approved):

test/int/project_test.go
```diff
@@ -388,12 +389,13 @@ var _ = Describe("AtlasProject", Label("int", "AtlasProject"), func() {
...
-				Expect(k8sClient.Get(context.Background(), kube.ObjectKeyFromObject(secondProject), secondProject)).To(Succeed())
-				secondProject.Spec.AlertConfigurationSyncEnabled = true
-				Expect(k8sClient.Update(context.Background(), secondProject)).To(Succeed())
+				Eventually(updateK8sObject[*mdbv1.AtlasProject](createdProject, func(createdProject *mdbv1.AtlasProject) {
+					createdProject.Spec.AlertConfigurationSyncEnabled = true
+				})).WithPolling(interval).WithTimeout(20 * time.Second).Should(BeTrue())
+
+				Eventually(updateK8sObject[*mdbv1.AtlasProject](secondProject, func(secondProject *mdbv1.AtlasProject) {
+					secondProject.Spec.AlertConfigurationSyncEnabled = true
+				})).WithPolling(interval).WithTimeout(20 * time.Second).Should(BeTrue())
```
igor-karpukhin, line 398: @josvazg  Looks good in general, but is there a reason to extract the logic into a function that is only called once? Consider moving the body of the function here using `Gomega`. That is way simpler than adding a generic function that takes another function

igor-karpukhin (approved):

PR #1258, opened 2023-11-23
CLOUDP-212564: Kick Helm Chart release using GitHub App

josvazg:
Use New GH App AKO Releaser to kick off the helm chart release from this repo.

Created a new `makejwt` tool to create a JWT token from the GitHub App RSA key.

Overall process:
1. Build & test the Go tool `makejwt`.
1. Use the new `scripts/release-helm.sh` script to kick of the Helm charts release.

The `scripts/release-helm.sh` needs to do some curl dance to go from the RSA key to a successful REST API call:
1. Run the `makejwt` tool to obtain the **JWT** from the App RSA key.
1. Use the **JWT** to query the **Installation Id** of the helm-charts repo via REST.
1. Use the **JWT** and the **Installation Id** to query **access token** the via REST.
1. Use the **access token** and Version to kick off the helm release via REST.

Test:

✅  [Test to kick Helm Chart release using AKO release GitHub App](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/7086068590/job/19283574246)
✅ [Helm Chart repo release branch job was started](https://github.com/mongodb/helm-charts/actions/runs/7086073058/job/19283588349)

### All Submissions:

### DIFF
--- a/.github/workflows/release-post-merge.yml
+++ b/.github/workflows/release-post-merge.yml
@@ -2,7 +2,7 @@
 # Trigger release branch should be merge into main
 # TODO add e2e/smoke test for autogen configuration
 
-name: Create Release.
+name: Create Release
 
 on:
   pull_request:
@@ -34,6 +34,7 @@ on:
         required: true
 jobs:
   create-release:
+    environment: release
     name: Create Release
     if: ${{ (github.event.pull_request.merged == true && startsWith(github.head_ref, 'release/')) || github.event.inputs.version != '' }}
     runs-on: ubuntu-latest
@@ -66,22 +67,21 @@ jobs:
           echo "version=$version" >> $GITHUB_OUTPUT
           echo "tag=$tag" >> $GITHUB_OUTPUT
           echo "certified_version=$certified_version" >> $GITHUB_OUTPUT
-      - name: Trigger helm post release workflow
-        if: ${{ env.RELEASE_HELM == 'true' }}
-        # Please provide a token with write access to the repository into secrets.HELM_REPO_TOKEN
-        run: |
-          curl \
-            -X POST \
-            -H "Accept: application/vnd.github+json" \
-            -H "Authorization: Bearer ${{ secrets.HELM_REPO_TOKEN }}"\
-            -H "X-GitHub-Api-Version: 2022-11-28" \
-            https://api.github.com/repos/mongodb/helm-charts/actions/workflows/post-atlas-operator-release.yaml/dispatches \
-            -d '{"ref":"main","inputs":{"version":"'"${{ steps.tag.outputs.version }}"'"}}'
       - name: Check out code
         uses: actions/checkout@v4
         with:
           submodules: true
           fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: "${{ github.workspace }}/tools/makejwt/go.mod"
+      - name: Trigger helm post release workflow
+        if: ${{ env.RELEASE_HELM == 'true' }}
+        run: |
+          make release-helm JWT_RSA_PEM_KEY_BASE64="${{ secrets.AKO_RELEASER_RSA_KEY_BASE64 }}" \
+            JWT_APP_ID="${{ secrets.AKO_RELEASER_APP_ID }}" \
+            VERSION="${{ steps.tag.outputs.version }}"
       - name: Upgrade build Makefile & Dockerfile # TODO: remove after version 1.9 is deprecated
         run: |
           git remote show origin
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ deploy/
 
 # ignore tool binaries
 tools/clean/clean
+tools/makejwt/makejwt
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,10 @@ BASE_GO_PACKAGE = github.com/mongodb/mongodb-atlas-kubernetes/v2
 GO_LICENSES = go-licenses
 DISALLOWED_LICENSES = restricted,reciprocal
 
+# JWT
+JWT_RSA_PEM_KEY_BASE64 ?= ""
+JWT_APP_ID ?= ""
+
 .DEFAULT_GOAL := help
 .PHONY: help
 help: ## Show this help screen
@@ -392,3 +396,11 @@ actions.txt: .github/workflows/ ## List GitHub Action dependencies
 check-major-version: ## Check that VERSION starts with MAJOR_VERSION
 	@[[ $(VERSION) == $(MAJOR_VERSION).* ]] && echo "Version OK" || \
 	(echo "Bad major version for $(VERSION) expected $(MAJOR_VERSION)"; exit 1)
+
+tools/makejwt/makejwt: tools/makejwt/*.go
+	cd tools/makejwt && go test . && go build .
+
+.PHONY: release-helm
+release-helm: tools/makejwt/makejwt ## kick the operator helm chart release
+	@APP_ID=$(JWT_APP_ID) RSA_PEM_KEY_BASE64=$(JWT_RSA_PEM_KEY_BASE64) \
+	VERSION=$(VERSION) ./scripts/release-helm.sh
--- a/scripts/release-helm.sh
+++ b/scripts/release-helm.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -euo pipefail
+
+helm_charts_installation_id() {
+	JWT=$1
+	curl -s -X GET -H "Accept: application/vnd.github+json" \
+		-H "Authorization: Bearer ${JWT}" \
+		-H "X-GitHub-Api-Version: 2022-11-28" \
+		"https://api.github.com/repos/mongodb/helm-charts/installation" | jq .id
+}
+
+helm_charts_token() {
+	JWT=$1
+	INSTALL_ID=$(helm_charts_installation_id "${JWT}")
+	curl -s -X POST -H "Accept: application/vnd.github+json" \
+		-H "Authorization: Bearer ${JWT}" \
+		-H "X-GitHub-Api-Version: 2022-11-28" \
+		"https://api.github.com/app/installations/${INSTALL_ID}/access_tokens" | jq -rc .token
+}
+
+JWT=$(tools/makejwt/makejwt -appId="${APP_ID}" -key="${RSA_PEM_KEY_BASE64}")
+
+ACCESS_TOKEN=$(helm_charts_token "${JWT}")
+
+curl -s --fail-with-body -X POST -H "Accept: application/vnd.github+json" \
+        -H "Authorization: Bearer ${ACCESS_TOKEN}"\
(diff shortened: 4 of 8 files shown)


josvazg:
@helderjs & @s-urbaniak does this look reasonable? I wonder how to test this properly

josvazg:
First successful test from laptop:
https://github.com/mongodb/helm-charts/actions/runs/6981764368

helderjs (approved):
LGTM, but I think we should be careful using Makefile in this direction.
* IMHO, Makefiles are hard to read by default, and as they grow they become even harder, especially with a lot of chained targets
* Scripting PHONY recipe is code we need to maintain
* Make has its own limitations, and every time we spot one we create a support app, we will end up with a dozen micro apps that, again, need some care.

josvazg:
> LGTM, but I think we should be careful using Makefile in this direction.
> 
> * IMHO, Makefiles are hard to read by default, and as they grow they become even harder, especially with a lot of chained targets
> * Scripting PHONY recipe is code we need to maintain
> * Make has its own limitations, and every time we spot one we create a support app, we will end up with a dozen micro apps that, again, need some care.

All those are valid concerns. But At the same time, they are no excuse to put logic on the CI only instead. I would like to be able to run all the heavy lifting from our laptops while the CI only contains invocations to things defined in our codebase normally, not only within the CI file set.

I am open to simplify the makefile as much as possible. Like the CI it could just invoke scripts defined elsewhere (like scripts/* folder). But as for the maintenance burden, I do not know how can we avoid it if we need the functionality.

.gitignore
```diff
@@ -45,3 +45,5 @@ deploy/
 
 # ignore tool binaries
 tools/clean/clean
+tools/makejwt/makejwt
+
```
s-urbaniak, line 49: superfluous newline

scripts/release-helm.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -euo pipefail
```
s-urbaniak, line 3: let's add a `-x` so we can debug things while this script runs.

scripts/release-helm.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
s-urbaniak, line 1: @josvazg not necessarily related to this PR: please add a shell script linter as well in the build process.

Makefile
```diff
@@ -85,6 +85,10 @@ BASE_GO_PACKAGE = github.com/mongodb/mongodb-atlas-kubernetes/v2
 GO_LICENSES = go-licenses
 DISALLOWED_LICENSES = restricted,reciprocal
 
+# JWT
+JWT_RSA_PEM_KEY_BASE64 ?= "please override rsa key in a one line base64 encoding"
+JWT_APP_ID ?= "please override app id"
```
s-urbaniak, line 90: let's keep sane default values here (i.e. empty string) and rather test for non-empty values in the actual script logic.

tools/makejwt/main.go
```diff
@@ -0,0 +1,126 @@
...
+func makeJWT(spec *JWTSpec) (string, error) {
+	// decode base64 of the key to PEM
+	pemBytes, err := base64.StdEncoding.DecodeString(string(spec.Base64PEMBytes))
+	if err != nil {
+		return "", fmt.Errorf("error decoding base64 of the PEM key: %v", err)
+	}
+
+	// Parse PEM block
+	block, _ := pem.Decode(pemBytes)
+	if block == nil || block.Type != "RSA PRIVATE KEY" {
```
s-urbaniak, line 49: let's not partially implement a RSA PEM parser. Let's stick to the `nil` and return an error that fits the golang error description:
```
if block == nil {
  return "", errors.New("no PEM data found")
}
```

As per godoc:
```
$ go doc pem.Decode
package pem // import "encoding/pem"

func Decode(data []byte) (p *Block, rest []byte)
    Decode will find the next PEM formatted block (certificate, private key etc)
    in the input. It returns that block and the remainder of the input. If no
    PEM data is found, p is nil and the whole of the input is returned in rest.
```

tools/makejwt/main.go
```diff
@@ -0,0 +1,126 @@
...
+		return "", fmt.Errorf("error decoding PEM block (Empty or NOT RSA PRIVATE KEY)")
```
s-urbaniak, line 50: use `errors.New` rather than `fmt.Errorf` if no formatting is involved.

tools/makejwt/main.go
```diff
@@ -0,0 +1,126 @@
...
+		"iss": spec.AppID,
+	}
+
+	// Create JWT token
+	token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
+
+	// Sign the token with the private key
+	tokenString, err := token.SignedString(privateKey)
+	if err != nil {
+		return "", fmt.Errorf("error signing token: %v", err)
```
s-urbaniak, line 72: s/%v/%w

tools/makejwt/main.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// ErrorEmptyAppId when the App ID was not passed in
+	ErrorEmptyAppId = fmt.Errorf("empty App Id")
+)
+
+type JWTSpec struct {
+	Base64PEMBytes string
+	AppID          string
+	Duration       time.Duration
+	Raw            bool
```
s-urbaniak, line 32: please add godoc what those fields mean.

tools/makejwt/main.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type jsonReply struct {
+	PublicKey string
+	JWT       string
+}
```
s-urbaniak, line 38: being a json reply set json tags:
```
type jsonReply struct {
	PublicKey string `json:"publicKey"`
	JWT       string `json:"jwt"`
}
```

s-urbaniak (changes requested):
see comments

.gitignore
josvazg, line 49:
```suggestion
tools/makejwt/makejwt
```

scripts/release-helm.sh
josvazg: We have one at the CI level, I would like to move it to be available from the Makefile as well. And have the CI use the same linting code path as the makefile.

scripts/release-helm.sh
s-urbaniak: ok

tools/makejwt/main.go
```diff
@@ -0,0 +1,135 @@
...
+
+func buildJsonReply(publicKey *rsa.PublicKey, jwt string) (string, error) {
+	pubKeyDER, err := asn1.Marshal(*publicKey)
+	if err != nil {
+		return "", err
+	}
+	pubKeyPem := pem.EncodeToMemory(&pem.Block{Type: "RSA PUBLIC KEY", Bytes: pubKeyDER})
+	jsonBytes, err := json.Marshal(jsonReply{PublicKey: string(pubKeyPem), JWT: jwt})
+	if err != nil {
+		return "", err
```
s-urbaniak, line 101: wrap error using fmt.Fprintf + %w

tools/makejwt/main.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if spec.Base64PEMBytes == "" {
+		return ErrorEmptyPEMKey
+	}
+	if spec.AppID == "" {
+		return ErrorEmptyAppId
+	}
+	newJWT, err := makeJWT(&spec)
+	if err != nil {
+		log.Fatalf("failed to generate JWT: %v", err)
```
s-urbaniak, line 125: this halts the complete program, this shouldn't happen here, we should bubble up the error and `os.Exit(1)` in main instead.

josvazg: right, will fix!
thx

tools/makejwt/main.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func printJWT(w io.Writer, args []string) error {
+	var spec JWTSpec
+	fs := flag.NewFlagSet("jwtSpecFlags", flag.ContinueOnError)
+	fs.StringVar(&spec.Base64PEMBytes, "key", "", "Base64 of the PEM key contents")
+	fs.StringVar(&spec.AppID, "appId", "", "Application ID for the JWT token")
+	fs.DurationVar(&spec.Duration, "duration", 10*time.Minute, "Duration the JWT token will be valid for")
+	fs.BoolVar(&spec.Raw, "raw", true, "Emit the raw JWT or a JSON with the jwt and the public key certificate")
+	if err := fs.Parse(args); err != nil {
```
s-urbaniak, line 113: nit: parsing flags should be in main, the function name `printJWT` indicates we are printing JWTs only in this function.

josvazg: uhm, the good bit of removing all logic from main is that is easier to test.

Let me refactor it a bit more.

s-urbaniak: right but parsing flags is about interacting with the terminal environment. It's an orthognal concern to printing JWTs.

s-urbaniak (approved):
LGTM, great job! 🎉

.github/workflows/release-post-merge.yml
```diff
@@ -34,6 +34,7 @@ on:
         required: true
 jobs:
   create-release:
+    environment: release
```
helderjs, line 37: This requires manual configuration on the repository settings, right?
Are you already separating workflow per env?

scripts/release-helm.sh
```diff
@@ -0,0 +1,30 @@
...
+helm_charts_token() {
+	JWT=$1
+	INSTALL_ID=$(helm_charts_installation_id "${JWT}")
+	curl -s -X POST -H "Accept: application/vnd.github+json" \
+		-H "Authorization: Bearer ${JWT}" \
+		-H "X-GitHub-Api-Version: 2022-11-28" \
+		"https://api.github.com/app/installations/${INSTALL_ID}/access_tokens" | jq -rc .token
+}
+
+JWT=$(tools/makejwt/makejwt -appId="${APP_ID}" -key="${RSA_PEM_KEY_BASE64}")
```
helderjs, line 22: Can JWTbe scriptized too? make sense do it?

.github/workflows/release-post-merge.yml
josvazg: Yes, see here:
https://github.com/mongodb/mongodb-atlas-kubernetes/settings/environments

Envs are NOT per workflow, but per access topic. After that recent task revisiting our actions security one outcome was we wanted to separate testing access from release access: there is NO need to expose release related credentials while running the cloud tests and viceversa. So those should be clearly separated environments.

Apart from that gov tests also seemed to have no need to overlap with the regular tests credentials and the openshift test also seems it can be isolated from all the rest.

scripts/release-helm.sh
josvazg: You mean turn **makejwt** into a script?
I would say no in our particular case. The examples we had for make JWT were in Python and Ruby, meaning that they are too much for bash script readability. In our case we did not have Ruby or Python dependencies here, so I would rather just use go for the tricky code tooling, just as we did with the clean tool.

PR #1265, opened 2023-11-29
CLOUDP-212108: Fix net peering cleanups

josvazg:
Seems that `ListPeeringConnections` would render only AWS related peering by default.

Apart from that, GCP VPC removals could not be done before removing their subnets.

Fixing both here should probably fix the issues for GCP. For Azure we have not detected stuck entries so far.

### All Submissions:

### DIFF
--- a/tools/clean/atlas/network_peering.go
+++ b/tools/clean/atlas/network_peering.go
@@ -8,16 +8,28 @@ import (
 	"go.mongodb.org/atlas-sdk/v20231001002/admin"
 )
 
-func (c *Cleaner) listNetworkPeering(ctx context.Context, projectID string) []admin.BaseNetworkPeeringConnectionSettings {
-	peers, _, err := c.client.NetworkPeeringApi.
-		ListPeeringConnections(ctx, projectID).
-		Execute()
-	if err != nil {
-		fmt.Println(text.FgRed.Sprintf("\tFailed to list networking peering for project %s: %s", projectID, err))
+var (
+	SupportedProviders = []string{"AWS", "AZURE", "GCP"}
+)
 
+func (c *Cleaner) listNetworkPeering(ctx context.Context, projectID string) []admin.BaseNetworkPeeringConnectionSettings {
+	peers := []admin.BaseNetworkPeeringConnectionSettings{}
+	for _, providerName := range SupportedProviders {
+		peers = append(peers, c.listNetworkPeeringForProvider(ctx, projectID, providerName)...)
+	}
+	if len(peers) == 0 {
 		return nil
 	}
+	return peers
+}
 
+func (c *Cleaner) listNetworkPeeringForProvider(ctx context.Context, projectID, providerName string) []admin.BaseNetworkPeeringConnectionSettings {
+	queryArgs := admin.ListPeeringConnectionsApiParams{GroupId: projectID, ProviderName: &providerName}
+	peers, _, err := c.client.NetworkPeeringApi.ListPeeringConnectionsWithParams(ctx, &queryArgs).Execute()
+	if err != nil {
+		fmt.Println(text.FgRed.Sprintf("\tFailed to list %s networking peering for project %s: %s", providerName, projectID, err))
+		return []admin.BaseNetworkPeeringConnectionSettings{}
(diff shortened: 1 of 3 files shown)


tools/clean/atlas/network_peering.go
```diff
@@ -73,3 +100,27 @@ func (c *Cleaner) deleteNetworkPeering(ctx context.Context, projectID string, pe
...
+		}
+		for _, peer := range peerReply.Results {
+			if peer.ContainerId == *container.Id {
+				_, _, err := c.client.NetworkPeeringApi.DeletePeeringConnection(ctx, projectID, *peer.Id).Execute()
+				if err != nil {
+					fmt.Println(text.FgRed.Sprintf("\t\t\tFailed to delete network peer %q with orphan container %q for project %q: %s",
+						*peer.Id, *container.Id, projectID, err))
+				}
+			}
+		}
```
josvazg, line 119: This did the trick to remove the GCP network peering I created manually for testing.

I now think the issue is that when we do `ListPeeringConnections` without parameters, it is NOT returning all the peerings, but only the AWS ones:
https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Network-Peering/operation/updatePeeringContainer

providerName | stringDefault: "AWS"Enum: "AWS" "AZURE" "GCP"Cloud service provider to use for this VPC peering connection.

> QUERY PARAMETERS
> ...
> providerName	
> string
> Default: "AWS"
> Enum: "AWS" "AZURE" "GCP"
> Cloud service provider to use for this VPC peering connection.

So I think the proper fix is to query with params for all providers supported at `ListNetworkPeering`. I would expect that way we would not have to deal with orphan containers at all.

cc @helderjs

s-urbaniak (approved):

tools/clean/atlas/network_peering.go
```diff
@@ -8,16 +8,28 @@ import (
...
 )
 
-func (c *Cleaner) listNetworkPeering(ctx context.Context, projectID string) []admin.BaseNetworkPeeringConnectionSettings {
-	peers, _, err := c.client.NetworkPeeringApi.
-		ListPeeringConnections(ctx, projectID).
-		Execute()
-	if err != nil {
-		fmt.Println(text.FgRed.Sprintf("\tFailed to list networking peering for project %s: %s", projectID, err))
+var (
+	SupportedProviders = []string{"AWS", "AZURE", "GCP"}
```
helderjs, line 12: In `atlas.go` there are already constants with the supported cloud providers, please reuse them

tools/clean/atlas/network_peering.go
```diff
@@ -8,16 +8,28 @@ import (
...
+)
 
+func (c *Cleaner) listNetworkPeering(ctx context.Context, projectID string) []admin.BaseNetworkPeeringConnectionSettings {
```
helderjs, line 15: See `GetProjectDependencies()` in `projects.go` and consider move the provider logic to there.
Reason is that Gov only support AWS and make a request for other provider results in an error

tools/clean/provider/gcp.go
```diff
@@ -19,18 +19,60 @@ type GCP struct {
...
 
 	networkClient       *compute.NetworksClient
+	subnetworksClient   *compute.SubnetworksClient
 	forwardRuleClient   *compute.ForwardingRulesClient
 	addressClient       *compute.AddressesClient
 	keyManagementClient *kms.KeyManagementClient
 }
 
 func (gcp *GCP) DeleteVpc(ctx context.Context, vpcName string) error {
+	vpcGetRequest := &computepb.GetNetworkRequest{
```
helderjs, line 29: Does GCP require subnets to be remove VPCs with subnets? seems strange

helderjs (changes requested):

tools/clean/atlas/network_peering.go
josvazg: ok

tools/clean/provider/gcp.go
josvazg: yes, it failed on my test until I added this code.

PR #1287, opened 2023-12-15
CLOUDP-213620: Query GitHub development metrics

josvazg:
New `metrics` tool to query development metrics from the GitHub API directly:
- Add new script `gh-access-token.sh` to get an access token to use with the `metrics` tool. Most code is extracted from the `release-helm.sh` which now invokes `gh-access-token.sh` to get the access token.
- The new tool depends on `go-github` only. For tests it also uses `testify`.
- The current version only checks for **regressions**. Queries for **flaky** tests metrics will be added later.
- Low level queries on the `go-github` library are on a thin interface so tests can mock them.
- Integration Tests use a recorder client that calls to `go-github` and records successful replies as JSON files under the directory `samples`.
- For Unit Tests a playback client is used that returns data from the corresponding JSON the file recorded earlier.
- Unit or integration tests are selected by const `unit`, which defaults to unit tests.

### All Submissions:


tools/metrics/main.go
```diff
@@ -0,0 +1,43 @@
...
+		return err
+	}
+	for slot, sr := range srr {
+		if slot == 0 {
+			fmt.Printf("%s -> %s this week:", sr.interval.start.Format(DayFormat), sr.interval.end.Format(DayFormat))
+		} else if slot == 1 {
+			fmt.Printf("%s -> %s last week:", sr.interval.start.Format(DayFormat), sr.interval.end.Format(DayFormat))
+		} else {
+			fmt.Printf("%s -> %s %d weeks ago:", sr.interval.start.Format(DayFormat), sr.interval.end.Format(DayFormat), slot)
+		}
```
s-urbaniak, line 26: nit for better readability:
```
switch slot:
  case 0:
   ...
  case 1:
    ...
  default:
   ...
```

tools/metrics/main.go
```diff
@@ -0,0 +1,43 @@
...
+	weeks := 7
+	srr, err := queryRegressions(qc, time.Now(), Weekly, weeks)
+	elapsed := time.Since(start)
+	fmt.Printf("Query took %v\n", elapsed)
+	if err != nil {
+		return err
+	}
+	for slot, sr := range srr {
+		if slot == 0 {
+			fmt.Printf("%s -> %s this week:", sr.interval.start.Format(DayFormat), sr.interval.end.Format(DayFormat))
```
s-urbaniak, line 21: suggestion: leave rendering to the caller, and the logic for string construction here. Formally, change the signature to:
```
func regressions() (string, error) {
 ...
   return fmt.Sprintf("..."), nil
...
```

and at the caller side you'd `fmt.Fprintf(os.Stdout, regressions)` to denote the fact that we're printing regression strings consistently on stdout, not stderr.

tools/metrics/query.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return "", err
+	}
+	req, err := http.NewRequest(http.MethodGet, qURL, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed prepare GET request for %q: %w", qURL, err)
+	}
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", qc.token))
+	req.Header.Add("X-GitHub-Api-Version", "2022-11-28")
+	rsp, err := qc.hc.Do(req)
```
s-urbaniak, line 37: what happens if the request fails?

s-urbaniak: (obviously the answer is that it errors out, my question is more about the bigger picture: will CI retry? will CI fail? should we retry here?)

josvazg: I think for now we can just let fail. This will be run manually, at least initially.

Also I am about to revamp the change completely using a different library.

tools/metrics/query.go
```diff
@@ -0,0 +1,70 @@
...
+			ListOptions: github.ListOptions{
+				Page:    page,
+				PerPage: PerPage,
+			},
+		})
+	return jobs, err
+}
+
+func newGoGitHubClient(token string) *github.Client {
+	client := github.NewClient(nil)
```
helderjs, line 65: I'm wondering if we would ever set up the client without a token. Does this case happen?

helderjs (approved):
LGTM, just left a simple question

tools/metrics/query.go
josvazg: Not sure, if we skip the token we could make the same queries but face rate limiting errors, so we probably won't use that.
No token is required for unit tests, but that is a use case that should not inform the runtime interface.

igor-karpukhin (approved):

tools/metrics/types.go
```diff
@@ -0,0 +1,68 @@
...
+type TestType int
+
+const (
+	Unit TestType = iota
+	Integration
+	E2E
+)
+
+func (tt TestType) String() string {
+	if tt == Unit {
```
s-urbaniak, line 42: nit: use switch statement for better readability.

tools/metrics/main.go
```diff
@@ -0,0 +1,48 @@
...
+
+func regressions() (string, error) {
+	buf := bytes.NewBufferString("")
+	start := time.Now()
+	weeks := 7
+	srr, err := QueryRegressions(NewDefaultQueryClient(), time.Now(), Weekly, weeks)
+	elapsed := time.Since(start)
+	fmt.Fprintf(buf, "Query took %v\n", elapsed)
+	if err != nil {
+		return "", err
```
s-urbaniak, line 19: in case of error you return empty string, hence the lines 16-17 have no effect here. please add error handling directly after `QueryRegressions`.

s-urbaniak (changes requested):
just a few comments, else lgtm

tools/metrics/main.go
josvazg: Done

tools/metrics/types.go
josvazg: Done

s-urbaniak (approved):
lgtm, thank you!

PR #1288, opened 2023-12-15
CLOUDP-218001: Fix IoC on controllers by injecting Atlas Client into them

helderjs:
### All Submissions:


pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -137,7 +136,7 @@ func TestProtectedAdvancedDeploymentManagedInAtlas(t *testing.T) {
 				},
 			}
 			deployment := v1.NewDeployment(project.Namespace, fakeDeployment, fakeDeployment)
-			te := newTestDeploymentEnv(t, protected, atlasClient, testK8sClient(), project, deployment)
+			te := newTestDeploymentEnv(t, true, &atlasClient, testK8sClient(), project, deployment)
```
josvazg, line 139: I liked the old version better, it was more readable. reading this line alone, what does true mean here? Before it was obvious.

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -149,44 +150,25 @@ func (r *AtlasProjectReconciler) teamReconcile(
...
-}
-
-func ensureTeamState(workflowCtx *workflow.Context, team *v1.AtlasTeam) (string, workflow.Result) {
+func ensureTeamState(workflowCtx *workflow.Context, orgID string, team *v1.AtlasTeam) (string, workflow.Result) {
 	var atlasTeam *mongodbatlas.Team
 	var err error
 
 	if team.Status.ID != "" {
-		atlasTeam, err = fetchTeamByID(workflowCtx, team.Status.ID)
+		atlasTeam, err = fetchTeamByID(workflowCtx, orgID, team.Status.ID)
```
josvazg, line 158: I think orgID was better within the custom context object, that context is for a reconciliation and it will NOT have a different value. I Understand the refactor affects it, but I would have prefered a way to leave it in the context somehow.

test/int/project_test.go
```diff
@@ -209,7 +209,7 @@ var _ = Describe("AtlasProject", Label("int", "AtlasProject"), func() {
 			createdProject.ObjectMeta = expectedProject.ObjectMeta
 			Expect(k8sClient.Create(context.Background(), expectedProject)).ToNot(HaveOccurred())
 
-			expectedCondition := status.FalseCondition(status.ProjectReadyType).WithReason(string(workflow.AtlasCredentialsNotProvided))
+			expectedCondition := status.FalseCondition(status.ProjectReadyType).WithReason(string(workflow.Internal))
```
josvazg, line 212: Why is this error type now less specific?

josvazg (approved):
LGTM with a couple of nitpicks and a question

pkg/controller/atlasproject/team_reconciler.go
helderjs: I have added orgID to the workflow context that is passed around the code.

test/int/project_test.go
helderjs: I replaced the reason for a more specific instead of using the generic one

josvazg (approved):
Still LGTM

PR #1289, opened 2023-12-18
CLOUDP-218515: Enforce binary version release correctness

josvazg:
✅ [Test Daily rebuild produces properly named versions](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/7259900211)
✅ [Test Create release checked the version correctly](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/7259948569)

### All Submissions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,9 @@ JWT_APP_ID ?= ""
 # golangci-lint
 GOLANGCI_LINT_VERSION := v1.54.2
 
+# The module import path varies after version 1
+MODULE_PATH = $(shell ./scripts/module-path.sh)
+
 .DEFAULT_GOAL := help
 .PHONY: help
 help: ## Show this help screen
@@ -166,7 +169,7 @@ bin/$(TARGET_OS)/$(TARGET_ARCH):
 
 bin/$(TARGET_OS)/$(TARGET_ARCH)/manager: $(GO_SOURCES) bin/$(TARGET_OS)/$(TARGET_ARCH)
 	@echo "Building operator with version $(VERSION); $(TARGET_OS) - $(TARGET_ARCH)"
-	CGO_ENABLED=0 GOOS=$(TARGET_OS) GOARCH=$(TARGET_ARCH) go build -o $@ -ldflags="-X github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/version.Version=$(VERSION)" cmd/manager/main.go
+	CGO_ENABLED=0 GOOS=$(TARGET_OS) GOARCH=$(TARGET_ARCH) go build -o $@ -ldflags="-X $(MODULE_PATH)/pkg/version.Version=$(VERSION)" cmd/manager/main.go
 	@touch $@
 
 bin/manager: bin/$(TARGET_OS)/$(TARGET_ARCH)/manager
@@ -408,6 +411,10 @@ check-major-version: ## Check that VERSION starts with MAJOR_VERSION
 
 tools/makejwt/makejwt: tools/makejwt/*.go
 	cd tools/makejwt && go test . && go build .
+
+.PHONY: check-version
+check-version:
+	VERSION=$(VERSION) BINARY=bin/$(TARGET_OS)/$(TARGET_ARCH)/manager ./scripts/version-check.sh
 
 .PHONY: release-helm
 release-helm: tools/makejwt/makejwt ## kick the operator helm chart release
--- a/scripts/module-path.sh
+++ b/scripts/module-path.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# TODO: remove the need for this script ASAP
+MODULE_VERSION=""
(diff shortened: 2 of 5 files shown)


.github/workflows/release-openshift.yaml
```diff
@@ -32,6 +32,8 @@ jobs:
           ref: main
           path: ${{ matrix.repository }}
           token: ${{ github.token }}
+      - name: Check binary version
+        run: make check-version VERSION=${{ matrix.version }}
```
s-urbaniak, line 36: what value does `matrix.version` have?

josvazg: It is the version of the releasable binary, which is what we want to check matches the binary version output.

josvazg: Actually you are right, that was a copy paste that did not go well.
Fixing

josvazg: fixed, I wonder if there is a safe way to test this one. I managed to test the other 2 releases.

Makefile
```diff
@@ -92,6 +92,10 @@ JWT_APP_ID ?= ""
 # golangci-lint
 GOLANGCI_LINT_VERSION := v1.54.2
 
+# The module import path varies stating with version 2
+MODULE_VERSION = $(shell if [[ "$(VERSION)" =~ 2\..* ]]; then echo "/v2"; else echo ""; fi)
```
s-urbaniak, line 96: suggestion: let's move this into `scripts`.

question: Is this going to work for versions higher than 2? We should avoid remembering bumping this here whenever we do a major release.

s-urbaniak: as an after-thought: Why would the Makefile need to be aware of older release versions? 🤔

.github/workflows/release-openshift.yaml
josvazg: I am going to withdraw the changes here, only release changes for the tested workflows which are actually in use now.
This workflow is not currently being used and, when we plam to start using it, it will need further preparations anyway, so no need to prematurely change it now.

Makefile
josvazg: The reason we use the new makefile for older builds is to allow for fast builds on versions that did not know how to make them happen. That combined with the version 2 bump mixed very badly.
I think I should follow up with a proper fix for this that does not require checking out new code into an older branch, but that should be a separate PR. For this one I would document and move that to a script as suggested.
As for v3, I plan to have a better solution in place in the next days, so we do not have to update the script to version 3 and instead we can delete such script.

josvazg: I have sent a change moving this logic to a script and making it work for any version. But I still plan to follow up with a better fix that does not overwrite the old branch Makefile.

.github/workflows/release-openshift.yaml
s-urbaniak: sgtm 👍

Makefile
s-urbaniak: @josvazg your plan sounds great, thank you! 🙌

scripts/module-path.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# TODO: remove the need for this script ASAP
+MODULE_VERSION=""
+if [[ ! "${VERSION}" =~ 1\..* ]]; then
+    if [[ "${VERSION}" =~ [0-9]+ ]]; then
```
s-urbaniak, line 8: no action needed for this PR: this is lots of bash magic 🪄 🤔

I am wondering if tools like https://github.com/magefile/mage could make our lifes easier here in the future 🤔 Most notably here I am missing a unit test asserting these evaluations work correctly.

s-urbaniak (approved):

helderjs (approved):

scripts/module-path.sh
josvazg: I tested manually:
```bash
% VERSION=1.9.9 ./scripts/module-path.sh 
github.com/mongodb/mongodb-atlas-kubernetes
jose.vazquez@M-LFJJHGWV6N mongodb-atlas-kubernetes % VERSION=2.9.9 ./scripts/module-path.sh
github.com/mongodb/mongodb-atlas-kubernetes/v2
jose.vazquez@M-LFJJHGWV6N mongodb-atlas-kubernetes % VERSION=3.9.9 ./scripts/module-path.sh
github.com/mongodb/mongodb-atlas-kubernetes/v3
jose.vazquez@M-LFJJHGWV6N mongodb-atlas-kubernetes % VERSION=4.9.9 ./scripts/module-path.sh
github.com/mongodb/mongodb-atlas-kubernetes/v4
```

Note this script is expected to go away soon, as I will make a follow up to remove the need to upgrade the makefile for older builds. Instead each version will build as it used to do and the fork will be done as alternate CI jobs.

Also about [mage](https://github.com/magefile/mage), I think it will be hard to replace make with mage at this point in the project. And if introduced as a complement, it might be adding more complexity for little gain. It would have been different if introduced since the beginning so it could replace all makefile AND scripts.

PR #1300, opened 2023-12-20
CLOUDP-185089: Upgrade to k8s api 0.27

josvazg:
Manual upgrade to k8s API version 0.27 which depends on controller-runtime version 0.15 including breaking changes.

### All Submissions:


pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -332,14 +332,15 @@ func TestKeepAnnotatedDeploymentAlwaysRemain(t *testing.T) {
...
-			deployment.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
 			customresource.SetAnnotation(deployment,
 				customresource.ResourcePolicyAnnotation,
 				customresource.ResourcePolicyKeep,
 			)
 			k8sclient := testK8sClient()
 			customresource.SetFinalizer(deployment, customresource.FinalizerLabel)
 			require.NoError(t, k8sclient.Create(context.Background(), deployment))
+			// set deletion timestamp after creation in k8s
+			deployment.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
```
igor-karpukhin, line 343: This won't be applied to the resource in K8S, but only to the resource in the operator memory. Not sure if this is made intentionally.

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -383,14 +384,15 @@ func TestDeleteAnnotatedDeploymentGetRemoved(t *testing.T) {
...
-			deployment.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
 			customresource.SetAnnotation(deployment,
 				customresource.ResourcePolicyAnnotation,
 				customresource.ResourcePolicyDelete,
 			)
 			k8sclient := testK8sClient()
 			customresource.SetFinalizer(deployment, customresource.FinalizerLabel)
 			require.NoError(t, k8sclient.Create(context.Background(), deployment))
+			// set deletion timestamp after creation in k8s
+			deployment.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
```
igor-karpukhin, line 395: Same thing here

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -854,9 +860,12 @@ func TestReconciliation(t *testing.T) {
 		sch.AddKnownTypes(v1.GroupVersion, &v1.AtlasBackupScheduleList{})
 		sch.AddKnownTypes(v1.GroupVersion, &v1.AtlasBackupPolicy{})
 		sch.AddKnownTypes(v1.GroupVersion, &v1.AtlasDatabaseUserList{})
+		// Subresources need to be explicitly set now since controller-runtime 1.15
+		// https://github.com/kubernetes-sigs/controller-runtime/issues/2362#issuecomment-1698194188
 		k8sClient := fake.NewClientBuilder().
 			WithScheme(sch).
 			WithObjects(secret, project, bPolicy, bSchedule, deployment).
+			WithStatusSubresource(bPolicy, bSchedule).
```
igor-karpukhin, line 868: Is this just for tracking the resources (controlled by the same controller: `AtlasDeployment`) status?

igor-karpukhin (commented):
Looks great! But I have a few questions (see comments)

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
josvazg: Yes and no. Yes I moved this here intentionally, as it was on line 335 before. The reason is that it seems that the creation step now will reset the deletion timestamp, as if it had never been set, sp the tests later would fail.

And no, it was not my intention for the timestamp to be only in memory, but for this test case it is fine. The important bit is that the reconciliation that runs later sees that deletion timestamp set and acts accordingly. So I do not think we need an update here.

Take away, seems the new library version will NOT allow you to create a k8s object that is already marked for deletion. It kind of makes sense, but the old library would respect the timestamp you set there.

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
josvazg: Same issue and explanation here.

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -332,14 +332,15 @@ func TestKeepAnnotatedDeploymentAlwaysRemain(t *testing.T) {
...
 			deployment := v1.NewDeployment(project.Namespace, fakeDeployment, fakeDeployment)
-			deployment.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
 			customresource.SetAnnotation(deployment,
 				customresource.ResourcePolicyAnnotation,
 				customresource.ResourcePolicyKeep,
 			)
 			k8sclient := testK8sClient()
 			customresource.SetFinalizer(deployment, customresource.FinalizerLabel)
 			require.NoError(t, k8sclient.Create(context.Background(), deployment))
+			// set deletion timestamp after creation in k8s
```
josvazg, line 342:
```suggestion
			// set deletion timestamp after creation in k8s, otherwise the creation would reset the deletion timestamp
```

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -383,14 +384,15 @@ func TestDeleteAnnotatedDeploymentGetRemoved(t *testing.T) {
...
 			deployment := v1.NewDeployment(project.Namespace, fakeDeployment, fakeDeployment)
-			deployment.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
 			customresource.SetAnnotation(deployment,
 				customresource.ResourcePolicyAnnotation,
 				customresource.ResourcePolicyDelete,
 			)
 			k8sclient := testK8sClient()
 			customresource.SetFinalizer(deployment, customresource.FinalizerLabel)
 			require.NoError(t, k8sclient.Create(context.Background(), deployment))
+			// set deletion timestamp after creation in k8s
```
josvazg, line 394:
```suggestion
			// set deletion timestamp after creation in k8s, otherwise the creation would reset the deletion timestamp
```

igor-karpukhin (approved):

PR #1302, opened 2023-12-21
CLOUDP-219167: Compute daily builds at runtime

josvazg:
Produce the supported list of releases at CI runtime rather than maintain a static list.

✅ [Daily builds produce a proper list of versions to build at CI runtime](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/7301803453)
✅ [Invocation to build a custom list of releases](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/7301805566)

### All Submissions:


igor-karpukhin:
@josvazg I think it is better to move this part to the Release process, rather than keep it in the `Makefile`. WDYT? cc @helderjs @s-urbaniak

s-urbaniak:
i am not 100% clear what the daily-builds.json actually does and what side-effects it implies. 🤔 do you mind to explain?

josvazg:
> i am not 100% clear what the daily-builds.json actually does and what side-effects it implies. 🤔 do you mind to explain?

The [daily-builds.json file](https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/daily-builds.json) holds 2 pieces of information:
- List of versions we [rebuild dailies for here](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/workflows/rebuild-released-images.yaml).
- List of os+architercture combinations we build.

josvazg:
> I have 2 points:
> 
> 1. Why not calculate the supported version on runtime(and get rid of the JSON) instead of have a separated job to update the json?

I think that is a better idea, yes.

> 2. Our support policy should be of an year for each release that goes out. This model of the last 3 releases is "not  valid" anymore

This is unfortunate, it complicates the logic quite a bit. I will try to figure it out.

scripts/supported-releases.sh
```diff
@@ -0,0 +1,42 @@
...
+supported_versions() {
+    local one_year_ago
+    one_year_ago=$(date +%s -d "1 year ago" 2> /dev/null || date -j -v-1y +%s)
+    # Filter out version X.Y.0 happening before a year ago
+    while IFS='' read -r line; do
+      local release
+      local released_date
+      released_date=$(echo "${line}" | awk '{print $2}')
+      if (( released_date > one_year_ago)); then
+        release=$(echo "${line}" | awk '{print $1}' |awk -F/ '{print $3}' | sed 's/\.0$//')
```
igor-karpukhin, line 16: Does `sed` works in this case as it should for both Mac and Linux?

josvazg:
> We can get rid of the `daily-builds.json`. My only wish for daily builds is to be able to rebuild any of it manually in case of need

We could add an test input field at workflow invocation time, so it takes a static list, rather than discover it.
Would that work?

scripts/supported-releases.sh
josvazg: The script worked initially for me in Mac, then failed in the ubuntu CI. I used a local ubuntu container to debug it, but the only thing affected was the grep regex expression, as Mac uses POSIX regex and GNU grep uses Perl' s regex.
The current expression seem to work properly in both.
I did not found any issues with sed in Mac or the CI.

.github/workflows/rebuild-released-images.yaml
```diff
@@ -4,7 +4,6 @@ name: Daily build
 on:
   schedule:
     - cron: "0 1 * * 1-5"
-    - cron: "0 3 * * 1-5"
```
josvazg, line 7: Not sure why we did this again at 3AM, I believe once a working day is enough.

daily-builds.json
```diff
@@ -1,14 +1,3 @@
...
-    "1.9.0",
-    "1.8.0",
-    "1.8.1",
-    "1.8.2",
-    "1.7.0",
-    "1.7.1",
-    "1.7.2",
-    "1.7.3"
-  ],
   "platforms": "linux/arm64,linux/amd64"
```
helderjs, line 2: Why not add this to the workflow matrix?

helderjs (changes requested):

daily-builds.json
josvazg: I don't understand, can you elaborate?
We no longer need the static list of supported versions, we are calculating it at runtime now.

helderjs: you're calculating the versions but the json file already exists to hold the list of platforms. I think you can move it to the workflow matrix

josvazg: You mean move the `platforms` list to the workflow Ci so we can get rid of this daily-builds.json file altogether?

josvazg: Just did that change

igor-karpukhin (approved):
Good job! 👍

helderjs (approved):

PR #1303, opened 2023-12-21
CLOUDP-219172: Respect old build flows

josvazg:
- Old versions use the old `Dockerfile`, rather than `fast.Dockerfile`.
- Only newest versions use `make check-version`, those which have id available.
- Module path detection logic goes away.

✅ [Daily builds work  after the changes](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/7290509277)

### All Submissions:


.github/workflows/rebuild-released-images.yaml
```diff
@@ -87,8 +82,22 @@ jobs:
         shell: bash
         run: |
           go mod download
+      - name: Choose Dockerfile
+        id: pick-dockerfile
+        run: |
+          if test -f "fast.Dockerfile"; then
+            echo "dockerfile=fast.Dockerfile" >> $GITHUB_OUTPUT
```
s-urbaniak, line 89: the concatenation using `>>` looks dangerous, why can't we use `>` ?

.github/workflows/release-post-merge.yml
```diff
@@ -83,17 +83,27 @@ jobs:
...
-      - name: Upgrade build Makefile & Dockerfile # TODO: remove after version 1.9 is deprecated
+      - name: Choose Dockerfile
+        id: pick-dockerfile
         run: |
-          git remote show origin
-          git checkout origin/${{ github.ref_name }} -- Makefile fast.Dockerfile scripts/version-check.sh scripts/module-path.sh
+          if test -f "fast.Dockerfile"; then
+            echo "dockerfile=fast.Dockerfile" >> $GITHUB_OUTPUT
+          else
+            echo "dockerfile=Dockerfile" >> $GITHUB_OUTPUT
```
s-urbaniak, line 92: same question as above. Additionally, this looks like copy-pasta 🤔 can we consolidate this in a script?

.github/workflows/release-post-merge.yml
```diff
@@ -83,17 +83,27 @@ jobs:
...
+          fi
       - name: Build all platforms & check version
-        run: make all-platforms check-version VERSION=${{ github.event.inputs.version }}
+        if: steps.pick-dockerfile.outputs.dockerfile == 'fast.Dockerfile'
+        run: |
+          make all-platforms VERSION=${{ github.event.inputs.version }}
+          # not all versiions Makefiles support the version check
+          if make |grep -q check-version; then
```
s-urbaniak, line 99:
```suggestion
          if make | grep -q check-version; then
```

Makefile
```diff
@@ -169,7 +166,7 @@ bin/$(TARGET_OS)/$(TARGET_ARCH):
 
 bin/$(TARGET_OS)/$(TARGET_ARCH)/manager: $(GO_SOURCES) bin/$(TARGET_OS)/$(TARGET_ARCH)
 	@echo "Building operator with version $(VERSION); $(TARGET_OS) - $(TARGET_ARCH)"
-	CGO_ENABLED=0 GOOS=$(TARGET_OS) GOARCH=$(TARGET_ARCH) go build -o $@ -ldflags="-X $(MODULE_PATH)/pkg/version.Version=$(VERSION)" cmd/manager/main.go
+	CGO_ENABLED=0 GOOS=$(TARGET_OS) GOARCH=$(TARGET_ARCH) go build -o $@ -ldflags="-X github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/version.Version=$(VERSION)" cmd/manager/main.go
```
s-urbaniak, line 169: That's cool that we can put this back in here. How do we ensure though that old builds still work?

.github/workflows/rebuild-released-images.yaml
josvazg: Seems to be the standard for GitHub outputs:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-masking-a-generated-output-within-a-single-job

So that you do not remove the previous arguments dump in there by mistake.

.github/workflows/release-post-merge.yml
josvazg: Same reply as above.

I also do not see an easy way to consolidate, note the previous step runs conditionally.

Makefile
josvazg: Because this change will use the build mechanism of each release, as it will NOT overwrite any of the files from the old branch:
- Old releases without the `fast.Dockerfile` will use the regular `Dockerfile` and build as they used to (quite slow BTW).
- New releases with `fast.Dockerfile` will use it.
- Newest releases with `make `check-version` target in the `Makefile` will also include that check.

s-urbaniak: perfect, thank you for the rationale!

s-urbaniak (approved):
LGTM

igor-karpukhin (approved):

PR #1317, opened 2024-01-11
CLOUDP-222194: Add a manager builder

helderjs:
### All Submissions:


internal/manager/config.go
```diff
@@ -0,0 +1,51 @@
...
+	// dev note: we pass the watched namespace as the env variable to use the Kubernetes Downward API. Unfortunately
+	// there is no way to use it for container arguments
+	watchedNamespace := os.Getenv("WATCH_NAMESPACE")
+
+	c.Namespaces = make([]string, 0, 1)
+
+	for _, namespace := range strings.Split(watchedNamespace, ",") {
+		c.Namespaces = append(c.Namespaces, strings.TrimSpace(namespace))
+	}
+}
```
josvazg, line 35: The non main code should not manage OS env vars or CLI flags, those are main CLI concepts.

I would move this to the main, but leave the config struct definition here.

internal/manager/config.go
```diff
@@ -0,0 +1,51 @@
...
+		SyncPeriod:             SyncPeriodDefault,
+	}
+}
+
+func RegisterFlags(conf *Config, fs *flag.FlagSet) {
+	fs.StringVar(&conf.MetricsBindAddress, "metrics-bind-address", MetricsBindAddressDefault, "The address the metric endpoint binds to.")
+	fs.StringVar(&conf.HealthProbeBindAddress, "health-probe-bind-address", HealthProbeBindAddressDefault, "The address the probe endpoint binds to.")
+	fs.BoolVar(&conf.EnableLeaderElection, "leader-elect", LeaderElectionDefault,
+		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+}
```
josvazg, line 51: Same here, I do not think flags belong here, they belong in the main.

BTW did not we have a lot more flags before?

This config.go file makes sense to me to define the default config and the config struct but not how the main CLI would fill those from env vars and CLI flags.

internal/manager/builder.go
```diff
@@ -0,0 +1,128 @@
+package manager
```
josvazg, line 1: BTW, why is this internal?
Do we want to forbid users from launching our manager?
I think they still can do it by leveraging pkg/controller code. I would give them a proper way instead of forcing then at a lower level.

internal/manager/config.go
```diff
@@ -0,0 +1,51 @@
+package manager
```
josvazg, line 1: I would move this with builder into pkg/

pkg/controller/atlas/client.go
```diff
@@ -0,0 +1,27 @@
+package atlas
+
+import (
+	"net/http"
+
+	"go.mongodb.org/atlas/mongodbatlas"
+	"go.uber.org/zap"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/util/httputil"
```
josvazg, line 9: this id dying soon! ;-)

pkg/controller/atlas/client.go
```diff
@@ -0,0 +1,27 @@
+package atlas
```
josvazg, line 1: Why do we need this separate here now and not before?

pkg/controller/config.go
```diff
@@ -0,0 +1,147 @@
+package controller
```
josvazg, line 1: feels weird to me we have a private internal config and another one here. Can you elaborate on the reasons for this?

pkg/controller/config.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (c *Config) ParseFlagsFromEnv(fs *flag.FlagSet) {
+	if c == nil {
+		return
+	}
+
+	parseDeletionProtection(c, fs)
+	c.FeatureFlags = featureflags.NewFeatureFlags(os.Environ)
+}
```
josvazg, line 46: IMHO as said before CLI stuff belongs on the main code. Internal code should be agnostic on how the config was put together, which should also make it more natural for internal code calling this or other consumers of the code.

pkg/controller/config.go
```diff
@@ -0,0 +1,147 @@
...
+
+func APISecretDefault() (client.ObjectKey, error) {
+	parts := strings.Split(getOperatorPodName(), "-")
+	if len(parts) <= 2 {
+		return client.ObjectKey{}, fmt.Errorf("the pod name must follow the format \"<deployment_name>-797f946f88-97f2q\" but got %s", getOperatorPodName())
+	}
+	deploymentName := strings.Join(parts[0:len(parts)-2], "-")
+
+	return client.ObjectKey{Namespace: GetOperatorNamespace(), Name: fmt.Sprintf("%s-api-key", deploymentName)}, nil
+}
```
josvazg, line 82: IMHO this CLI stuff belongs in main

pkg/controller/config.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	return operatorPodName
+}
+
+func parseDeletionProtection(c *Config, fs *flag.FlagSet) {
+	objectDeletionSet := false
+	subObjectDeletionSet := false
+
+	fs.Visit(func(f *flag.Flag) {
```
josvazg, line 114: I am on the fence on this, I think the logic might make sense here, but this package should not be aware of CLI flags at all. Maybe we should convert those flags to a struct used here and ion the main CLI code we do the conversion from flags to the struct this uses?

pkg/controller/config_test.go
```diff
@@ -0,0 +1,246 @@
...
+	"flag"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestParseDeletionProtection(t *testing.T) {
+	t.Run("should use default values when flags or env vars were not set", func(t *testing.T) {
```
josvazg, line 13: env vars and flags test belong to main_test IMHO

here we would test config internal logic if any, maybe thing like `ParseDeletionProtection` although without flags it would be more like `SetupDeletionProtection`.

josvazg (commented):
LGTM, this is great stuff.

My biggest issue with the current shape is that we are teaching internal code about CLI stuff like env vars and arg flags it should not be concerned about. Internal code in my mind should deal with already prepared structs. It is the main code the one that is CLI aware and should do all translations required.

Another thing missing here is removing teh ducplicated code from e2e doing the same initialisation and change it to use this. But it might be a better idea to make that into a separate PR, in which case I 👍 that.

internal/manager/builder.go
igor-karpukhin: @josvazg  `internal` usually contains code that is not supposed to be imported by other projects

josvazg: Precisely, I am not sure we want to prevent consumers from being able to use Setup when they still can go and use pkg code directly, at a lower level and with more pitfalls for them. I'd rather let them use what we use.

PR #1318, opened 2024-01-11
CLOUDP-221636: Consolidate on context.Background

josvazg:
We should use `context.Background()` everywhere. TODO() is only for cases where code is still work in progress:
https://pkg.go.dev/context#TODO
> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

### All Submissions:


helderjs (approved):
Great Search & Replace :smile:

igor-karpukhin (approved):
LGTM

PR #1319, opened 2024-01-11
CLOUDP-222237: Make integration tests use atlas-go-sdk

helderjs:
### All Submissions:


pkg/controller/atlas/provider.go
```diff
@@ -111,6 +113,46 @@ func (p *ProductionProvider) Client(ctx context.Context, secretRef *client.Objec
...
+	//if err != nil {
+	//	return nil, "", err
+	//}
+
+	c, err := NewClient(p.domain, secretData[publicAPIKey], secretData[privateAPIKey])
+
+	return c, secretData[orgIDKey], err
+}
+
+func getSecrets(ctx context.Context, k8sClient client.Client, secretRef, fallbackRef *client.ObjectKey) (map[string]string, error) {
```
josvazg, line 134: why not return a explicit struct with the fields we use for the secret instead of a map?

test/helper/atlas/waiters.go
```diff
@@ -54,34 +56,7 @@ func WaitForAtlasDatabaseUserStateToNotBeReached(ctx context.Context, atlasClien
...
 			}
 
-			g.Expect(allTrue).To(g.BeFalse())
-
-			return allTrue
-		}
-	}
-}
-
-// WaitForAtlasProjectStateToNotBeReached periodically checks the given atlas project for a given condition.
```
josvazg, line 64: no longer needed?

helderjs: Not used anywhere

test/helper/atlas/waiters.go
```diff
@@ -27,7 +27,7 @@ func WaitForAtlasDeploymentStateToNotBeReached(ctx context.Context, atlasClient
 				}
 			}
 
-			g.Expect(allTrue).To(g.BeFalse())
+			Expect(allTrue).To(BeFalse())
```
igor-karpukhin, line 30: Was this method ever accepting the `g Gomega`?

helderjs: it was an alias in the module import

test/int/databaseuser_unprotected_test.go
```diff
@@ -785,40 +795,30 @@ func checkSecretsDontExist(namespace string, secretNames []string) func() bool {
...
+		user.SetLabels([]admin.ComponentLabel{})
 	}
-	if user.OIDCAuthType == "" {
-		user.OIDCAuthType = "NONE"
-	}
-	if user.DeleteAfterDate != "" {
-		user.DeleteAfterDate = timeutil.FormatISO8601(timeutil.MustParseISO8601(user.DeleteAfterDate))
-	}
-	user.GroupID = projectID
-	user.Password = ""
```
igor-karpukhin, line 821: This is not needed any more? Does the `DbUser` from the new SDK not require any normalization?

igor-karpukhin (approved):
Great work! 👍

test/helper/atlas/waiters.go
igor-karpukhin: Ok, I see. Just wanted to make sure it is not used in the `Eventually` blocks without accepting a `Gomega`

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -219,6 +221,20 @@ func (p AtlasDatabaseUser) ToAtlas(ctx context.Context, kubeClient client.Client
...
+	password, err := p.ReadPassword(ctx, kubeClient)
+	if err != nil {
+		return nil, err
+	}
+
+	result := admin.NewCloudDatabaseUserWithDefaults()
+	err = compat.JSONCopy(result, p.Spec)
+	result.Password = &password
+
+	return result, err
```
s-urbaniak, line 235: either `return nil, err` or `return result, nil`, never both.

pkg/controller/atlas/provider.go
```diff
@@ -108,17 +103,63 @@ func (p *ProductionProvider) Client(ctx context.Context, secretRef *client.Objec
...
-	for _, key := range requiredKeys {
-		if _, ok := secretData[key]; !ok {
-			missingFields = append(missingFields, key)
-		}
+	// TODO review we need add a custom logger to http client
+	//httpClientWithCustomLogger := http.DefaultClient
+	//err = httputil.LoggingTransport(log)(http.DefaultClient)
+	//if err != nil {
+	//	return nil, "", err
+	//}
```
s-urbaniak, line 120: don't check in commented code.

pkg/controller/atlas/provider.go
```diff
@@ -108,17 +103,63 @@ func (p *ProductionProvider) Client(ctx context.Context, secretRef *client.Objec
...
+
+	c, err := NewClient(p.domain, secretData.PublicKey, secretData.PrivateKey)
+
+	return c, secretData.OrgID, err
```
s-urbaniak, line 124: same as above, either `return nil, nil, err` or `return c, secretData.OrgID, nil`, never both.

pkg/controller/atlas/provider.go
```diff
@@ -108,17 +103,63 @@ func (p *ProductionProvider) Client(ctx context.Context, secretRef *client.Objec
...
+	}
+
+	secretData := credentialsSecret{
+		OrgID:      string(secret.Data[orgIDKey]),
+		PublicKey:  string(secret.Data[publicAPIKey]),
+		PrivateKey: string(secret.Data[privateAPIKey]),
+	}
+
+	if missingFields, valid := validateSecretData(&secretData); !valid {
+		return credentialsSecret{}, fmt.Errorf("the following fields are missing in the secret %v: %v", secretRef, missingFields)
```
s-urbaniak, line 144: why not use a pointer return for `credentialsSecret`?

test/int/integration_suite_test.go
```diff
@@ -112,57 +100,66 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+
+		cfg, err := testEnv.Start()
+		Expect(err).ToNot(HaveOccurred())
+		Expect(cfg).ToNot(BeNil())
+
+		e := gob.NewEncoder(&b)
+		err = e.Encode(*cfg)
+		Expect(err).ToNot(HaveOccurred())
+
+		fmt.Printf("Api Server is listening on %s\n", cfg.Host)
```
s-urbaniak, line 127: let's use the intrinsic logger of the test framework instead of `fmt.Println`/`fmt.Printf`

s-urbaniak (changes requested):
a couple of comments, else looks great 👍

pkg/controller/atlas/provider.go
helderjs: I left a TODO to review this in the upcoming PRs because the tests don't care about logging in the HTTP client, but there was a config for it in the operator. I'll check they are relevant to the operator and decide whether to keep or remove for real.

s-urbaniak: If we are absolutely sure that this comment goes away in a short-term subsequent PR this sounds acceptable. However, if we do know that this is going to be addressed in a subsequent PR I would still advocate in not checking in commented code but rather leave it in a local branch and address it in a subsequent PR.

helderjs: I assure you this code will either be added/removed in subsequent PRs. This function needs to be revisited because it will be replacing the current one in every controller we have.

s-urbaniak (approved):
lgtm minus the commented code. Please make sure we don't carry this over the subsequent PRs.

PR #1321, opened 2024-01-15
CLOUDP-222246: Send CI reports via Slack

josvazg:
Added:
- Summary report format for the reports.
- Script to slack the reports.
- Makefile rules to create and slack the reports.
- A new workflow to call the makefile that computes & slacks the reports.

### All Submissions:


josvazg:
Unfortunately, I think I cannot test the new workflow until this merges and main is aware of it.

.github/workflows/reports.yml
```diff
@@ -0,0 +1,29 @@
+name: CI Reports
+
+on:
+  workflow_call:
+  schedule:
+    - cron: "0 0 * * 1" # At 0:00 on Monday
```
helderjs, line 6: How long does it take to run? isn't it better to run it at 7 am instead?

helderjs (approved):
LGTM, just left a suggestion

.github/workflows/reports.yml
josvazg: The local runs were under 4 minutes for both. I guess 10 minutes tops on the CI?

Why is 7 AM better? we want the data from last week, which ends at Monday midnight

helderjs: While the message would not tag anyone, sending a message to our channel in the middle of the night from Monday to Sunday doesn't sound a good idea.

josvazg, line 6:
```suggestion
    - cron: "0 7 * * 1" # At 7:00 on Monday
```

s-urbaniak (approved):

PR #1348, opened 2024-01-31
CLOUDP-226369: Fix instance size updates when auto-scale ON

josvazg:
Atlas rejects updates with empty instance sizes with the following error:
> 400 (request "MISSING_ATTRIBUTE") The required attribute instanceSize was not specified.'

This forces us to select a value to be sent on updates even when we would want Atlas to ignore it and override it for us.

This fix does the following:

1. Instead of setting empty values when autoscaling is ON and something outside the region config was changed, the values from the Atlas side are set, when available and needed.
2. For comparing the region config or the cluster config we specifically ignore instance size differences between the Atlas status and the Kubernetes settings ONLY when autoscaling is ON for the region.

Unit tests have been modified to verify this new behavior.

### All Submissions:

### DIFF
--- a/pkg/controller/atlasdeployment/advanced_deployment.go
+++ b/pkg/controller/atlasdeployment/advanced_deployment.go
@@ -83,7 +83,7 @@ func advancedDeploymentIdle(ctx *workflow.Context, project *mdbv1.AtlasProject, 
 		return atlasDeploymentAsAtlas, workflow.Terminate(workflow.Internal, err.Error())
 	}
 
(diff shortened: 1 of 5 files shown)


igor-karpukhin (approved):
LGMT 👍

s-urbaniak (approved):
LGTM. Question: does it make sense to have an e2e test for this specific edge case as well?

josvazg:
> LGTM. Question: does it make sense to have an e2e test for this specific edge case as well?

I was checking, we might have some e2e test already that could help us with some tweaks or we might need new ones.
But I am changing gears now and moving to have the 1.9 branch fix first.

PR #1371, opened 2024-02-12
CLOUDP-229283: disable ownership detection

s-urbaniak:
### All Submissions:


cmd/manager/main.go
```diff
@@ -225,6 +232,31 @@ func main() {
 		os.Exit(1)
 	}
 
+	err = mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+		setupLog.Info(SUBOBJECT_DELETION_PROTECTION_MESSAGE)
+
+		akoPod := &corev1.Pod{}
+		err = mgr.GetClient().Get(ctx, client.ObjectKey{
```
s-urbaniak, line 239: recheck if we need RBAC for this

igor-karpukhin (approved):
Great fix! 👍

josvazg (approved):
LGTM!

Be aware though, that this is missing the subobject checks. Those are not called `isOwner`, but instead `can...Reconcile`, eg. `canUditingReconcile`.

BTW, the `IsOwner` is used on top level objects that we are pretty sure will remove the check for good, with no replacement. Where as the `can...Reconcile` will need to be revisited with a proper alternative.

An useful regexp to find the affected code is `can.*Reconcile` which renders me 36 results in 12 files. I would vote to do that in a separate PR.

These are

s-urbaniak:
@josvazg if there is something missing, we should definitely add it to this PR:

> Be aware though, that this is missing the subobject checks. Those are not called isOwner, but instead can...Reconcile, eg. canUditingReconcile.

This definitely should be covered by setting `SubObjectDeletionProtection: false` at the reconciler struct level. Please advise what else we should do here.

josvazg:
> @josvazg if there is something missing, we should definitely add it to this PR:
> 
> > Be aware though, that this is missing the subobject checks. Those are not called isOwner, but instead can...Reconcile, eg. canUditingReconcile.
> 
> This definitely should be covered by setting `SubObjectDeletionProtection: false` at the reconciler struct level. Please advise what else we should do here.

Yes, they are also disabled already, but maybe it would be good to mark the code as disabled explicitly?
We could skip and just go for a later removal. Not a big deal.

josvazg:
> We could skip and just go for a later removal. Not a big deal.

Lets do that and move on.

cmd/manager/main.go
s-urbaniak: @igor-karpukhin @josvazg: we indeed do need additional RBAC here which we don't ship today.

s-urbaniak: Removing emitting the event on the AKO pod for now. As we found, the inclusion of additional RBAC permissions is too intrusive for this release.

PR #1379, opened 2024-02-13
CLOUDP-229520: Log diff results

roothorp:
We currently use cmp.Diff() a lot to compare Atlas state with Kubernetes state. Sometimes we log this at Debug level, and sometimes we don't log it at all. We should always log this to info (if there's a diff), so customers/HELP tickets have an easier time spotting where and why reconciles are occurring.

### All Submissions:

### DIFF
--- a/pkg/controller/atlasproject/team_reconciler.go
+++ b/pkg/controller/atlasproject/team_reconciler.go
@@ -348,7 +348,11 @@ func teamsManagedByAtlas(workflowCtx *workflow.Context) customresource.AtlasChec
 		for _, username := range team.Spec.Usernames {
 			usernames = append(usernames, string(username))
 		}
-
-		return cmp.Diff(usernames, atlasTeam.Usernames) != "", nil
+		workflowCtx.Log.Infof("x")
+		if d := cmp.Diff(usernames, atlasTeam.Usernames); d != "" {
+			workflowCtx.Log.Infof("Atlas team usernames differ from spec: %s", d)
+			return true, nil
+		}
+		return false, nil
 	}
 }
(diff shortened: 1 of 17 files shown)


s-urbaniak (approved):

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -348,7 +348,11 @@ func teamsManagedByAtlas(workflowCtx *workflow.Context) customresource.AtlasChec
 		for _, username := range team.Spec.Usernames {
 			usernames = append(usernames, string(username))
 		}
-
-		return cmp.Diff(usernames, atlasTeam.Usernames) != "", nil
+		workflowCtx.Log.Infof("x")
```
josvazg, line 351: this one is not needed right?
```suggestion
```

josvazg (approved):
LGTM!

Only found some line that might not be needed.

roothorp:
Fixing these unit tests has made me realise that a lot of tests create a workflow context by hand, usually without a logger. I've fixed the ones where the diff is meant to fail (and therefore a logger is required) but now if there is an actual failure that causes a diff, I think the test will nil pointer dereference. Is it worth me going through and adding a logger to every creation of a workflow context?

josvazg:
> Fixing these unit tests has made me realise that a lot of tests create a workflow context by hand, usually without a logger. I've fixed the ones where the diff is meant to fail (and therefore a logger is required) but now if there is an actual failure that causes a diff, I think the test will nil pointer dereference. Is it worth me going through and adding a logger to every creation of a workflow context?

I think so yes. Attention to code hygiene is good here. Just do not rush it. It does not need to be included on the upcoming release IMHO.

PR #1388, opened 2024-02-20
CLOUDP-230403: Avoid error 'version' not provided

josvazg:
The version is calculated at the `Print Env and Get version` step. Such version takes the input version as a hint, or calculates it from the push event branch name.

In either case, the version to use after that step is always `steps.tag.outputs.version`, **never** `github.event.inputs.version` again. 

### All Submissions:


josvazg:
cc @helderjs @s-urbaniak @igor-karpukhin 
Problem is, I do not see an easy way to test this before an actual release. This was breaking the push workflow, not the manually triggered one we use for testing usually.

s-urbaniak (approved):

helderjs (approved):

PR #1392, opened 2024-02-20
CLOUDP-221980: Added search nodes implementation

igor-karpukhin:
### All Submissions:


config/crd/bases/atlas.mongodb.com_atlasdeployments.yaml
```diff
@@ -368,6 +368,34 @@ spec:
...
+                          - S50_LOWCPU_NVME
+                          - S60_LOWCPU_NVME
+                          - S80_LOWCPU_NVME
+                          - S90_LOWCPU_NVME
+                          - S100_LOWCPU_NVME
+                          - S110_LOWCPU_NVME
+                          type: string
+                        nodeCount:
+                          maximum: 32
+                          type: integer
```
josvazg, line 396: These does not seem to fit what we had in the docs:

**`instanceSize`** there was an expression of `{item} | |{item}` where each item was an enum with the values listed here. But here we are just saying instance size is a single `{item}`of that enum.

Also the **`nodeCount`** in examples was something like `1..32` which is a string representing a range. To get the same we would need to either user a string and parse the ranges such as `1`, `3..12`, or `..23`. Or we should provide a **min** and **max** integers with default values maybe.

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -70,6 +72,14 @@ type AtlasDeploymentSpec struct {
 	ProcessArgs *ProcessArgs `json:"processArgs,omitempty"`
 }
 
+type SearchNode struct {
+	// +kubebuilder:validation:Enum:=S20_HIGHCPU_NVME;S30_HIGHCPU_NVME;S40_HIGHCPU_NVME;S50_HIGHCPU_NVME;S60_HIGHCPU_NVME;S70_HIGHCPU_NVME;S80_HIGHCPU_NVME;S30_LOWCPU_NVME;S40_LOWCPU_NVME;S50_LOWCPU_NVME;S60_LOWCPU_NVME;S80_LOWCPU_NVME;S90_LOWCPU_NVME;S100_LOWCPU_NVME;S110_LOWCPU_NVME
+	InstanceSize string `json:"instanceSize,omitempty"`
+	// +kubebuilder:validation:Mininum:=2
+	// +kubebuilder:validation:Maximum:=32
+	NodeCount uint8 `json:"nodeCount,omitempty"`
```
josvazg, line 80: As stated before, I think these types do not support the examples I saw in specs.

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +81,94 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
 	}
 }
 
+func handleSearchNodes(ctx *workflow.Context, deployment *mdbv1.AtlasDeployment, projectID string) error {
```
josvazg, line 84: I think it makes sense to move this to its own file `searchnodes.go`

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -138,6 +138,14 @@ func (r *AtlasDeploymentReconciler) Reconcile(context context.Context, req ctrl.
 		return result.ReconcileResult(), nil
 	}
 
+	atlasSdkClient, _, err := r.AtlasProvider.SdkClient(workflowCtx.Context, project.ConnectionSecretObjectKey(), log)
```
josvazg, line 141: Why is this needed here now?

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -23,6 +23,8 @@ import (
 	"regexp"
 	"strconv"
 
+	"go.mongodb.org/atlas-sdk/v20231115004/admin"
```
s-urbaniak, line 26: package imports

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -139,6 +152,21 @@ func (s *AdvancedDeploymentSpec) ToAtlas() (*mongodbatlas.AdvancedCluster, error
 	return result, err
 }
 
+func (s *AdvancedDeploymentSpec) SearchNodesToAtlas() *[]admin.ApiSearchDeploymentSpec {
+	if s.SearchNodes == nil {
```
s-urbaniak, line 156:
```suggestion
	if len(s.SearchNodes) == 0 {
```

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -4,8 +4,11 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 
+	"go.mongodb.org/atlas-sdk/v20231115004/admin"
```
s-urbaniak, line 10: package ordering

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +80,87 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
 }
 
+func handleSearchNodes(ctx *workflow.Context, deployment *akov2.AtlasDeployment, projectID string) workflow.Result {
+	// Get the current state from Atlas
+	// If no nodes configured in atlas, but some in the operator - create them. If not - do nothing
+	// If nodes already configured in atlas and in the operator - update them. If not - delete them
+	ctx.Log.Debug("starting search node processing")
+	defer ctx.Log.Debug("finished search node processing")
+
+	nodesInAkoEmpty := deployment.Spec.DeploymentSpec.SearchNodes == nil || len(deployment.Spec.DeploymentSpec.SearchNodes) == 0
```
s-urbaniak, line 90: move this between line 115 and 116

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +80,87 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+			nodesInAtlasEmpty = true
+		default:
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
+		}
+	}
+
+	if !nodesInAtlasEmpty && currentNodesInAtlas.GetStateName() != "IDLE" {
+		msg := fmt.Sprintf("search nodes are not ready: %v", currentNodesInAtlas.GetStateName())
+		ctx.Log.Debug(msg)
+		return workflow.InProgress(workflow.SearchNodesReady, msg)
```
s-urbaniak, line 113: what if `nodesInAkoEmpty` is `true` in this case?

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +80,87 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
 	}
 }
 
+func handleSearchNodes(ctx *workflow.Context, deployment *akov2.AtlasDeployment, projectID string) workflow.Result {
+	// Get the current state from Atlas
+	// If no nodes configured in atlas, but some in the operator - create them. If not - do nothing
```
s-urbaniak, line 85: put this to line 141

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +80,87 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+	// If nodes already configured in atlas and in the operator - update them. If not - delete them
```
s-urbaniak, line 86: put this to line 126

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +80,87 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+	}
+
+	if !nodesInAtlasEmpty && currentNodesInAtlas.GetStateName() != "IDLE" {
+		msg := fmt.Sprintf("search nodes are not ready: %v", currentNodesInAtlas.GetStateName())
+		ctx.Log.Debug(msg)
+		return workflow.InProgress(workflow.SearchNodesReady, msg)
+	}
+
+	switch {
+	case !nodesInAkoEmpty && nodesInAtlasEmpty:
```
s-urbaniak, line 117: add a comment what this case handles like in the other ones.

pkg/api/v1/atlasdeployment_types.go
roothorp: Hopefully clarified in my above comment

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -851,8 +854,40 @@ func TestReconciliation(t *testing.T) {
...
+					StateName: pointer.MakePtr("IDLE"),
+					Specs: &[]admin.ApiSearchDeploymentSpec{
+						{
+							InstanceSize: "S100_LOWCPU_NVME",
+							NodeCount:    4,
+						},
+					},
+				},
+				&http.Response{},
+				nil,
```
s-urbaniak, line 873: are we testing all the cases with this one unit test? what is the coverage?

s-urbaniak (changes requested):
good start, my biggest concern is about potentially missing unit testing and missing normalization of search node slices.

config/crd/bases/atlas.mongodb.com_atlasdeployments.yaml
```diff
@@ -362,6 +362,35 @@ spec:
...
+                          - S90_LOWCPU_NVME
+                          - S100_LOWCPU_NVME
+                          - S110_LOWCPU_NVME
+                          type: string
+                        nodeCount:
+                          maximum: 32
+                          minimum: 2
+                          type: integer
+                      type: object
+                    type: array
```
s-urbaniak, line 393: if the API only allows one entry, we should limit it here as well.

roothorp: Done

pkg/controller/atlasdeployment/advanced_deployment.go
roothorp: Done

pkg/api/v1/atlasdeployment_types.go
roothorp: Is this fixed?

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +80,87 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+			Specs: deployment.Spec.DeploymentSpec.SearchNodesToAtlas(),
+		}).Execute()
+		if err != nil {
+			ctx.Log.Debugf("unable to create search nodes: %v", err)
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
+		}
+	case !nodesInAkoEmpty && !nodesInAtlasEmpty:
+		ctx.Log.Debugf("updating search nodes %v", deployment.Spec.DeploymentSpec.SearchNodes)
+		currentAkoNodesAsAtlas := deployment.Spec.DeploymentSpec.SearchNodesToAtlas()
+		if reflect.DeepEqual(currentAkoNodesAsAtlas, currentNodesInAtlas) {
```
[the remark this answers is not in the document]
roothorp: We don't need to normalize this because the list cannot be longer than 1 entry

pkg/controller/atlasdeployment/advanced_deployment.go
roothorp: Moved this down so that we check if the nodes are idle only if the AKO and Atlas specs are the same.

pkg/controller/atlasdeployment/advanced_deployment.go
roothorp: Hopefully fixed as well

pkg/controller/atlasdeployment/advanced_deployment.go
s-urbaniak: agreed

config/crd/bases/atlas.mongodb.com_atlasdeployments.yaml
```diff
@@ -362,6 +362,36 @@ spec:
                     type: array
                   rootCertType:
                     type: string
+                  searchNodes:
+                    description: Settings for Search Nodes for the cluster. Despite
```
s-urbaniak, line 366:
```suggestion
                    description: Settings for Search Nodes for the cluster. Currently, at most one search node configuration may be defined.
```

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -130,6 +139,10 @@ type AdvancedDeploymentSpec struct {
 	// Flag that indicates whether termination protection is enabled on the cluster. If set to true, MongoDB Cloud won't delete the cluster. If set to false, MongoDB Cloud will delete the cluster.
 	// +kubebuilder:default:=false
 	TerminationProtectionEnabled bool `json:"terminationProtectionEnabled,omitempty"`
+	// Settings for Search Nodes for the cluster. Despite being a list, only one configuration may be defined.
```
s-urbaniak, line 142:
```suggestion
	// Settings for Search Nodes for the cluster. Currently, at most one search node configuration may be defined.
```

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+		// TODO: Should be only NotFound: CLOUDP-239015
+		case http.StatusBadRequest, http.StatusNotFound:
+			ctx.Log.Debug("no search nodes in atlas found")
+			nodesInAtlasEmpty = true
+		default:
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
+		}
+	}
+
+	nodesInAkoEmpty := deployment.Spec.DeploymentSpec.SearchNodes == nil || len(deployment.Spec.DeploymentSpec.SearchNodes) == 0
```
s-urbaniak, line 106: This can be reduced to:
```suggestion
	nodesInAkoEmpty := len(deployment.Spec.DeploymentSpec.SearchNodes) == 0
```
because it will also catch the `nil` case.

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+
+	switch {
+	case !nodesInAkoEmpty && nodesInAtlasEmpty:
+		// If no nodes configured in atlas, but some in the operator - create them.
+		ctx.Log.Debugf("creating search nodes %v", deployment.Spec.DeploymentSpec.SearchNodes)
+		_, _, err = ctx.SdkClient.AtlasSearchApi.CreateAtlasSearchDeployment(ctx.Context, projectID, deployment.GetDeploymentName(), &admin.ApiSearchDeploymentRequest{
+			Specs: deployment.Spec.DeploymentSpec.SearchNodesToAtlas(),
+		}).Execute()
+		if err != nil {
+			ctx.Log.Debugf("unable to create search nodes: %v", err)
```
s-urbaniak, line 116: two questions:
1. Why is this a `ctx.Log.Debugf` and not a `ctx.Log.Errorf`?
2. Generally: Can't we handle non-`workflow.OK()` results including error logs higher in the chain to reduce chattiness here?

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
```
s-urbaniak, line 117:
```suggestion
			return workflow.Terminate(workflow.SearchNodesReady, fmt.Sprintf("unable to create search nodes: %v", err.Error()))
```

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+		// If nodes already configured in atlas and in the operator - update them.
+		ctx.Log.Debugf("updating search nodes %v", deployment.Spec.DeploymentSpec.SearchNodes)
+		currentAkoNodesAsAtlas := deployment.Spec.DeploymentSpec.SearchNodesToAtlas()
+		// We can deepequal without normalization here because there is only ever 1 spec in the array
+		if reflect.DeepEqual(currentAkoNodesAsAtlas, currentNodesInAtlas.Specs) {
+			// If the nodes are not marked as IDLE, they are not yet ready.
+			if currentNodesInAtlas.GetStateName() != "IDLE" {
+				msg := fmt.Sprintf("search nodes are not ready: %v", currentNodesInAtlas.GetStateName())
+				ctx.Log.Debug(msg)
+				return workflow.InProgress(workflow.SearchNodesReady, msg)
```
s-urbaniak, line 129: same question as for the error case: Can't we handle non-`workflow.OK()` results including in progress logs higher in the chain to reduce chattiness here?

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+			ctx.Log.Debug("search nodes in AKO and Atlas are equal")
+			return workflow.OK()
+		}
+
+		_, _, err = ctx.SdkClient.AtlasSearchApi.UpdateAtlasSearchDeployment(ctx.Context, projectID, deployment.GetDeploymentName(), &admin.ApiSearchDeploymentRequest{
+			Specs: deployment.Spec.DeploymentSpec.SearchNodesToAtlas(),
+		}).Execute()
+		if err != nil {
+			ctx.Log.Debugf("unable to update search nodes: %v", err)
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
```
s-urbaniak, line 141:
```suggestion
			return workflow.Terminate(workflow.SearchNodesReady, fmt.Sprintf("unable to update search nodes: %v", err))
```

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+		}
+	case nodesInAkoEmpty && !nodesInAtlasEmpty:
+		// If no nodes configured in the operator, but some in atlas - delete them.
+		ctx.Log.Debug("deleting search nodes")
+		_, err = ctx.SdkClient.AtlasSearchApi.DeleteAtlasSearchDeployment(ctx.Context, projectID, deployment.GetDeploymentName()).Execute()
+		if err != nil {
+			ctx.Log.Debugf("unable to delete search nodes: %v", err)
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
```
s-urbaniak, line 149:
```suggestion
			return workflow.Terminate(workflow.SearchNodesReady, fmt.Sprintf("unable to delete search nodes: %v", err))
```

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+		if err != nil {
+			ctx.Log.Debugf("unable to update search nodes: %v", err)
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
+		}
+	case nodesInAkoEmpty && !nodesInAtlasEmpty:
+		// If no nodes configured in the operator, but some in atlas - delete them.
+		ctx.Log.Debug("deleting search nodes")
+		_, err = ctx.SdkClient.AtlasSearchApi.DeleteAtlasSearchDeployment(ctx.Context, projectID, deployment.GetDeploymentName()).Execute()
+		if err != nil {
+			ctx.Log.Debugf("unable to delete search nodes: %v", err)
```
s-urbaniak, line 148: see comment above

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+func handleSearchNodes(ctx *workflow.Context, deployment *akov2.AtlasDeployment, projectID string) workflow.Result {
+	// Get the current state from Atlas
+	ctx.Log.Debug("starting search node processing")
+	defer ctx.Log.Debug("finished search node processing")
+
+	// NOTE: Despite being an array/slice, the Atlas API only allows 1 item.
+	nodesInAtlasEmpty := false
+	currentNodesInAtlas, httpResp, err := ctx.SdkClient.AtlasSearchApi.GetAtlasSearchDeployment(ctx.Context, projectID, deployment.GetDeploymentName()).Execute()
+	if err != nil {
+		if httpResp == nil {
```
s-urbaniak, line 91: This whole block is not handled in unit tests.

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+			nodesInAtlasEmpty = true
+		default:
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
+		}
+	}
+
+	nodesInAkoEmpty := deployment.Spec.DeploymentSpec.SearchNodes == nil || len(deployment.Spec.DeploymentSpec.SearchNodes) == 0
+
+	switch {
+	case !nodesInAkoEmpty && nodesInAtlasEmpty:
```
s-urbaniak, line 109: This whole block is not handled in unit tests.

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -77,12 +79,91 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedDeploymentState(ctx *workflow.
...
+		if err != nil {
+			ctx.Log.Debugf("unable to create search nodes: %v", err)
+			return workflow.Terminate(workflow.SearchNodesReady, err.Error())
+		}
+	case !nodesInAkoEmpty && !nodesInAtlasEmpty:
+		// If nodes already configured in atlas and in the operator - update them.
+		ctx.Log.Debugf("updating search nodes %v", deployment.Spec.DeploymentSpec.SearchNodes)
+		currentAkoNodesAsAtlas := deployment.Spec.DeploymentSpec.SearchNodesToAtlas()
+		// We can deepequal without normalization here because there is only ever 1 spec in the array
+		if reflect.DeepEqual(currentAkoNodesAsAtlas, currentNodesInAtlas.Specs) {
```
s-urbaniak, line 124: This block is not handled in unit tests.

pkg/controller/atlasdeployment/advanced_deployment.go
s-urbaniak: Additionally: This block is not handled in unit tests.

s-urbaniak (changes requested):
Much better, I still have a couple of questions regarding logging, suggestions around status reporting, and comments about missing unit test coverage.

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
s-urbaniak: I am still missing coverage of the main reconciliation method in unit tests. See comments below.

config/crd/bases/atlas.mongodb.com_atlasdeployments.yaml
josvazg: ah ok, I got it now. Thanks for explaining!

test/e2e/search_test.go
helderjs, line 1: no big deal, but I'd make it an integration test

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -767,182 +770,274 @@ func TestUniqueKey(t *testing.T) {
...
-				Namespace: "default",
-				Labels: map[string]string{
-					"atlas.mongodb.com/type": "credentials",
-				},
-			},
-			Data: map[string][]byte{
-				"orgId":         []byte("1234567890"),
-				"publicApiKey":  []byte("a1b2c3"),
-				"privateApiKey": []byte("abcdef123456"),
+	secret := &corev1.Secret{
```
helderjs, line 773: for better isolation of each test case you need, I'd not update this test but add new cases starting with an existing cluster and test only search node parts.
A single unit test should no be used to test all paths inside the reconciliation flow, it will become unmaintainable.

helderjs (changes requested):

pkg/controller/atlasdeployment/advanced_deployment.go
s-urbaniak: agreed, with adding more meaningful reasons to the status conditions we can move logging higher in the chain.

test/e2e/search_test.go
s-urbaniak: agreed

roothorp:
I think I've addressed all concerns; I've cleaned up the status conditions and errors per @helderjs comments, and added (100%?) unit testing per @s-urbaniak. Let me know if there's anything that you think I've missed!

pkg/api/v1/atlasdeployment_types.go
```diff
@@ -70,6 +71,14 @@ type AtlasDeploymentSpec struct {
 	ProcessArgs *ProcessArgs `json:"processArgs,omitempty"`
 }
 
+type SearchNode struct {
+	// +kubebuilder:validation:Enum:=S20_HIGHCPU_NVME;S30_HIGHCPU_NVME;S40_HIGHCPU_NVME;S50_HIGHCPU_NVME;S60_HIGHCPU_NVME;S70_HIGHCPU_NVME;S80_HIGHCPU_NVME;S30_LOWCPU_NVME;S40_LOWCPU_NVME;S50_LOWCPU_NVME;S60_LOWCPU_NVME;S80_LOWCPU_NVME;S90_LOWCPU_NVME;S100_LOWCPU_NVME;S110_LOWCPU_NVME
+	InstanceSize string `json:"instanceSize,omitempty"`
+	// +kubebuilder:validation:Minimum:=2
+	// +kubebuilder:validation:Maximum:=32
+	NodeCount uint8 `json:"nodeCount,omitempty"`
```
s-urbaniak, line 79: this does not have any documentation:
```
$ k explain atlasdeployments.atlas.mongodb.com.spec.deploymentSpec.searchNodes
GROUP:      atlas.mongodb.com
KIND:       AtlasDeployment
VERSION:    v1

FIELD: searchNodes <[]Object>

DESCRIPTION:
    Settings for Search Nodes for the cluster. Currently, at most one search
    node configuration may be defined.
    
FIELDS:
  instanceSize	<string>
    <no description>

  nodeCount	<integer>
    <no description>
```

s-urbaniak (changes requested):
everything looks great except we don't set status conditions correctly.  This contribution lacks setting `SearchNodesCreating` and `SearchNodesUpdating` in all cases.

s-urbaniak:
[state machine](https://dreampuf.github.io/GraphvizOnline/#digraph%20G%20%7B%0A%20%20%20%20compound%3Dtrue%3B%0A%0A%20%20%20%20subgraph%20cluster_pending%20%7B%0A%20%20%20%20%20%20%20%20idling%20-%3E%20unmanaged%20%5Blabel%3D%22unmanage%22%5D%3B%0A%20%20%20%20%20%20%20%20terminated%3B%0A%20%20%20%20%20%20%20%20label%20%3D%20%22pending%22%3B%0A%20%20%20%20%20%20%20%20style%3D%22rounded%22%3B%0A%20%20%20%20%7D%0A%0A%20%20%20%20pending%20-%3E%20creating%20%5Blabel%3D%22create%22%5D%3B%0A%20%20%20%20creating%20-%3E%20creating%20%5Blabel%3D%22progress%22%5D%3B%0A%20%20%20%20creating%20-%3E%20pending%20%5Blabel%3D%22terminate%2C%5Cnidle%22%5D%0A%0A%20%20%20%20pending%20-%3E%20updating%20%5Blabel%3D%22update%22%5D%3B%0A%20%20%20%20updating%20-%3E%20updating%20%5Blabel%3D%22progress%22%5D%3B%0A%20%20%20%20updating%20-%3E%20pending%20%5Blabel%3D%22terminate%2C%5Cnidle%22%5D%0A%0A%20%20%20%20pending%20-%3E%20deleting%20%5Blabel%3D%22delete%22%5D%3B%0A%20%20%20%20deleting%20-%3E%20deleting%20%5Blabel%3D%22progress%22%5D%3B%0A%20%20%20%20deleting%20-%3E%20pending%20%5Blabel%3D%22terminate%2C%5Cnunmanage%22%5D%0A%0A%7D%0A):
image

s-urbaniak (approved):

igor-karpukhin:
I can not approve my own PR, but LGTM from my side 👍

pkg/api/v1/atlasdeployment_types.go
josvazg: How does one make kubebuilder pass the annotations from Go comments into the explain output?

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -767,182 +770,211 @@ func TestUniqueKey(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "api-secret",
+			Namespace: "default",
+			Labels: map[string]string{
+				"atlas.mongodb.com/type": "credentials",
```
josvazg, line 778: Not a blocker, but just expressing some preference.

These kind of tests are hard to write and hard to read. They have a great deal of setting up and one has to decipher what each setting does or does not do.

Readable tests have a setup phase, the test itself and an assertion phase. It is best if tests setups can be a few self explanatory lines. One way to achieve that would be to have explicit functions or methods for setting up the preconditions, with descriptive names. Some functions might be decomposed in smaller ones and some might be reused across tests.

It takes some initial effort, but once there is a battery of initialisation functions to be used, one can even make the tests into test case tables with a title, a initial setup status function and an expectation. Adding more tests to the table becomes, if not trivial, at least easier.

For the expectations I think some similar decomposition and simplification can be done.

The idea is that at the test code level you do not need to be into the weeds on how some setup is done, the setup function or function names say it all. Similarly with the expectations. The test code is short and easy to read, the intent is clear. If it needs to be debugged or inspected in more detail you can always go down the function calls, but on each of those there is a function name that expreses the intent of the setup or expectation which you can easily check whether or not matches the code within. And the functions can also be decomposed into finer details, like setting up substructures, etc.

Reading a long function is too much of a cognitive burden usually and would prompt some people to add comments on the way, which is a sign that you probably would have needed a function there.

pkg/controller/atlasdeployment/search_nodes.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	// determine the current controller state and invoke state handling methods
+	c, ok := ctx.GetCondition(status.SearchNodesReadyType)
+	if ok {
+		switch reason := workflow.ConditionReason(c.Reason); reason {
+		case workflow.SearchNodesCreating:
+			return s.handleUpserting(reason)
+		case workflow.SearchNodesUpdating:
+			return s.handleUpserting(reason)
```
josvazg, line 58: Why not instead?

```suggestion
		case workflow.SearchNodesCreating, workflow.SearchNodesUpdating:
			return s.handleUpserting(reason)
```

pkg/controller/atlasdeployment/search_nodes.go
```diff
@@ -0,0 +1,312 @@
...
+
+	atlasNodes, found, err := s.getAtlasSearchDeployment()
+	if err != nil {
+		// transition back to pending, something went wrong.
+		return s.terminate(workflow.ErrorSearchNodesNotDeletedInAtlas, err)
+	}
+
+	switch {
+	case err != nil:
+		return s.terminate(workflow.ErrorSearchNodesNotUpsertedInAtlas, err)
```
josvazg, line 175: This if seems redundant:

```suggestion
	atlasNodes, found, err := s.getAtlasSearchDeployment()
	switch {
	case err != nil:
		return s.terminate(workflow.ErrorSearchNodesNotUpsertedInAtlas, err)
```

pkg/controller/atlasdeployment/search_nodes.go
```diff
@@ -0,0 +1,312 @@
...
+		)
+	default:
+		return s.unmanage()
+	}
+}
+
+// create executes the actual creation of search nodes in Atlas and transitions to the following states:
+// - creating: after search nodes have been created in Atlas
+// - terminated: when an error occurred.
+func (s *searchNodeController) create() workflow.Result {
```
josvazg, line 190: Nit: this almost matches a translation layer function on top of the API call. But I would probably not include workflow.Result in such a layer.

pkg/controller/atlasdeployment/search_nodes.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// unmanage transitions to pending state if no search nodes are managed.
+func (s *searchNodeController) unmanage() workflow.Result {
+	s.ctx.UnsetCondition(status.SearchNodesReadyType)
+	return workflow.OK()
+}
+
+// unmanage transitions to pending state search nodes are ready and idle.
+func (s *searchNodeController) idle() workflow.Result {
```
josvazg, line 289: This seems a copy paste of the previous comment, we should explain idle instead here

pkg/controller/atlasdeployment/search_nodes.go
```diff
@@ -0,0 +1,312 @@
...
+	s.ctx.SetConditionTrue(status.SearchNodesReadyType)
+	return workflow.OK()
+}
+
+func (s *searchNodeController) getAtlasSearchDeployment() (*admin.ApiSearchDeploymentResponse, bool, error) {
```
josvazg, line 300: Nit: to be a proper translation layer function this would return an internal operator struct rather than a admin.* API one. That way if the struct in the API evolves, the fix is in the translation layer.

josvazg (approved):
LGTM, just some side comments and nit picks. There is some code that we could remove I think.

I really like the way the state machine is broken down, makes the code way easier to follow, and I bet it will be also easier to debug later.

pkg/controller/atlasdeployment/search_nodes.go
s-urbaniak: this is a state machine transition method so it must return a `workflow.Result`

pkg/controller/atlasdeployment/search_nodes.go
s-urbaniak: oh indeed, @roothorp do you mind to fix this here?

pkg/api/v1/atlasdeployment_types.go
s-urbaniak: @josvazg this is done in API-Server by introspecting the CRD definitions, hence it is a side-effect of generating CRDs.

pkg/controller/atlasdeployment/search_nodes.go
josvazg: Yes, I think when we add the translation layer it should wrap it here. The code will be a bit silly, just a call to the translation layer and handling the return into a proper `workflow.Result`. Note that Contract Tests should NOT be aware of any workflow.Result, it is not its concern.

pkg/controller/atlasdeployment/search_nodes.go
josvazg: I am advancing what we would need in order to have a translation layer, not asking to add one right now.

But a translation layer does the following:
- Accept ONLY internal types (not API types) in function or method arguments.
- Returns ONLY internal types as results.
- Internally translate the internal argument types to API request types and the API response types back to internal types.
- Returned internal types are normalised at translation time, so they can be compared with ease.

API updates affect only code within the translation layer, unless the update is a breaking change that trickles also to internal types because some fields in use have been removed or added.

Contract tests call the translation layer and its internal types. They not need to concern with reconciliation code or the API types or functions.

pkg/api/v1/atlasdeployment_types.go
josvazg: I meant, what do we have to put in the Go code so that the generated CRDs will include documentation when using explain. I assume there are some special annotations to use or something.

pkg/controller/atlasdeployment/search_nodes.go
s-urbaniak: I think that's fine but out of scope for a `create() workflow.Result` state machine transition method 👍

pkg/controller/atlasdeployment/search_nodes.go
s-urbaniak: I think this is a nice proposal as part of https://docs.google.com/document/d/11yWbf1SGd5W3YeYCpDyGJ5w05nVyZhR3T_heSF2Hk_s/edit#heading=h.xwvxi8ifckvk but out of scope for this PR.

pkg/api/v1/atlasdeployment_types.go
s-urbaniak: yes, see https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1392/commits/728e0fb5ca88b0f2bfde945572bc452363df46f2 (part of this PR)

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
roothorp: Yeah, I think we can simplify this by making some of the more generic setup into separate functions that we can use across the codebase - I'll take a look at that now.

igor-karpukhin: @josvazg this sounds as a good topic for the team to discuss. For now, I think we can leave it as is.

pkg/controller/atlasdeployment/search_nodes.go
igor-karpukhin: @roothorp please take a look. Add the short explanation of the state like here: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1392/files/728e0fb5ca88b0f2bfde945572bc452363df46f2#diff-f0151c74d09af63db28986ce11d51e640ffeffd797faf035ea9f2ef390e11541R208-R209

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
josvazg: I will make a discuss ticket for this and close this comment.

josvazg: https://jira.mongodb.org/browse/CLOUDP-245111

PR #1399, opened 2024-02-23
Add condition to skip OpenShift test

roothorp:
### All Submissions:

### DIFF
--- a/.github/workflows/openshift-upgrade-test.yaml
+++ b/.github/workflows/openshift-upgrade-test.yaml
@@ -34,7 +34,6 @@ jobs:
     name: Prepare E2E configuration and image
     runs-on: ubuntu-latest
     environment: test
-    if: env.SKIP_OPENSHIFT != 'true'
     steps:
       - name: roo-test
         run: |
--- a/.github/workflows/openshift-upgrade-test.yaml
+++ b/.github/workflows/openshift-upgrade-test.yaml
@@ -30,6 +30,7 @@ jobs:
   e2e-tests:
     name: Prepare E2E configuration and image
     runs-on: ubuntu-latest
+    if: ${{ ! vars.SKIP_OPENSHIFT }}
     environment: test
     steps:
       - if: ${{ inputs.forked == false }}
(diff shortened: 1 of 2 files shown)


josvazg (approved):

.github/workflows/openshift-upgrade-test.yaml
```diff
@@ -33,6 +33,8 @@ jobs:
   e2e-tests:
     name: Prepare E2E configuration and image
     runs-on: ubuntu-latest
+    environment: test
+    if: vars.SKIP_OPENSHIFT != 'true'
```
josvazg, line 37: Maybe it is `env.` ?

```suggestion
    if: env.SKIP_OPENSHIFT != 'true'
```

josvazg (changes requested):
Will remove the approval while we fix

.github/workflows/openshift-upgrade-test.yaml
```diff
@@ -33,7 +33,13 @@ jobs:
   e2e-tests:
     name: Prepare E2E configuration and image
     runs-on: ubuntu-latest
+    environment: test
+    if: env.SKIP_OPENSHIFT != 'true'
```
josvazg, line 37: Lets try without the if and see the debug info
```suggestion
```

josvazg: I suspect the problem we might have here is that this ìf`chere does not have access to secrets or environment, it only can see GitHub events and things like that. Not sure though.

.github/workflows/openshift-upgrade-test.yaml
```diff
@@ -32,7 +32,9 @@ concurrency:
 jobs:
   e2e-tests:
     name: Prepare E2E configuration and image
+    if: ${{  vars.OPENSHIFT_SKIP != 'true'  }}
     runs-on: ubuntu-latest
+    environment: test
```
josvazg, line 37: It did not skip it:
https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/8021707947/job/21914395620

I honestly do not understand how this is not working

josvazg:
Hey @roothorp ! I see you tested without an environment, using a repo variable instead.

Did it work? If so that could be good enough, we move the definition outside of the environment and we are done.

Environments are useful for isolating which secrets each type of workflow can access. For this flag it is not needed.

.github/workflows/openshift-upgrade-test.yaml
```diff
@@ -24,40 +23,35 @@ on:
...
-
 concurrency:
   group: test-openshift-operator-upgrade
   cancel-in-progress: true
-
 jobs:
   e2e-tests:
     name: Prepare E2E configuration and image
     runs-on: ubuntu-latest
+    if: ${{ vars.SKIP_OPENSHIFT != 'true' }}
```
josvazg, line 33: Nice this works with the repo level variable!

josvazg (approved):
LGTM but could we add a mention here on how to use this?
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/docs/dev/ci.md

A section mentioning that the repo variable `SKIP_OPENSHIFT` can be used whenever we need to skip using the OpenShift cluster, due to maintenance or other similar issues.

The cluster is now operational again, so we should set the variable to `false` now.

josvazg (approved):
LGTM!

PR #1400, opened 2024-02-23
CLOUDP-231603: Make AKOBot create PRs as needed

josvazg:
After #1389 we have:
- Regular PRs failign if we forget to `make recompute-licenses` after changes to dependencies in `go.mod`.
- But `dependabot` PRs won't do that, so we skip that check for them.

This PR adds a new bot AKA `AKOBot` that runs after `dependabot` to ensure 2 things:
- The licenses will be recomputed WHEN needed on a new PR.
- When the PR is created, it will run cloud tests, to ensure we did not break `main`.

### All Submissions:

### DIFF
--- a/.github/workflows/akobot.yml
+++ b/.github/workflows/akobot.yml
@@ -1,0 +1,23 @@
+name: AKOBot
+
+on:
+  schedule:
+    - cron: "0 7 * * 3,5" # At 7:00 on Wednesday and Friday
+  workflow_dispatch:
+
+jobs:
+  check-licenses:
+    name: AKO Bot
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          cache: false
+
+      - name: Run akobot
+        run: make akobot
--- a/pkg/controller/touch.go
+++ b/pkg/controller/touch.go
@@ -1,0 +1,1 @@
+package controller // updated at: 2024-02-23T10:01+00:00
--- a/scripts/make-pr.sh
+++ b/scripts/make-pr.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+set -euo pipefail
+
+repo=${REPO:-mongodb/mongodb-atlas-kubernetes}
+github_token=${GITHUB_TOKEN:-unset}
+
+if [[ $(git diff --stat) = '' ]]; then
+  echo 'No PR needed, git is clean'
+  exit 0
+fi
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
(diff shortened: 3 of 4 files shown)


josvazg:
Trued to use **act** to test, but did not work. I fear I will need to merge to be able to verify it properly.

pkg/controller/touch.go
helderjs, line 1: What's that?

.github/workflows/akobot.yml
```diff
@@ -0,0 +1,23 @@
+name: AKOBot
+
+on:
+  schedule:
+    - cron: "0 7 * * 3,5" # At 7:00 on Wednesday and Friday
```
helderjs, line 5: Why a cron job?

scripts/make-pr.sh
```diff
@@ -0,0 +1,34 @@
...
+git commit -m "${title}"
+git push -fu origin akobot-update
+
+echo "Creating PR"
+curl -L \
+  -X POST \
+  -H "Accept: application/vnd.github+json" \
+  -H "Authorization: Bearer ${github_token}" \
+  -H "X-GitHub-Api-Version: 2022-11-28" \
+  "https://api.github.com/repos/${repo}/pulls" -d "{\"title\":\"${title}\"}"
```
helderjs, line 34: Suggestion: You can use Github CLI and make it better to manage

helderjs (changes requested):
I'm missing the PR managing license file as mentioned in the ticket

.github/workflows/akobot.yml
josvazg: Because we will probably need it to run after we merged the `dependabot` PRs. And some other day of the week for good measure.

pkg/controller/touch.go
josvazg: This allows the bot to force the PR to run all tests.

The idea is we could pass a flag to the make-pr.sh on whether changes should be done or not.

I am open to other simple ways to force cloud tests to run (or not).

scripts/make-pr.sh
josvazg: If we add `gh` we need to manage its installation in Makefile as well, `curl` comes preinstalled in CI and almost everywhere, and it is easier to install. I would be happy to change my mind if `gh` can be easily installable with `go get ...` but that did not work for me.

I can look into it, but for just one endpoint, depending on `gh` might be an overkill.

josvazg:
> I'm missing the PR managing license file as mentioned in the ticket

The make rule for akobot `akobot: recompute-licenses` will refresh the licenses WHEN needed. See #1389 for details on how recompute-licenses was added

scripts/make-pr.sh
helderjs: With GH cli we could not only make it more elegant(IMO) but also add label(cloud-tests) to the PR

helderjs:
I was wondering now, whyi s ako bot a mix of scripts and make targets instead of a more reliable and testable Go app?

josvazg:
As pointed by @roothorp, rather than fix after the fact, we could instead patch live `dependabot` PRs like this:
https://github.com/dependabot/dependabot-core/issues/935#issuecomment-603882655

Closing in favor of attempting that approach.

PR #1402, opened 2024-02-26
CLOUDP-228584: Migrate Project Cloud Provider Integration to use Go Atlas SDK client

helderjs:
### All Submissions:


pkg/controller/atlasproject/cloud_provider_integration.go
```diff
@@ -67,12 +69,14 @@ func ensureCloudProviderIntegration(workflowCtx *workflow.Context, project *mdbv
...
-	atlasCPAs, _, err := workflowCtx.Client.CloudProviderAccess.ListRoles(workflowCtx.Context, projectID)
+	atlasCPAs, _, err := workflowCtx.SdkClient.CloudProviderAccessApi.
+		ListCloudProviderAccessRoles(workflowCtx.Context, projectID).
+		Execute()
 	if err != nil {
 		return false, fmt.Errorf("unable to fetch cloud provider access from Atlas: %w", err)
 	}
 
-	AWSRoles := sortAtlasCPAsByRoleID(atlasCPAs.AWSIAMRoles)
+	AWSRoles := sortAtlasCPAsByRoleID(atlasCPAs.GetAwsIamRoles())
```
josvazg, line 79: [Go convention on initialisms](https://go.dev/wiki/CodeReviewComments#initialisms) is for this to be called:

```suggestion
	AWSRoles := sortAtlasCPAsByRoleID(atlasCPAs.GetAWSIAMRoles())
```

josvazg (approved):
LGTM with some minor formal comments

pkg/controller/atlasproject/cloud_provider_integration.go
helderjs: @josvazg, I know and usually follow it, but Go Atlas SDK decided to use a different convention and your suggestion is on a method from them. =/

josvazg: Nothing we can do then. Sorry for the noise!

pkg/controller/atlasproject/cloud_provider_integration.go
```diff
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"sort"
 
-	"go.mongodb.org/atlas/mongodbatlas"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/timeutil"
+
+	"go.mongodb.org/atlas-sdk/v20231115004/admin"
```
s-urbaniak, line 11: package sorting, please reformat to add `github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/timeutil` in the set of imports between lines 13-17.

pkg/controller/atlasproject/cloud_provider_integration.go
```diff
@@ -224,51 +228,50 @@ func getCloudProviderIntegrations(projectSpec mdbv1.AtlasProjectSpec) []mdbv1.Cl
...
+	if len(atlasCPA.GetFeatureUsages()) > 0 {
+		cpiStatus.FeatureUsages = make([]status.FeatureUsage, 0, len(atlasCPA.GetFeatureUsages()))
 
+		for _, feature := range atlasCPA.GetFeatureUsages() {
 			id := ""
 
-			if feature.FeatureID != nil {
-				id = feature.FeatureID.(string)
+			if fID, ok := feature.GetFeatureIdOk(); ok {
+				id = fmt.Sprintf("%s.%s", fID.GetGroupId(), fID.GetBucketName())
```
s-urbaniak, line 253: do we retain the format from before the migration?

pkg/controller/atlasproject/cloud_provider_integration_test.go
```diff
@@ -3,39 +3,48 @@ package atlasproject
 import (
 	"context"
 	"errors"
+	"net/http"
 	"testing"
+	"time"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/timeutil"
```
s-urbaniak, line 10: please resort imports as commented above

s-urbaniak (changes requested):
one nit about package imports

pkg/controller/atlasproject/cloud_provider_integration.go
helderjs: @s-urbaniak, No, but before making this change I verified what cause this struct to be changed and it's related to data lake(data federation), it seems that it has never been used by an operator user because it actually panics with the current implementation.

```
panic: interface conversion: interface {} is map[string]interface {}, not string [recovered]
        panic: interface conversion: interface {} is map[string]interface {}, not string
```

s-urbaniak: i see, that's a great catch then and I see that exactly this piece of code is omitted by the unit tests:
image

s-urbaniak: @helderjs in that case: please create a jira bug explaining the faulty behaviour, when it can occur and point it to this PR as well and describe under what cases a panic can occur.

please mark it as a blocker bug as well.

s-urbaniak (approved):

pkg/controller/atlasproject/cloud_provider_integration.go
helderjs: Ticket created https://jira.mongodb.org/browse/CLOUDP-235207

PR #1407, opened 2024-02-27
CLOUDP-231603: Update dependabot licenses

josvazg:
### All Submissions:


josvazg:
I think we will have to merge this first in order to test it.

helderjs (approved):

.github/workflows/update-licenses.yml
```diff
@@ -0,0 +1,33 @@
+name: Update dependabot PR Licenses
```
s-urbaniak, line 1: I would love to see each workflow action have a small paragraph in https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/docs/dev/ci.md.

How about we start with this one?

Reason: This does not include any comment what it does and why. Reasoning about the existence of this workflow can be somehow reverse engineered by looking at the `name` and the implementation of `make recompute-licenses` but it might be hard for a future reader.

s-urbaniak (approved):
one ask to add documentation for this workflow.

.github/workflows/update-licenses.yml
josvazg: Done

s-urbaniak (approved):
looks great, thank you!

roothorp (approved):
Looks awesome! You mentioned that you were concerned about adding another action dependancy; could we just do it without an action, just with bash?
e.g.

  run: |
    git add licenses.txt
    git commit -m "update licenses"
    git push ${{ github.head_ref }}

josvazg:
> Looks awesome! You mentioned that you were concerned about adding another action dependancy; could we just do it without an action, just with bash? e.g.
> 
> run: | git add licenses.txt git commit -m "update licenses" git push ${{ github.head_ref }}

I would leave it for a follow up, once we have this working. Note that we will need to conditionally create the PR if there are changes, which it is not guaranteed. The action does that for us for free I think.

Still I expect "back and forths" when we test this, we can change at that moment.

PR #1411, opened 2024-02-29
CLOUDP-231603: Fix bot permission to commit

josvazg:
### All Submissions:


josvazg:
> is it possible to test this pre-merge?

Unfortunately no that I know of. This will only trigger with the bot permissions we want to test when it is merged in `main`. If I try it out manually it will run with my permissions, which already made us think this was fine before.

PR #1416, opened 2024-03-01
CLOUDP-234072: Fix metrics to show & flag run numbers

josvazg:
- For **regressions** we count **all runs** done in the period.
- For **flakiness** we count **all successful cloud tests runs** in the period.
In all cases we flag it f the number of runs is low, meaning less than the flakiness average in the report.

Apart from that the report CLI was revamped to accept a TILL env var and also make the `Report` main entry more configurable.

### All Submissions:


helderjs (approved):
lgtm

PR #1428, opened 2024-03-05
Force retest with a label on the PR

josvazg:
### All Submissions:


josvazg:
Was testing this on my own repo and still could not make it work.

Seems not worth the effort, as the current pain is we need just to label the `dependabot` PR with `retest` so that it will run basic tests. After that, if it passes ✅  we can approve.

PR #1434, opened 2024-03-07
CLOUDP-236434: Fix role mapping patch in FedAuth

helderjs:
### All Submissions:

### DIFF
--- a/pkg/api/v1/atlasfederatedauth_types.go
+++ b/pkg/api/v1/atlasfederatedauth_types.go
@@ -67,7 +67,6 @@ func (f *AtlasFederatedAuthSpec) ToAtlas(orgID, idpID string, projectNameToID ma
 		}
 		atlasRoleMappings = append(atlasRoleMappings, admin.AuthFederationRoleMapping{
 			ExternalGroupName: roleMapping.ExternalGroupName,
-			Id:                &idpID,
 			RoleAssignments:   &atlasRoleAssignments,
 		})
 	}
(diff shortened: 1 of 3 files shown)


test/int/federated_auth_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/test/helper/resources"
 )
 
-var _ = Describe("AtlasFederatedAuth test", Label("AtlasFederatedAuth", "federated-auth"), func() {
+var _ = FDescribe("AtlasFederatedAuth test", Label("AtlasFederatedAuth", "federated-auth"), func() {
```
igor-karpukhin, line 22: Looks like a left-over from the local testing :)

pkg/api/v1/atlasfederatedauth_types.go
```diff
@@ -67,7 +67,6 @@ func (f *AtlasFederatedAuthSpec) ToAtlas(orgID, idpID string, projectNameToID ma
 		}
 		atlasRoleMappings = append(atlasRoleMappings, admin.AuthFederationRoleMapping{
 			ExternalGroupName: roleMapping.ExternalGroupName,
-			Id:                &idpID,
```
igor-karpukhin, line 70: Does it work without providing the ID?

helderjs: Yes

test/int/federated_auth_test.go
helderjs: We need a rule to catch this in the linter :grimacing:

igor-karpukhin (approved):
Great fix 👍

josvazg (approved):

PR #1438, opened 2024-03-12
[PoC] Adding contract testing

josvazg:
---
```shell
% export AKO_CONTRACT_TEST=1; go clean -testcache && go test -v ./test/contract/...
?   	github.com/mongodb/mongodb-atlas-kubernetes/v2/test/contract	[no test files]
2024/03/19 17:30:25 Wipe resources set to false
2024/03/19 17:30:25 Reusing existing resources
=== RUN   TestCreateSearchIndex
--- PASS: TestCreateSearchIndex (0.60s)
PASS
2024/03/19 17:30:26 Trying to stored search test resource references for reuse...
2024/03/19 17:30:26 Stored search test resource references for reuse
ok  	github.com/mongodb/mongodb-atlas-kubernetes/v2/test/contract/search	1.509s
2024/03/19 17:30:25 Wipe resources set to false
2024/03/19 17:30:26 Reusing existing resources
=== RUN   TestUser
2024/03/19 17:30:26 Creating user user...
2024/03/19 17:30:26 User user already created, updating password
--- PASS: TestUser (11.30s)
PASS
2024/03/19 17:30:37 Trying to stored user test resource references for reuse...
2024/03/19 17:30:37 Stored user test resource references for reuse
ok  	github.com/mongodb/mongodb-atlas-kubernetes/v2/test/contract/user	12.395s
```

### All Submissions:


test/contract/db.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+func WithDatabase(prefix string) OptResourceFunc {
+	return func(ctx context.Context, resources *TestResources) (*TestResources, error) {
+		//
+		auth := options.Credential{
+			AuthMechanism: "SCRAM-SHA-1",
```
wtrocki, line 16: Why using SHA-1?

josvazg: SCRAM-SHA-256 never works in my tests. SCRAM-SHA-1 works sometimes.

josvazg:
Feedback after discussing with @helderjs 

1. Not only we NOT use the API calls directly on the tests, which is pending in current PR but also...
2. Try to reuse the production functions instead of direct calls for setting up the test infra.

test/contract/db.go
josvazg: Still SCRAM-SHA-256 never works, but we know SCRAM-SHA-1 was failing when tested to early, as the changes to the cluster user took some time to be applied after the call returned with success.

josvazg:
Lessons around contract testing:
- For us it requires quite some infrastructure automation to make sure the initial state is as expected. Usually a project and deployment need to be in place for most tests.
- For local dev it is useful to keep the infrastructure around to re-test faster. For CI it needs to be cleaned up always at the end.
- Ad-hoc infra automation is a waste, we probably want to use the Operator itself for this.
- Calling the API directly is pointless: we must test the production code that calls the API instead. This needs our production code to separate such code and probably include full translation of types as well in such layer.

PR #1441, opened 2024-03-14
Ensure tests checkout PR branch

josvazg:
If not explicit, on some PR or push events it might checkout main instead, which is not what we want on PR tests.

### All Submissions:

### DIFF
--- a/.github/actions/build-push-image/action.yaml
+++ b/.github/actions/build-push-image/action.yaml
@@ -76,7 +76,7 @@ runs:
         cache: false
     - name: Setup cache
       if: ${{ inputs.get-existing-cache == 'false' }}
-      uses: actions/cache@v3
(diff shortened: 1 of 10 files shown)


roothorp (approved):
Have you got this working on a fork somewhere?

josvazg:
> Have you got this working on a fork somewhere?

I tested here directly. I will run cloud tests so that we can double check it picked up the right PR branch to checkout also there.

josvazg:
> looks like it's still checking out main...

Seems when you re-label it does not work, I guess `${{ github.head_ref || github.ref }}` renders nothing so it goes to defaults.

I am going to add some traces.

josvazg:
Seems `ref: ${{github.event.pull_request.head.sha}}` is the answer everywhere, independent of the trigger event.

PR #1466, opened 2024-03-21
CLOUDP-228585: Migrate Project Alert Configuration to Atlas Go SDK

helderjs:
### All Submissions:

### DIFF
--- a/pkg/api/v1/status/alert_configurations.go
+++ b/pkg/api/v1/status/alert_configurations.go
@@ -5,7 +5,12 @@ import (
 	"fmt"
 	"strconv"
 
-	"go.mongodb.org/atlas/mongodbatlas"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/unstructured"
+
+	"go.mongodb.org/atlas-sdk/v20231115004/admin"
+	"go.uber.org/zap"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/timeutil"
 )
 
 type AlertConfiguration struct {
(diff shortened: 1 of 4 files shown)


pkg/api/v1/status/alert_configurations.go
```diff
@@ -195,53 +196,33 @@ type Matcher struct {
...
 
-	if alertConfiguration.Matchers != nil {
-		status.Matchers = make([]Matcher, 0, len(alertConfiguration.Matchers))
-		for _, matcher := range alertConfiguration.Matchers {
-			matcherFromAtlas := MatcherFromAtlas(matcher)
-			status.Matchers = append(status.Matchers, matcherFromAtlas)
+	if unstructuredMatchers, ok := alertConfiguration.GetMatchersOk(); ok {
+		matchers, err := unstructured.TypedFromUnstructured[[]map[string]interface{}, []Matcher](*unstructuredMatchers)
+		if err != nil {
+			logger.Errorf("unable to convert matchers to structured type: %s", err)
```
josvazg, line 212:
```suggestion
			logger.Errorf("unable to convert matchers to structured type: %v", err)
```

josvazg (approved):

pkg/controller/atlasproject/alert_configurations.go
```diff
@@ -5,11 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"go.mongodb.org/atlas/mongodbatlas"
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
```
josvazg, line 8: NOT for this PR...Food for though on following SDK migrations:

If we want to make the reconciliation logic more isolated from the SDK API so that:
- We can add contract tests easily later
- Or add reconciliation logic unit testing mocking just the SDK API part away.
Then the controller reconciler code file should NOT refer to the API module at all in the imports.

We do not need to separate the API calls into a different go package, a separate go file within the same package may suffice.

Note in this particular case, the wrapping is already almost done: The only thing we need the SDK import here for is because we are NOT properly isolating the `admin.GroupAlertsConfig` type. Ideally reconciliation should not handle API types here directly and instead use one on its own, even if it is later a 1:1 mapping to the API, but that is a concern that needs to be translation layer right before calling the API, not at reconciliation time.

Not this might seem a bit of a overkill now, but it pays off not only for contract testing or mocking unit tests in the right place, but also with subsequent API upgrades that will be more frequent with the new SDK. The code to migrate will never be mixed with the reconciliation logic and will have explicit type translations to revisit.

PR #1481, opened 2024-03-28
CLOUDP-239182: Ensure teams are always cleaned up

igor-karpukhin:
### All Submissions:

Ensure teams are always cleaned up

### DIFF
--- a/pkg/api/v1/status/team.go
+++ b/pkg/api/v1/status/team.go
@@ -23,22 +23,6 @@ func AtlasTeamSetProjects(projects []TeamProject) AtlasTeamStatusOption {
 	}
 }
 
-func AtlasTeamUnsetProject(projectID string) AtlasTeamStatusOption {
-	return func(s *TeamStatus) {
-		index := -1
-		for i := 0; i < len(s.Projects); i++ {
-			if s.Projects[i].ID == projectID {
-				index = i
-				break
-			}
-		}
-		if index == -1 {
-			return
-		}
-		s.Projects = append(s.Projects[:index], s.Projects[index+1:]...)
-	}
-}
-
 type TeamStatus struct {
 	Common `json:",inline"`
 
--- a/pkg/api/v1/status/team.go
+++ b/pkg/api/v1/status/team.go
@@ -23,6 +23,22 @@ func AtlasTeamSetProjects(projects []TeamProject) AtlasTeamStatusOption {
 	}
 }
 
+func AtlasTeamUnsetProject(projectID string) AtlasTeamStatusOption {
+	return func(s *TeamStatus) {
+		index := -1
+		for i := 0; i < len(s.Projects); i++ {
+			if s.Projects[i].ID == projectID {
+				index = i
+				break
+			}
+		}
+		if index == -1 {
+			return
+		}
+		s.Projects = append(s.Projects[:index], s.Projects[index+1:]...)
+	}
+}
+
 type TeamStatus struct {
 	Common `json:",inline"`
 
--- a/pkg/controller/atlasproject/atlasproject_controller.go
+++ b/pkg/controller/atlasproject/atlasproject_controller.go
@@ -256,9 +256,9 @@ func (r *AtlasProjectReconciler) handleDeletion(workflowCtx *workflow.Context, a
 				}
 
 				if err := r.garbageCollectTeams(workflowCtx.Context, project.ID()); err != nil {
-					result := workflow.Terminate(workflow.Internal, err.Error())
-					log.Errorw("failed to cleanup project teams", "error", err)
-					return result
+					e := fmt.Errorf("failed to cleanup project teams. error: %w", err)
+					log.Error(e)
+					return workflow.Terminate(workflow.Internal, e.Error())
 				}
 
 				if err := r.deleteAtlasProject(workflowCtx.Context, atlasClient, project); err != nil {
--- a/pkg/controller/atlasproject/atlasproject_controller.go
+++ b/pkg/controller/atlasproject/atlasproject_controller.go
@@ -255,10 +255,10 @@ func (r *AtlasProjectReconciler) handleDeletion(workflowCtx *workflow.Context, a
 					return result
 				}
 
-				if err := r.garbageCollectTeams(workflowCtx.Context, project.ID()); err != nil {
-					e := fmt.Errorf("failed to cleanup project teams. error: %w", err)
-					log.Error(e)
-					return workflow.Terminate(workflow.Internal, e.Error())
+				err := r.syncAssignedTeams(workflowCtx, project.ID(), project, map[string]*akov2.Team{})
+				if err != nil {
+					workflowCtx.SetConditionFalse(status.ProjectTeamsReadyType)
+					return workflow.Terminate(workflow.ProjectTeamUnavailable, err.Error())
 				}
 
 				if err := r.deleteAtlasProject(workflowCtx.Context, atlasClient, project); err != nil {
--- a/pkg/controller/atlasproject/atlasproject_controller.go
+++ b/pkg/controller/atlasproject/atlasproject_controller.go
@@ -255,7 +255,7 @@ func (r *AtlasProjectReconciler) handleDeletion(workflowCtx *workflow.Context, a
 					return result
 				}
 
-				err := r.syncAssignedTeams(workflowCtx, project.ID(), project, map[string]*akov2.Team{})
+				err := r.syncAssignedTeams(workflowCtx, project.ID(), project, nil)
 				if err != nil {
 					workflowCtx.SetConditionFalse(status.ProjectTeamsReadyType)
 					return workflow.Terminate(workflow.TeamNotCleaned, err.Error())
--- a/pkg/controller/atlasproject/atlasproject_controller.go
+++ b/pkg/controller/atlasproject/atlasproject_controller.go
@@ -258,7 +258,7 @@ func (r *AtlasProjectReconciler) handleDeletion(workflowCtx *workflow.Context, a
 				err := r.syncAssignedTeams(workflowCtx, project.ID(), project, map[string]*akov2.Team{})
 				if err != nil {
 					workflowCtx.SetConditionFalse(status.ProjectTeamsReadyType)
-					return workflow.Terminate(workflow.TeamCleanupFailed, err.Error())
+					return workflow.Terminate(workflow.TeamNotCleaned, err.Error())
 				}
 
 				if err := r.deleteAtlasProject(workflowCtx.Context, atlasClient, project); err != nil {
--- a/pkg/controller/atlasproject/atlasproject_controller.go
+++ b/pkg/controller/atlasproject/atlasproject_controller.go
@@ -258,7 +258,7 @@ func (r *AtlasProjectReconciler) handleDeletion(workflowCtx *workflow.Context, a
 				err := r.syncAssignedTeams(workflowCtx, project.ID(), project, map[string]*akov2.Team{})
 				if err != nil {
 					workflowCtx.SetConditionFalse(status.ProjectTeamsReadyType)
-					return workflow.Terminate(workflow.ProjectTeamUnavailable, err.Error())
+					return workflow.Terminate(workflow.TeamCleanupFailed, err.Error())
 				}
 
 				if err := r.deleteAtlasProject(workflowCtx.Context, atlasClient, project); err != nil {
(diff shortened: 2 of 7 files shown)


pkg/controller/atlasproject/teams.go
```diff
@@ -30,8 +31,56 @@ type TeamDataContainer struct {
...
+
+				if len(team.Status.Projects) == 0 &&
+					customresource.HaveFinalizer(&team, customresource.FinalizerLabel) {
+					customresource.UnsetFinalizer(&team, customresource.FinalizerLabel)
+				}
+
+				if err = r.Client.Update(ctx, &team); err != nil {
+					r.Log.Errorw("failed to update Team object", "error", err)
+					return err
+				}
```
helderjs, line 67:
```suggestion
				if len(team.Status.Projects) == 0 {
					return customresource.ManageFinalizer(ctx, r.Client, &team, customresource.UnsetFinalizer)
				}
```

pkg/controller/atlasproject/teams.go
```diff
@@ -30,8 +31,56 @@ type TeamDataContainer struct {
...
+	return g.Wait()
+}
+
 func (r *AtlasProjectReconciler) ensureAssignedTeams(workflowCtx *workflow.Context, project *akov2.AtlasProject, protected bool) workflow.Result {
 	resourcesToWatch := make([]watch.WatchedObject, 0, len(project.Spec.Teams))
+
+	err := r.garbageCollectTeams(workflowCtx.Context, project.ID())
+	if err != nil {
+		return workflow.Terminate(workflow.TeamNotCleaned, err.Error())
+	}
```
helderjs, line 82: Calling the garbage collector directly here will remove the project from team status even if it wasn't de-assigned.
While the next part of the code will add it back, this may have a workflow that leads to a new bug.
Here I'd call the garage collector only when `project.Spec.Teams` is empty
```suggestion
        if len(project.Spec.Teams) == 0 {
 	    err := r.garbageCollectTeams(workflowCtx.Context, project.ID())
	    if err != nil {
	 	    return workflow.Terminate(workflow.TeamNotCleaned, err.Error())
	    }
	}
```

helderjs (changes requested):

pkg/controller/atlasproject/teams.go
igor-karpukhin: What if only one team was removed? E.g. there are two teams, one got removed. In this case, it won't be garbage collected.

helderjs: I see... with that, we need to be sure a cleaned-up team gets its status corrected later or do a separate clean-up function for this case. If the current implementation already covers first mentioned option, I'm good

igor-karpukhin: Let me add one more test to make sure we cover this scenario as well

pkg/api/v1/status/team.go
```diff
@@ -23,6 +23,22 @@ func AtlasTeamSetProjects(projects []TeamProject) AtlasTeamStatusOption {
...
+		for i := 0; i < len(s.Projects); i++ {
+			if s.Projects[i].ID == projectID {
+				index = i
+				break
+			}
+		}
+		if index == -1 {
+			return
+		}
+		s.Projects = append(s.Projects[:index], s.Projects[index+1:]...)
```
josvazg, line 38:
```suggestion
		for i := 0; i < len(s.Projects); i++ {
			if s.Projects[i].ID == projectID {
				s.Projects = append(s.Projects[:i], s.Projects[i+1:]...)
				return
			}
		}
```

josvazg (approved):

pkg/api/v1/status/team.go
```diff
@@ -23,6 +23,17 @@ func AtlasTeamSetProjects(projects []TeamProject) AtlasTeamStatusOption {
 	}
 }
 
+func AtlasTeamUnsetProject(projectID string) AtlasTeamStatusOption {
+	return func(s *TeamStatus) {
+		for i := 0; i < len(s.Projects); i++ {
+			if s.Projects[i].ID == projectID {
+				s.Projects = append(s.Projects[:i], s.Projects[i+1:]...)
```
s-urbaniak, line 30: nit: add a small godoc that this is removing an element from `s.Project` at index `i` preserving order of the remaining elements. This golang idiom might not be obvious to a future reader.

pkg/controller/atlasproject/teams.go
```diff
@@ -30,8 +31,54 @@ type TeamDataContainer struct {
...
+					team.Status.ID = ""
+				}
+
+				if customresource.HaveFinalizer(&team, customresource.FinalizerLabel) {
+					customresource.UnsetFinalizer(&team, customresource.FinalizerLabel)
+				}
+
+				if err = r.Client.Status().Update(ctx, &team); err != nil {
+					r.Log.Errorw("failed to update Team status", "error", err)
+					return err
```
s-urbaniak, line 64:
```suggestion
					return fmt.Errorf("error updating team status: %w", err)
```

pkg/controller/atlasproject/teams.go
```diff
@@ -30,8 +31,54 @@ type TeamDataContainer struct {
...
+
+	return g.Wait()
+}
+
 func (r *AtlasProjectReconciler) ensureAssignedTeams(workflowCtx *workflow.Context, project *akov2.AtlasProject, protected bool) workflow.Result {
 	resourcesToWatch := make([]watch.WatchedObject, 0, len(project.Spec.Teams))
+
+	err := r.garbageCollectTeams(workflowCtx.Context, project.ID())
+	if err != nil {
+		return workflow.Terminate(workflow.TeamNotCleaned, err.Error())
```
s-urbaniak, line 79:
```suggestion
		return workflow.Terminate(workflow.TeamNotCleaned, fmt.Sprintf("error garbage collecting teams: %v", err))
```

pkg/controller/atlasproject/teams_test.go
```diff
@@ -438,3 +449,159 @@ func TestUpdateTeamState(t *testing.T) {
...
+		logger := zaptest.NewLogger(t).Sugar()
+		workflowCtx := &workflow.Context{
+			Context: context.Background(),
+			Log:     logger,
+		}
+		testScheme := runtime.NewScheme()
+		testScheme.AddKnownTypes(akov2.GroupVersion, &akov2.AtlasProject{})
+		testScheme.AddKnownTypes(akov2.GroupVersion, &akov2.AtlasTeam{})
+		testScheme.AddKnownTypes(akov2.GroupVersion, &akov2.AtlasTeamList{})
+		testScheme.AddKnownTypes(akov2.GroupVersion, &corev1.Secret{})
```
s-urbaniak, line 464: This whole block can be replaced with:
```suggestion
		akov2.AddToScheme(testScheme)
		corev1.AddToScheme(testScheme)
```

pkg/controller/atlasproject/teams_test.go
```diff
@@ -438,3 +449,159 @@ func TestUpdateTeamState(t *testing.T) {
...
```
s-urbaniak, line 543: see above

s-urbaniak (changes requested):
Looks great, just a couple of comments 👍

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -255,6 +255,12 @@ func (r *AtlasProjectReconciler) ensureDeletionFinalizer(workflowCtx *workflow.C
 					return result
 				}
 
+				if err := r.garbageCollectTeams(workflowCtx.Context, project.ID()); err != nil {
+					e := fmt.Errorf("failed to cleanup project teams. error: %w", err)
+					log.Error(e)
+					return workflow.Terminate(workflow.Internal, e.Error())
+				}
+
```
helderjs, line 263:
```suggestion
				if err := r.syncAssignedTeams(workflowCtx, project.ID(), project, map[string]*akov2.Team{}); err != nil {
		                        return workflow.Terminate(workflow.TeamCleanupFailed, err.Error())
	                       }

```

pkg/controller/atlasproject/teams.go
```diff
@@ -30,8 +33,77 @@ type TeamDataContainer struct {
 	Context     *workflow.Context
 }
 
+func (r *AtlasProjectReconciler) garbageCollectTeams(ctx context.Context, projectID string) error {
```
helderjs, line 36: I was analyzing this case and the only issue in syncing teams is when a project is deleted before the team gets managed beforehand. Given that, we don't need a "garbage collector" but sync teams (de-assign them) and update their status.
Finally, I believe we only need my code suggestion in the previous comment

helderjs (changes requested):

igor-karpukhin:
@helderjs please take a look when you have time. I moved changed the logic as we agreed

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -255,6 +255,12 @@ func (r *AtlasProjectReconciler) ensureDeletionFinalizer(workflowCtx *workflow.C
 					return result
 				}
 
+				err := r.syncAssignedTeams(workflowCtx, project.ID(), project, map[string]*akov2.Team{})
+				if err != nil {
+					workflowCtx.SetConditionFalse(status.ProjectTeamsReadyType)
+					return workflow.Terminate(workflow.ProjectTeamUnavailable, err.Error())
```
helderjs, line 261:
```suggestion
					return workflow.Terminate(workflow.TeamCleanupFailed, err.Error())
```

helderjs (approved):
LGTM, just a last comment

s-urbaniak (approved):
LGTM given @helderjs's comments are addressed 👍

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -255,6 +255,12 @@ func (r *AtlasProjectReconciler) ensureDeletionFinalizer(workflowCtx *workflow.C
 					return result
 				}
 
+				err := r.syncAssignedTeams(workflowCtx, project.ID(), project, map[string]*akov2.Team{})
```
s-urbaniak, line 258:
```suggestion
				err := r.syncAssignedTeams(workflowCtx, project.ID(), project, nil)
```

s-urbaniak (changes requested):
One last nit and a missing unit test.

s-urbaniak:
@igor-karpukhin This contribution adds a unit test for `updateTeamState`, however `handleDeletion` in `AtlasProjectReconciler` is not tested at all. Please add a unit test covering the finalizer case which invokes the added `syncAssignedTeams` handling.

pkg/controller/atlasproject/teams_test.go
```diff
@@ -437,4 +445,78 @@ func TestUpdateTeamState(t *testing.T) {
...
+			Client:        k8sClient,
+			Log:           logger,
+			AtlasProvider: atlasProvider,
+		}
+		teamRef := &common.ResourceRefNamespaced{
+			Name:      team.Name,
+			Namespace: "testNS",
+		}
+
+		err := reconciler.updateTeamState(workflowCtx, project, teamRef, true)
```
s-urbaniak, line 517: I am confused, this tests `updateTeamState ` while no changes have been done in this function in this contribution in the first place.

igor-karpukhin: Yes, but if verifies if it actually removes the team if it is unassigned from all AtlasProjects. See the new unit-test added.

pkg/controller/atlasproject/atlasproject_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+		workflowCtx := &workflow.Context{
+			Client:  &atlasClient,
+			Context: context.Background(),
+			SdkClient: &admin.APIClient{
+				PrivateEndpointServicesApi: mockPrivateEndpointAPI,
+				NetworkPeeringApi:          mockPeeringEndpointAPI,
+			},
+			Log: zap.S(),
+		}
+		_ = reconciler.handleDeletion(workflowCtx, workflowCtx.Client, testProject)
```
s-urbaniak, line 155: please assert the returned `result` as well. Looking at the coverage profile It looks like the zero value is being returned.

s-urbaniak (changes requested):
@igor-karpukhin please assert the returned result from `handleDeletion`. Looking at the coverage profile, the zero value is returned and not an explicit `OK()`.

pkg/controller/atlasproject/atlasproject_controller_test.go
igor-karpukhin: Fixed

PR #1487, opened 2024-04-02
[PoC] Launch test YAMLs directly

josvazg:
PoC  not to merge but to be used for API testing.

### All Submissions:

### DIFF
--- a/test/helper/launcher/launcher.go
+++ b/test/helper/launcher/launcher.go
@@ -1,0 +1,210 @@
+package launcher
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"strings"
+)
+
+const (
+	ExpectedContext = "kind-kind"
+
+	LauncherTestInstall = "test-ako"
+
+	HelmRepoURL = "https://mongodb.github.io/helm-chart"
+
+	RepoRef = "mongodb"
+
+	OperatorChart = "mongodb-atlas-operator"
+
+	AtlasURI = "https://cloud-qa.mongodb.com"
+)
+
+type Launcher struct {
+	credentials   AtlasCredentials
+	version       string
+	cmdOutput     bool
+	clearKind     bool
+	clearOperator bool
+}
+
+// NewLauncher creates a new operator Launcher
+func NewLauncher(creds AtlasCredentials, version string, cmdOutput bool) *Launcher {
+	return &Launcher{credentials: creds, version: version, cmdOutput: cmdOutput}
+}
+
+// NewFromEnv creates an operator Launcher using defaults and credentials from env vars
+func NewFromEnv(version string) *Launcher {
+	return NewLauncher(credentialsFromEnv(), version, true)
+}
+
+// MustSetEnv sets the env var value given, or panics if the env var is not set
+func MustSetEnv(envvar string) string {
+	value, ok := os.LookupEnv(envvar)
+	if !ok {
+		panic(fmt.Errorf("environment variable %s not set", envvar))
+	}
+	return value
+}
+
+// Launch will try to launch the operator and apply the given YAML for it to handle
+func (l *Launcher) Launch(yml string, waitCfg *WaitConfig) error {
+	if err := l.ensureK8sCluster(); err != nil {
+		return err
+	}
+	if err := l.ensureOperator(); err != nil {
+		return err
+	}
+	secretYml := l.credentials.secretYML()
+	if err := l.kubeApply(secretYml); err != nil {
+		return err
+	}
+	if err := l.kubeApply(yml); err != nil {
+		return err
+	}
+	return l.kubeWait(waitCfg)
+}
+
+// Cleanup related Launcher test resources, merely wipe the kind cluster
+func (l *Launcher) Cleanup() error {
+	if l.clearOperator {
+		if err := l.uninstall(); err != nil {
+			return err
+		}
+	}
+	if l.clearKind {
+		return l.stopKind()
+	}
+	return nil
+}
+
+func (l *Launcher) ensureK8sCluster() error {
+	if !l.isKubeConfigAvailable() {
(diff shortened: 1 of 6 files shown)


test/helper/launcher/launcher.go
```diff
@@ -0,0 +1,210 @@
...
+	if !ok {
+		panic(fmt.Errorf("environment variable %s not set", envvar))
+	}
+	return value
+}
+
+// Launch will try to launch the operator and apply the given YAML for it to handle
+func (l *Launcher) Launch(yml string, waitCfg *WaitConfig) error {
+	if err := l.ensureK8sCluster(); err != nil {
+		return err
```
s-urbaniak, line 57: wrap all errors below. In this case:
```suggestion
		return fmt.Errorf("error ensuring k8s cluster: %w", err)
```

s-urbaniak:
See https://kuttl.dev/docs/kuttl-test-harness.html#writing-your-first-test and https://kuttl.dev/docs/api-integration.html#test-harness.

josvazg:
> See https://kuttl.dev/docs/kuttl-test-harness.html#writing-your-first-test and https://kuttl.dev/docs/api-integration.html#test-harness.

This is not a testing framework, it is a simple testing helper. It could be used as a framework maybe, but it is NOT the intent here and for that I agree that KUTTL is a better option.

Think of this as the function I call in Go to get the Atlas infrastructure in the state I need to run a contract test, which is defined in Go code, not in YAML. A test that calls our translation layer that talks to the API using the same code that the reconcile logic uses. If KUTTL could be used in the same unobtrusive way, then maybe we could use it for all, I am not 100% convinced yet.

Sample custom test:

```go
  l := launcher.NewFromEnv(testVersion) // instantiate a launcher
  l.Launch(testYml, launcher.WaitReady("atlasprojects/my-project", 30*time.Second))) // launch the YAML and wait for it
  
  // run custom test code - might not be part of KUTTLE, like a custom go code driven contract test

   l.Cleanup()) // optionally cleanup the infrastructure required for the test (does KUTTL allow for optional cleanup?)
```

s-urbaniak:
@josvazg I am leaving the final assertion to you 👍 I didn't have a deeper look at kuttl, just worried about NIH implementations. Again, final call is yours :-)

test/helper/launcher/wait.go
```diff
@@ -0,0 +1,28 @@
...
+	Timeout   time.Duration
+}
+
+func WaitReady(target string, timeout time.Duration) *WaitConfig {
+	return &WaitConfig{Condition: "condition=Ready", Target: target, Timeout: timeout}
+}
+
+func (cfg *WaitConfig) waitArgs() []string {
+	args := []string{"wait", fmt.Sprintf("--for=%s", cfg.Condition), cfg.Target}
+	if cfg.Timeout == 0 {
```
helderjs, line 24: It's a mistake or do you really want to set timeout only when it's zero?

josvazg:
> @josvazg I am leaving the final assertion to you 👍 I didn't have a deeper look at kuttl, just worried about NIH implementations. Again, final call is yours :-)

I think the best way forward now is to actually force myself to use KUTTL for some real Contract tests, for Atlas Search. If I see no blockers, then we can probably use for everything and forget about this PoC for good.

helderjs (changes requested):

helderjs:
@josvazg This PoC links well to my comment on the CI Revamp document regarding GitHub Actions.

Why didn't you write a Github action workflow for it? I'd say we probably won't need any code for it.
I see your code doing what Github Action should be doing, and I believe that's why @s-urbaniak also saw it as a framework wrapping the tooling needed to achieve the goal. Code is a liability and I prefer to have no code when we already have a toolchain in place to do the job. In the end, you would still use a workflow to trigger this task.
Think that a CI tool like GH action offers us out-of-box multi-stage pipeline and jobs, some level of security, lifecycle control... keep in mind that once you have signaled you will drop this to a new PoC with KUTTL

josvazg:
> @josvazg This PoC links well to my comment on the CI Revamp document regarding GitHub Actions.
> 
> Why didn't you write a Github action workflow for it? I'd say we probably won't need any code for it. I see your code doing what Github Action should be doing, and I believe that's why @s-urbaniak also saw it as a framework wrapping the tooling needed to achieve the goal. Code is a liability and I prefer to have no code when we already have a toolchain in place to do the job. In the end, you would still use a workflow to trigger this task. Think that a CI tool like GH action offers us out-of-box multi-stage pipeline and jobs, some level of security, lifecycle control... keep in mind that once you have signaled you will drop this to a new PoC with KUTTL

I want us to be able to do our development without any Github involvement, using the github CI actions only to automatically re-run tests at the proper time to ensure what we release is correct and did not include any regressions.

The idea is to be able to write contract tests in Go that look like this:

1. Get a particular set of Atlas resources in a particular state.
1. Run one or more contract tests against the Atlas API.
1. Tear down all resources required for the test.

During development, in your laptop and before any code reaches github, you iterate with steps 1-2 until you test all APIs you needed to verify. Then do step 3. It is only later on you send a PR and then the CI might also want to run those tests at some point.

This might be doable with KUTTL, because it seems to have a programmatic interface. I will need to test it.
But as a pure github action or workflow it would be completely useless for contract testing.

josvazg:
Going to close this PoC now:
- It was never the intention to merge this at this point.
- It was needed for me as a way to run contract tests, but haven't gotten there yet.
- Seems KUTTL has a programmatic interface that might allow it to be used within tests as a helper, in which case this code will not be needed.
- Should KUTTL prove inadequate later, we can always dig up this Pr again.
- The main goal of the PoC is achieved, launching declarative YAML infrastructure for a test OR as a helper to setup before a test is not only doable but also relatively simple.

test/helper/launcher/wait.go
josvazg: This was a mistake, the logic had to be reversed

PR #1491, opened 2024-04-04
CLOUDP-235981: Mark certain fields as deprecated for ServerlessSpec

igor-karpukhin:
### All Submissions:

This PR marks the following `providerSettings` fields as deprecated:
* DiskIOPS
* DiskTypeName
* EncryptEBSVolume
* InstanceSizeName
* VolumeType
* AutoScaling


josvazg:
My two cents on how we should deprecate this kind of things:

- **Step 1: Mark fields deprecated**, as this PR does.
- **Step 2: Introduce flag to fail validation** when deprecated fields are used BUT disabled by default.
- **Step 3: Turn on flag by default*. Backward compatibility is still possible, but we have given users more than enough time to migrate away from those fields.

WDYT?

igor-karpukhin:
> My two cents on how we should deprecate this kind of things:
> 
> * **Step 1: Mark fields deprecated**, as this PR does.
> * **Step 2: Introduce flag to fail validation** when deprecated fields are used BUT disabled by default.
> * **Step 3: Turn on flag by default**. Backward compatibility is still possible, but we have given users more than enough time to migrate away from those fields.
> 
> Each step is done in a subsequent release and is properly highlighted in the release notes.
> 
> WDYT?

I'm not sure if we have a deprecation policy written somewhere. I'd say we need to talk to @s-urbaniak when he is back.

PR #1492, opened 2024-04-05
CLOUDP-241390: Generate upload SBOM script

josvazg:
Add SBOM generation and S3 push script for future use.

### All Submissions:

### DIFF
--- a/scripts/generate_upload_sbom.sh
+++ b/scripts/generate_upload_sbom.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -Eeou pipefail
 
+platforms=("linux/arm64" "linux/amd64")
 image_pull_spec=""
 image_name=""
 tag_name=""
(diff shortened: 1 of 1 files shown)


slaskawi (commented):
LGTM so far! Note that this script doesn't create the top bucket (you need to do it manually). This is intentional just in case CI goes crazy, it's easier to delete everything. 

But I'm happy to revisit this decision if needed.

josvazg:
> LGTM so far! Note that this script doesn't create the top bucket (you need to do it manually). This is intentional just in case CI goes crazy, it's easier to delete everything.
> 
> But I'm happy to revisit this decision if needed.

Yes, I did create a bucket myself, then removed as I did not lock down its permissions to set its permissions ¿?... so I decided to remove the bucket. But the code seemed to work on my simple manual test.

josvazg:
In fact @slaskawi, I believe the bucket should always be a given. I would expect that it will always exist in advance and we might need to add credentials in the CI to access it properly.

s-urbaniak (approved):

slaskawi (approved):

PR #1497, opened 2024-04-08
CLOUDP-238747: Fix CI release flow

josvazg:
This fixes the automated release workflow AFTER we merge the **Release Branch** by splitting the workflow in 2 parts:

1. First, the PR closing triggers a workflow that tags the release appropriately: `release/X.Y.Z` becomes tag `vX.Y.Z` or `pre-release/X.Y.Z-...` becomes `vX.Y.Z-...`
1.  Then, that same workflow triggers the release workflow to use the tag and its code as the version to release.

The release workflow can also be triggered by posting the tag or triggering such workflow manually.

NO changes to the release documentation steps are required, other than to clarify how this automation works.

### All Submissions:


.github/workflows/release-post-merge.yml
```diff
@@ -5,20 +5,20 @@
...
-  pull_request:
-    branches:
-      - pre-release/**
-      - release/**
-    types:
-      - closed
-  workflow_dispatch:
+  push:
+    tags:
+      - 'v*'
```
s-urbaniak, line 10: i guess a regex would be too much here? 😬 just worried this will catch a tag like `very-nice-tag`.

s-urbaniak (approved):
one non-blocking question 👍

.github/workflows/release-post-merge.yml
josvazg: I am not sure that danger avails for the cost. Regex is all but standard, and a proper semver regexp is costly to test. I do not even trust the regex by **act** would be the same as the actual regex used by GH CI actions. If we have a regex that is good enough and is known to work with GitHub CI I would not mind.

helderjs (approved):

.github/workflows/release-post-merge.yml
s-urbaniak: good points

PR #1519, opened 2024-04-17
CLOUDP-243996: Fix GCP privatelink e2e IP conflict flakiness

josvazg:
Change the method to get a GCP IP address into one that retries up to 7 times if the IP address selected randomly is already in use.

Added a test that verified the error message was coming from `createVirtualAddress`. Also added partial context passing to the GCP method.

### All Submissions:

### DIFF
--- a/test/helper/e2e/actions/cloud/provider.go
+++ b/test/helper/e2e/actions/cloud/provider.go
@@ -280,15 +280,16 @@ func (a *ProviderAction) SetupPrivateEndpoint(request PrivateEndpointRequest) *P
 }
 
 func (a *ProviderAction) RetryGCPCreateEndpoint(maxRetries int, name, region, subnet, target string, index int) (string, string, error) {
-	retries := maxRetries
+	retries := 0
 	for {
 		rule, ip, err := a.gcpProvider.CreatePrivateEndpoint(name, region, subnet, target, index)
-		if retries > 0 && err != nil {
+		if retries < maxRetries && err != nil {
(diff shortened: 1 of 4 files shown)


test/helper/e2e/actions/cloud/provider.go
```diff
@@ -276,6 +279,23 @@ func (a *ProviderAction) SetupPrivateEndpoint(request PrivateEndpointRequest) *P
 	return nil
 }
 
+func (a *ProviderAction) RetryGCPCreateEndpoint(maxRetries int, name, region, subnet, target string, index int) (string, string, error) {
```
s-urbaniak, line 282: this is too naive, please use exponential backoff retries, see https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/wait/backoff.go#L477

josvazg: Will do. But please note this is test helper code, quite far from production requirements.

s-urbaniak: This is not about production requirements.

Remote APIs (in this case GCP) cannot be retried in a brute-force manner in endless loops. You will most likely hit a rate-limiter on the GCP side.

Hence, retryring things on remote APIs needs to have:
- jitter
- delay between retries
- exponential backoff between retries

This is especially important for test code.

s-urbaniak (approved):

PR #1527, opened 2024-04-22
CLOUDP-242965: Name matrix compute step

josvazg:
Fix the computation of the test matrix for **scheduled** events.

Looking at [this example from today's schedule](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/8776607163/job/24080359511):

```
    GH_REF: refs/heads/main
    GH_HEAD_REF: 
```

The value `github.head_ref` (**GH_HEAD_REF**) is not set on a scheduled event, but `github.ref` (**GH_REF**) is set as `refs/heads/main`. This change takes that into account to fix the comparison to be triggered correctly on scheduled test runs.

### All Submissions:

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -20,6 +20,7 @@ jobs:
       test_matrix: ${{ steps.test.outputs.matrix }}
     steps:
       - id: test
+        name: Compute Test Matrix
         run: |
           # Note the use of external single quotes to allow for doble quotes at inline YAML array
           matrix='["v1.27.1-kind"]'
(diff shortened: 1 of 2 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -15,15 +15,16 @@ jobs:
     name: "Compute test matrix"
     runs-on: ubuntu-latest
     env:
-      GH_HEAD_REF: ${{ github.head_ref }}
+      GH_REF: ${{ github.ref }}
```
helderjs, line 18: Shouldn't be better to set this env var in the context of the step? This will prevent someone use it as generally available value for any GH event

helderjs (approved):
LGTM, just a suggestion

.github/workflows/test-e2e.yml
josvazg: I simplified even further, moving the expansion to the script directly

PR #1529, opened 2024-04-22
CLOUDP-238527: Atlas streams instance controller

helderjs:
### All Submissions:

An integration test will be added in an upcoming PR, the connection controller needs to be finished first as the controllers communicate with each other.


pkg/controller/atlasstream/atlasstream_connection.go
```diff
@@ -0,0 +1,256 @@
...
+
+		data[key] = string(val)
+	}
+
+	return data, nil
+}
+
+func hasStreamConnectionChanged(
+	streamConnection *akov2.AtlasStreamConnection,
+	atlasStreamConnection admin.StreamsConnection,
```
s-urbaniak, line 242: let's make this a pointer, there is no need for a shallow copy here.

pkg/controller/atlasstream/atlasstream_connection.go
```diff
@@ -0,0 +1,256 @@
...
+	atlasStreamConnection.Links = nil
+	connection, err := mapper(streamConnection)
+	if err != nil {
+		return false, err
+	}
+	if _, ok := connection.GetAuthenticationOk(); ok {
+		connection.Authentication.Password = nil
+	}
+
+	return !reflect.DeepEqual(*connection, atlasStreamConnection), nil
```
s-urbaniak, line 255: We cannot do this, this will fail on nested slices in the `admin.StreamsConnection` type, i.e. `.links`, `.authentication.links`, and `.security.links`. We have two options:

1. clear (i.e. set to `nil`) any non-comparable fields in `atlasStreamConnection`

or

2. use normalization, where:
- we convert `atlasStreamConnection` to `*akov2.AtlasStreamConnection`
- use normalization from https://github.com/mongodb/mongodb-atlas-kubernetes/tree/semanticequals
- `reflect.DeepEqual` our own `*akov2.AtlasStreamConnection` types

Option 2. is preferred but not a showstopper for this PR.

In any case this is brittle as it relies on non-slices in embedded fields in `admin.StreamsConnection` so we have to make it very clear in the godoc of this function that we rely on the invariant of no slices being embedded in this type.

pkg/controller/atlasstream/atlasstream_instance_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		Named("AtlasStreamInstance").
+		For(&akov2.AtlasStreamInstance{}, builder.WithPredicates(r.GlobalPredicates...)).
+		Watches(
+			&akov2.AtlasStreamConnection{},
+			handler.EnqueueRequestsFromMapFunc(r.findStreamInstancesForStreamConnection),
+		).
+		Complete(r)
+}
+
+func (r *AtlasStreamsInstanceReconciler) findStreamInstancesForStreamConnection(_ context.Context, obj client.Object) []reconcile.Request {
```
s-urbaniak, line 142: This function is not covered with unit tests.

pkg/controller/atlasstream/atlasstream_instance_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	streamConnection, ok := obj.(*akov2.AtlasStreamConnection)
+	if !ok {
+		r.Log.Warnf("watching AtlasStreamConnection but got %t", obj)
```
s-urbaniak, line 145:
```suggestion
		r.Log.Warnf("watching AtlasStreamConnection but got %T", obj)
```

Note lowercase `%t` is for bools

pkg/controller/atlasstream/atlasstream_connection.go
```diff
@@ -0,0 +1,256 @@
...
+
+	if len(ops.Delete) > 0 {
+		// if there are connection to be deleted from the instance
+		err = deleteConnections(ctx, project, akoStreamInstance, ops.Delete)
+		if err != nil {
+			return r.terminate(ctx, workflow.StreamConnectionNotRemoved, err)
+		}
+	}
+
+	for i := range ops.NoOp {
```
s-urbaniak, line 90: this noop block is not handled by unit tests.

s-urbaniak (changes requested):
This is looking great 🎉 

Generally: We're missing a streams connection controller handling the finalizer of `AtlasStreamConnection` custom resources.

pkg/controller/atlasstream/atlasstream_connection.go
```diff
@@ -0,0 +1,256 @@
...
+) *streamConnectionOperations {
+	return &streamConnectionOperations{
+		NoOp:   make([]*akov2.AtlasStreamConnection, 0, len(akoStreamInstance.Spec.ConnectionRegistry)),
+		Create: make([]*akov2.AtlasStreamConnection, 0, len(akoStreamInstance.Spec.ConnectionRegistry)),
+		Update: make([]*akov2.AtlasStreamConnection, 0, len(akoStreamInstance.Spec.ConnectionRegistry)),
+		Delete: make([]*admin.StreamsConnection, 0, len(atlasStreamConnections)),
+	}
+}
+
+// this is the dispatcher of connection registry management
```
josvazg, line 46: Follow Go doc conventions

```suggestion
// handledConnectionRegistry is the dispatcher of connection registry management
```

pkg/controller/atlasstream/atlasstream_connection.go
```diff
@@ -0,0 +1,256 @@
+package atlasstream
```
josvazg, line 1: Cannot see the main reconciler entry point nor the state machine definition or code.

pkg/controller/atlasstream/atlasstream_instance_controller.go
```diff
@@ -0,0 +1,160 @@
...
+// +kubebuilder:rbac:groups=atlas.mongodb.com,namespace=default,resources=atlasstreamconnections,verbs=get;list
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+
+// https://dreampuf.github.io/GraphvizOnline/#digraph%20G%20%7B%0A%20%20%20%20subgraph%20cluster_pending%20%7B%0A%20%20%20%20%20%20%20%20skipped%3B%0A%20%20%20%20%20%20%20%20invalid%3B%0A%20%20%20%20%20%20%20%20unsupported%3B%0A%20%20%20%20%20%20%20%20terminated%3B%0A%20%20%20%20%20%20%20%20label%20%3D%20%22pending%22%3B%0A%20%20%20%20%7D%0A%0A%20%20%20%20deleted%20%5Blabel%3D%22deleted%5Cnfinalizer%20unset%22%5D%0A%0A%20%20%20%20pending%20-%3E%20pending%20%5Blabel%3D%22skip%5Cninvalidate%5Cnunsupport%5Cnterminate%22%5D%0A%20%20%20%20pending%20-%3E%20ready%20%5Blabel%3D%22create%22%5D%0A%20%20%20%20pending%20-%3E%20deleted%20%5Blabel%3D%22delete%22%5D%0A%20%20%20%20ready%20-%3E%20ready%20%5Blabel%3D%22update%22%5D%0A%20%20%20%20ready%20-%3E%20deleted%20%5Blabel%3D%22delete%22%5D%0A%20%20%20%20ready%20-%3E%20pending%20%5Blabel%3D%22terminate%22%5D%0A%7D%0A
```
s-urbaniak, line 47: please add the rendered version of this state machine to the spec: https://docs.google.com/document/d/1lJumw_SSt9BylLDO_YFytl99qKXbb_9fIxBm9kNftYQ/edit?usp=drivesdk

pkg/controller/atlasstream/atlasstream_connection_test.go
```diff
@@ -0,0 +1,1093 @@
...
+				Name:      "my-project",
+				Namespace: "default",
+			},
+			Spec: akov2.AtlasProjectSpec{
+				Name: "my-project",
+			},
+			Status: status.AtlasProjectStatus{
+				ID: "my-project-id",
+			},
+		}
```
josvazg, line 183: This tests with pages of setup code are hard to read and many times the setup is repetitive. This project is exactly the same as above right? so why not just this?

```suggestion
		project := defaultTestAtlasProject()
```

pkg/controller/atlasstream/atlasstream_instance_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	akoStreamInstance := akov2.AtlasStreamInstance{}
+	result := customresource.PrepareResource(ctx, r.Client, req, &akoStreamInstance, log)
+	if !result.IsOk() {
+		return result.ReconcileResult(), nil
+	}
+
+	return r.ensureAtlasStreamsInstance(ctx, log, &akoStreamInstance)
+}
+
+// this is the central state dispatcher
```
josvazg, line 62: Does our linter not complain if doc strings are not conformant to conventions?

```suggestion
// ensureAtlasStreamsInstance is the central state dispatcher
```

pkg/controller/atlasstream/atlasstream_instance_controller.go
```diff
@@ -0,0 +1,160 @@
...
+func (r *AtlasStreamsInstanceReconciler) ensureAtlasStreamsInstance(ctx context.Context, log *zap.SugaredLogger, akoStreamInstance *akov2.AtlasStreamInstance) (ctrl.Result, error) {
```
josvazg, line 63: There might be some kind of state machine, but seems hidden, at least to me, it is not clear what the states are nor the transitions. I would rather have it cleared in code than on a comment, but I would expect at least one of the two.

pkg/controller/atlasstream/atlasstream_instance_controller.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (r *AtlasStreamsInstanceReconciler) findStreamInstancesForStreamConnection(_ context.Context, obj client.Object) []reconcile.Request {
+	streamConnection, ok := obj.(*akov2.AtlasStreamConnection)
+	if !ok {
+		r.Log.Warnf("watching AtlasStreamConnection but got %t", obj)
+		return nil
+	}
+
+	requests := make([]reconcile.Request, 0, len(streamConnection.Status.Instances))
```
s-urbaniak, line 149: hmmm, this looks a bit brittle. not a blocker but the alternative is:

- list all streams instances <- expensive (optimization is possible, we could build an **index** of stream instances having the atlas streams connections refs as keys)
- iterate over them
- check if the given instance references `streamConnection`
- if so, add it to the queue

s-urbaniak: and the finder method would look like so:
```
func (r *InstanceReconciler) findStreamInstancesForAtlasStreamConnection(ctx context.Context, atlasStreamConnection client.Object) []reconcile.Request {
	atlasStreamInstances := &akov2.AtlasStreamInstanceList{}
	listOps := &client.ListOptions{
		FieldSelector: fields.OneTermEqualSelector(".spec.connectionRegistry", client.ObjectKeyFromObject(atlasStreamConnection).String()),
		Namespace:     atlasStreamConnection.GetNamespace(),
	}

	err := r.Client.List(ctx, atlasStreamInstances, listOps)
	if err != nil {
		return []reconcile.Request{}
	}

	requests := make([]reconcile.Request, len(atlasStreamInstances.Items))
	for i, item := range atlasStreamInstances.Items {
		requests[i] = reconcile.Request{
			NamespacedName: types.NamespacedName{
				Name:      item.GetName(),
				Namespace: item.GetNamespace(),
			},
		}
	}
	return requests
}
```

pkg/controller/atlasstream/atlasstream_instance_controller.go
s-urbaniak: see https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1529#discussion_r1575740696

pkg/controller/atlasstream/atlasstream_instance_controller.go
josvazg: I vote for using the indexes given by k8s like explained here.

pkg/api/v1/status/atlasstreaminstance.go
```diff
@@ -0,0 +1,56 @@
...
+type AtlasStreamInstanceStatusOption func(s *AtlasStreamInstanceStatus)
+
+func AtlasStreamInstanceDetails(ID string, hostnames []string) AtlasStreamInstanceStatusOption {
+	return func(s *AtlasStreamInstanceStatus) {
+		s.ID = ID
+		s.Hostnames = hostnames
+	}
+}
+
+func AtlasStreamInstanceAddConnection(name string, ref common.ResourceRefNamespaced) AtlasStreamInstanceStatusOption {
```
s-urbaniak, line 33: I am worried that this will lead to status overload. Imagine we have thousands of stream instances pointing to a connection, this will imply thousands of status entries.

How about we just write a summary, i.e. "this connection is referenced by 1000 instances" ?

pkg/controller/atlasstream/atlasstream_instance_controller.go
s-urbaniak: the danger zone here is that we fully rely on a status field maintaining all streams instances. I am worried about a) the amount of entries here (see comment above) and b) potential subtle bugs in maintaining that list.

pkg/controller/atlasstream/atlasstream_connection.go
helderjs: As I mutate the struct in the function I believe is safer to pass a copy to avoid side effects (even though right now I only unset a no relevant field `Links`)

pkg/controller/atlasstream/atlasstream_instance_controller.go
helderjs: There is a danger zone here too. Given the direction of the relationship, the instance holds the reference for the instance which might not be in the same namespace. For proper handling, we should list across all namespaces. I'm not sure how expensive this is and if it would still be a good replacement for current strategy.
wdty @s-urbaniak?

pkg/api/v1/status/atlasstreaminstance.go
helderjs: This is the status of the instance, while it would also be possible to have 1000 connections referred by an instance, seems unmanageable from the perspective of the processor.
Anyway, the reason to have it here is that in the event of failure of Atlas API to sync a resource the Status should reflect what is actually in Atlas, i.e., if a connection is removed from spec and Atlas API fails to remove it, it will still be in the status, reflecting this connection is still part of the instance.

pkg/controller/atlasstream/atlasstream_connection_test.go
helderjs: I'm passing this for now once we are going to discuss this topic externally

pkg/controller/atlasstream/atlasstream_connection.go
s-urbaniak: the struct has nested pointers so shallow-copying is having no effect. there is no point in passing a value here.

pkg/controller/atlasstream/atlasstream_instance_controller.go
s-urbaniak: @helderjs you are using index lookups, what is the concrete concern?

pkg/controller/atlasstream/atlasstream_connection.go
s-urbaniak: if you want to be safe, you need to do a deep copy of `admin.StreamsConnection` which I believe the SDK has no methods for.

pkg/api/v1/status/atlasstreaminstance.go
s-urbaniak: oh i see, it's the reverse, nevermind then.

pkg/controller/atlasstream/atlasstream_instance_controller.go
s-urbaniak: @helderjs I think there is confusion with how this works:

If you use `GetClient` when [initializing the controller](https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1529/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633R228)  you'll get a client that does reads from the local cache that is filled by your indexer function.

If you want to override the local cache and do direct (expensive as you say) lookups towards API Server, you would use [`GetAPIReader()`](https://github.com/kubernetes-sigs/controller-runtime/blob/b74908fd50f01704785762e94e79484d35ef6814/pkg/cluster/cluster.go#L71) which indeed is a very expensive call.

But we always use a cached client here.

s-urbaniak: See https://github.com/kubernetes-sigs/controller-runtime/blob/b74908fd50f01704785762e94e79484d35ef6814/pkg/client/client.go#L378

s-urbaniak: And since you're using indexed fields this will yield objects from the local cache: https://github.com/kubernetes-sigs/controller-runtime/blob/b74908fd50f01704785762e94e79484d35ef6814/pkg/cache/internal/cache_reader.go#L128

pkg/controller/atlasstream/atlasstream_instance_controller.go
helderjs: Added to spc doc

pkg/controller/atlasstream/atlasstream_instance_controller.go
helderjs: I have added the state machine graph to the spec doc as suggested by Serge

cmd/manager/main.go
```diff
@@ -136,6 +138,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx := context.TODO()
```
josvazg, line 141: Why TODO instead of Background()?

This should be wired to Ctrl+C so that it tries to gracefully finish the program right?
Not for this PR, but if we are to leave a TODO maybe we could leave a comment with the intended follow up:
https://medium.com/@matryer/make-ctrl-c-cancel-the-context-context-bd006a8ad6ff

BTW I think we should really move these code into `pkg/` so that the manager is ONLY ever configured once, not 4 times (here and then again for e2e tests and twice for int tests)

@s-urbaniak could we get that prioritised soon? ^ I do not think is so big it needs a quaterly planned thing.
1-2 days should be more than enough

pkg/controller/atlasstream/atlasstream_connection.go
```diff
@@ -0,0 +1,271 @@
...
+		ctx.EnsureStatusOption(
+			status.AtlasStreamInstanceAddConnection(
+				akoStreamConnection.Spec.Name,
+				common.ResourceRefNamespaced{
+					Name:      akoStreamConnection.Name,
+					Namespace: akoStreamConnection.Namespace,
+				},
+			),
+		)
+	}
```
josvazg, line 103: I do not understand no ops, we are doing something so, are they really no ops?

pkg/controller/atlasstream/atlasstream_instance_controller.go
```diff
@@ -0,0 +1,186 @@
...
+	if err := r.Client.Get(ctx, akoStreamInstance.AtlasProjectObjectKey(), &project); err != nil {
+		return r.terminate(workflowCtx, workflow.Internal, err)
+	}
+
+	atlasClient, orgID, err := r.AtlasProvider.SdkClient(workflowCtx.Context, project.ConnectionSecretObjectKey(), log)
+	if err != nil {
+		return r.terminate(workflowCtx, workflow.AtlasAPIAccessNotConfigured, err)
+	}
+	workflowCtx.SdkClient = atlasClient
+	workflowCtx.OrgID = orgID
```
josvazg, line 95: Are we sure we use this?
In the PoC I could not see usages, so I removed it.

pkg/controller/atlasstream/atlasstream_connection.go
helderjs: Those are connections that didn't change anything in a reconcile loop, we still need to push them in the status.

pkg/controller/atlasstream/atlasstream_connection.go
```diff
@@ -0,0 +1,271 @@
...
+		data[key] = string(val)
+	}
+
+	return data, nil
+}
+
+func hasStreamConnectionChanged(
+	streamConnection *akov2.AtlasStreamConnection,
+	atlasStreamConnection *admin.StreamsConnection,
+	mapper streamConnectionMapper,
```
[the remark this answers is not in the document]
helderjs: I still haven't checked your PR, but yes, if we had a middle layer that wouldn't be needed. However, I'm wondering if the transformation back and forth from both sides worth it.

pkg/controller/atlasstream/atlasstream_instance_controller.go
```diff
@@ -0,0 +1,186 @@
...
+
+// ensureAtlasStreamsInstance is the central state dispatcher
+func (r *AtlasStreamsInstanceReconciler) ensureAtlasStreamsInstance(ctx context.Context, log *zap.SugaredLogger, akoStreamInstance *akov2.AtlasStreamInstance) (ctrl.Result, error) {
+	// check if stream instance is in "skipped" state
+	if customresource.ReconciliationShouldBeSkipped(akoStreamInstance) {
+		return r.skip(ctx, log, akoStreamInstance), nil
+	}
+
+	workflowCtx := customresource.MarkReconciliationStarted(r.Client, akoStreamInstance, log, ctx)
+	defer statushandler.Update(workflowCtx, r.Client, r.EventRecorder, akoStreamInstance)
```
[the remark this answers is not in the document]
helderjs: This applies all the status changes to the resources. It's kinda of a magic, IMO, but we have it in all of our controllers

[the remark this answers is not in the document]
s-urbaniak: not really magic, it simply updates the status conditions on the underlying custom resource (`akoStreamInstance` in this case). If we don't call this function, none of the status updates in the transitions would be persisted.

cmd/manager/main.go
s-urbaniak: @helderjs @josvazg i believe we don't need to priorize here anything the code is there a few lines below. `ctrl.SetupSignalHandler()` already wires things to `SIGINT` and `SIGTERM`

s-urbaniak: fixed in https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1529/commits/72c02478d062f3d9a17db77dad582d601a1a1b5c which I pushed to this PR

pkg/controller/atlasstream/atlasstream_instance_controller.go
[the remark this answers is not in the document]
s-urbaniak: side note: I just found out that this is generally broken in all controllers, see https://jira.mongodb.org/browse/CLOUDP-246312

josvazg (approved):
LGTM when CI agrees

s-urbaniak:
we need to do some more logistics around helm

roothorp (approved):
LGTM - good unit test coverage 😄

s-urbaniak (approved):

PR #1538, opened 2024-04-24
CLOUDP-245131: Run govulncheck at lint time

josvazg:
Selectively ignoring vulnerabilities.

The check is relatively cheap and makes more sense to do as early as possible. Developers are responsible for sending PRs that either pass the check or add ignored vulnerabilities to the list if they cannot be solved right away.

The script `scripts/vulncheck.sh` runs **govulncheck** removing all explicitly ignored vulnerabilities. It fails only any vulnerabilities are found after filtering the ignored ones.

The `vuln-ignore` file is expected to contain vulnerabilities using a pair of lines each in the following format:
```
# https://pkg.go.dev/vuln/GO-xxxx-yyy {reason for ignoring}
GO-xxxx-yyyy
```
- Lines starting with `"##"` will be ignored, they are self documentation.
- Lines starting with `"# "` can be used for the SDLC checklist reporting the ignored vulnerabilities on a release.
- The rest of non empty lines are used to match and skip vulnerabilities from `govulncheck`.

Note the idea is NOT to ignore vulnerabilities for ever or keep accumulating them. Instead, as soon as a fix can be applied, it should be applied and the ignore entry be removed from `vuln-ignore`.

### All Submissions:

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -24,3 +24,6 @@ jobs:
 
       - name: Run ShellCheck
         uses: bewuethr/shellcheck-action@v2
+
+      - name: Run govulncheck
+        run: make vulncheck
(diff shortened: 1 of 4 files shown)


helderjs (approved):

josvazg:
CLOUDP-245131

PR #1543, opened 2024-04-24
DB User Translation Layer & SDK migration

josvazg:
Sample Translation layer for your consideration:
- The packages under `internal/translation` contain per Altas resource translation layers for users and deployments.
- Files in the controller package should not import the SDK library to call the API or use its types.
- All translations and upgrades are confined to the `translation` code.
- Contract tests would use the translation layer directly.
- This layout also enforces proper translations and comparisons, no more funky CMP or JSONCopies.

- See commit 142314606ac35d3d5d5568c1f6acccdf10c87bc8 for a version when the translation layer is on an `atlas.go` file within the same package using the old Go API client
- See commit  5d11f7cf28ca20b62e7e589356798a80fc4318de for the migration changes to the nee SDK alone.

[Engineering Proposal: Translation Layer](https://docs.google.com/document/d/1b5Sw6-WxylDIvhZL8ZeZ0_tpGygqdtF_hluXmqo33E8/)

### All Submissions:


internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, err
+	}
+	return &Service{DatabaseUsersApi: client.DatabaseUsersApi}, nil
+}
+
+func (dus *Service) Get(ctx context.Context, db, projectID, username string) (*User, error) {
+	atlasDBUser, _, err := dus.GetDatabaseUser(ctx, projectID, db, username).Execute()
+	if err != nil {
+		if admin.IsErrorCode(err, atlas.UsernameNotFound) {
+			return nil, nil
```
s-urbaniak, line 30: how does the consumer know that the given username was not found? `nil, nil` is ambiguous.

internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, err
+	}
+	return toK8sDatabaseUser(atlasDBUser)
+}
+
+func (dus *Service) Delete(ctx context.Context, db, projectID, username string) (bool, error) {
+	_, _, err := dus.DeleteDatabaseUser(ctx, projectID, db, username).Execute()
+	if err != nil {
+		if admin.IsErrorCode(err, atlas.UsernameNotFound) {
+			return false, nil
```
s-urbaniak, line 41: the consumer only knows that deletion failed (`false`) but not why.

internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func (dus *Service) Create(ctx context.Context, au *User) error {
+	_, _, err := dus.CreateDatabaseUser(ctx, au.ProjectID, toAtlas(au)).Execute()
```
s-urbaniak, line 49: what if we need data after creation, filled by the SDK or API endpoint?

internal/translayer/dbuser/translation.go
```diff
@@ -0,0 +1,139 @@
...
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+)
+
+type User struct {
+	akov2.AtlasDatabaseUserSpec
+	Password  string
+	ProjectID string
+}
+
+func toK8sDatabaseUser(dbUser *admin.CloudDatabaseUser) (*User, error) {
```
s-urbaniak, line 20: Naming seems inconsistent. here, it is called `toK8sDatabaseUser`, below `toAtlas`. one is returning an error but not the other.

My hypothesis is that, especially in the translation layer the conversion between types needs to have a well defined contract.

internal/translayer/deployment/deployment.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"go.mongodb.org/atlas/mongodbatlas"
+	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translayer"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
+)
+
+type Service struct {
```
s-urbaniak, line 15: what is a `Service`?

s-urbaniak (commented):
general question: how does the "contract" look like which the translation layer seems to be a prerequisite for?

internal/translayer/dbuser/dbuser.go
josvazg: As said, we could translate the errors as well, the ones the consumer cares about.
In this case the bool mean `deleted` but in this case it was not, because it was not there already.

There are 2 ways to treat deletions:
- Non idempotent: producen an error if the item was already deleted.
- Idempotent: interpret the deletion command as successful is the item ceases to exists, so if it was deleted in this call or a subsequent one, it does not matter. Still you might want to distinguish if it was deleted now or already deleted.

I chose the idempotent option here, but we could revisit this decision and have a translated error case for that.

internal/translayer/dbuser/dbuser.go
josvazg: For this example I just tailored the calls to their actual usage. The creation was not using the object after the fact, so this suffices here.

If we needed to use the returned object from the API call we would use the conversion `fromAtlas...` and return it back to the consumer.

internal/translayer/deployment/deployment.go
josvazg: We could revisit naming, but a Service is a particular interface of a feature, you have a service to handle users, another for deployments, etc. The client is a separate entity that allows you to access and user services.

josvazg:
> general question: how does the "contract" look like which the translation layer seems to be a prerequisite for?

The contract is not really fully defined until you have the tests with the validations that check precisely the contract behavior. Here we only have the shape of the interactions in the service methods and the translations we make.

This is NOT the contract test PoC, just the translation layer, which is probably more than half the code required.

What is left now is the helper code to be able to put the API side in the state required to be able to test the API calls in the service in the situations that define the contract with the Atlas. Those tests inform how and when the service calls defined in the translation layer should be called from the controller to achieve the expected results. Which would allow to test the reconciliation without the API with a great deal of confidence, other than if something changes on the API side, which the contract tests will detect.

Client Contract Tests do NOT need to cover all API behaviours, just the ones our client cares about.

internal/translayer/dbuser/translation.go
josvazg: done

internal/translayer/dbuser/dbuser.go
josvazg: We could translate the errors, at least the ones the consumer cares about.

There are 2 options usually in gets7finds for an object, **not found is fine** and a nil or empty object is returned in such case or **not found is an error** that is not usually expected. In this case I chose the first option, but as said, the explicit error might be better.

internal/translayer/dbuser/translation.go
```diff
@@ -0,0 +1,179 @@
...
+	akov2.AtlasDatabaseUserSpec
+	Password  string
+	ProjectID string
+}
+
+func NewUser(spec akov2.AtlasDatabaseUserSpec, projectID, password string) *User {
+	return &User{AtlasDatabaseUserSpec: spec, ProjectID: projectID, Password: password}
+}
+
+func Normalize(spec *akov2.AtlasDatabaseUserSpec) *akov2.AtlasDatabaseUserSpec {
```
josvazg, line 25: consider deepcopy here

internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,74 @@
...
+	_, _, err = dus.CreateDatabaseUser(ctx, au.ProjectID, u).Execute()
+	return err
+}
+
+func (dus *Service) Update(ctx context.Context, au *User) error {
+	u, err := toAtlas(au)
+	if err != nil {
+		return err
+	}
+	_, _, err = dus.UpdateDatabaseUser(ctx, au.ProjectID, au.DatabaseName, au.Username, u).Execute()
```
s-urbaniak, line 72: same question as https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1543/files#r1607735324

internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
+)
+
+var (
+	// ErrorNotFound is returned when an database user was not found
+	ErrorNotFound = errors.New("database user not found")
+)
+
+type Service struct {
+	admin.DatabaseUsersApi
```
s-urbaniak, line 21: I am sceptical in embedding a foreign API `admin.DatabaseUsersApi` and thus exposing all of its methods here as it easily allows bypassing the whole service layer. Is it intentional?

internal/translayer/deployment/deployment.go
s-urbaniak: Let's make this crystal clear in the godoc then as the term `Service` has a much broader meaning than `Client`.

internal/translayer/deployment/deployment.go
```diff
@@ -0,0 +1,85 @@
...
+	serverless, _, err := ds.ListServerlessInstances(ctx, projectID).Execute()
+	if err != nil {
+		return nil, err
+	}
+	serverlessConns := serverlessToConns(serverless.GetResults())
+
+	return connSet(clusterConns, serverlessConns), nil
+}
+
+func (ds *Service) Exists(ctx context.Context, projectID, clusterName string) (bool, error) {
```
s-urbaniak, line 68: `Service.Exists` is confusing, as it implies that the method here tries to determine whether the "service" exists.

internal/translayer/deployment/deployment.go
```diff
@@ -0,0 +1,85 @@
...
+	if admin.IsErrorCode(err, atlas.ClusterNotFound) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func (ds *Service) IsReady(ctx context.Context, projectID, deploymentName string) (bool, error) {
```
s-urbaniak, line 79: same comment as for `Exists`.

How come an abstract "service" is "Ready"? I can see that this is in the "deployment" package, so as a consumer I might think "Is this method returning whether the deployment service is ready? Or is the underlying deployment ready? Or is the general purpose service ready?".

internal/translayer/deployment/translation.go
```diff
@@ -0,0 +1,96 @@
+package deployment
+
+import "go.mongodb.org/atlas-sdk/v20231115008/admin"
+
+type Conn struct {
```
s-urbaniak, line 5: please spell this out. It is not a ubiquitous abbreviation.

s-urbaniak (commented):
this is going in a good direction 👍 I am still confused by the term "Service" as it has a very abstract meaning and the methods attached to it are very concrete.

internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translayer"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
+)
+
+var (
+	// ErrorNotFound is returned when an database user was not found
+	ErrorNotFound = errors.New("database user not found")
+)
+
+type Service struct {
```
s-urbaniak, line 20: I am still hesitant about the term "Service" here. The methods are concrete so my intuition is that the struct should also be concrete, i.e. `AtlasDatabaseUsers` that (maybe) implements some common "Service" interface.

`AtlasDatabaseUsers.Get`, `AtlasDatabaseUsers.Delete`, `AtlasDatabaseUsers.Exists`, etc. then have a concrete meaning.

internal/translayer/deployment/deployment.go
josvazg: While I agree this might need a rename, lets consider that "Service" here is always qualified by the package name so in reality is `deployment.Service` here or `dbuser.Service` on the other layer.

Maybe I could rename as `Deployments` and `Users` but keep in mind it will still be a bit of stuttering as `deployment.Deployments` and `dbuser.Users`.

josvazg:
> general question: how does the "contract" look like which the translation layer seems to be a prerequisite for?

The contract is the service interface. On the other PoC I made that explicit, on this I will add the interface now.

The only requisite of the contract is that it only accepts and returns internal or Kubernetes types, it completely encapsulates Atlas SDK types as a lower protocol layer.

internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, err
+	}
+	return fromAtlas(atlasDBUser)
+}
+
+func (dus *Service) Delete(ctx context.Context, db, projectID, username string) error {
+	_, _, err := dus.DeleteDatabaseUser(ctx, projectID, db, username).Execute()
+	if err != nil {
+		if admin.IsErrorCode(err, atlas.UsernameNotFound) {
+			return errors.Join(ErrorNotFound, err)
```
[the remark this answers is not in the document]
josvazg: will fix this, thanks!

internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	return nil
+}
+
+func (dus *Service) Create(ctx context.Context, au *User) error {
+	u, err := toAtlas(au)
+	if err != nil {
+		return err
+	}
+	_, _, err = dus.CreateDatabaseUser(ctx, au.ProjectID, u).Execute()
```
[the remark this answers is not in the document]
josvazg: In the use cases we had it was not interested. I will opt for [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) here for now.

internal/translayer/dbuser/dbuser.go
josvazg: Same https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it principle

internal/translayer/deployment/deployment.go
josvazg: I will rename to `DeploymentExists`.

internal/translayer/deployment/deployment.go
```diff
@@ -0,0 +1,85 @@
...
+	for _, d := range *clusters.Results {
+		name := pointer.GetOrDefault(d.Name, "")
+		if name != "" {
+			deploymentNames = append(deploymentNames, name)
+		}
+	}
+	return deploymentNames, nil
+}
+
+func (ds *Service) ListDeploymentConns(ctx context.Context, projectID string) ([]Conn, error) {
```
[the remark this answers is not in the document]
josvazg: Will rename

internal/translayer/deployment/deployment.go
josvazg: Will rename to `DeploymentIsReady`

BTW we have a complication here. On one side we have `clusters`, on the other `serverless`. In general I use `deployment` to mean both, but will revisit that I use the terms correctly:
- `Deployment` both `Cluster` or `Serverless`.
- `Cluster` means what the old SDK called "AdvancedClusters" and ended up replacing old Clusters also.
- `Serverless` means only the serverless deployments.

internal/translayer/dbuser/dbuser.go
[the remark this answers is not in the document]
s-urbaniak: ah I see, so one can simply add it here once needed?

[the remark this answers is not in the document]
josvazg: well, if we ever need it our contract has changed, so we want to change our internal interface and add contract tests to it.
It is not as problematic as an external contract/interface, but we still need to test it as we are going to be using more of the remote API and we need to validate our assumptions on what it does return are correct.

internal/translayer/dbuser/dbuser.go
josvazg: coming in next update

internal/translayer/deployment/deployment.go
josvazg: coming in next update: I opted for `Atlas{resource plural}Service` for the interface and `ProductionAtlas{resource plural}` for the concrete type.

internal/translayer/deployment/deployment.go
josvazg: coming in next update

internal/translayer/deployment/deployment.go
josvazg: coming in next update

internal/translayer/deployment/translation.go
josvazg: coming in next update

internal/translayer/dbuser/dbuser.go
```diff
@@ -0,0 +1,81 @@
...
+	// ErrorNotFound is returned when an database user was not found
+	ErrorNotFound = errors.New("database user not found")
+)
+
+type AtlasUsersService interface {
+	Get(ctx context.Context, db, projectID, username string) (*User, error)
+	Delete(ctx context.Context, db, projectID, username string) error
+	Create(ctx context.Context, au *User) error
+	Update(ctx context.Context, au *User) error
+}
```
s-urbaniak, line 25: This looks great! 👍 🎉

internal/translayer/dbuser/translation.go
```diff
@@ -0,0 +1,183 @@
...
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+)
+
+type User struct {
+	akov2.AtlasDatabaseUserSpec
+	Password  string
+	ProjectID string
+}
+
+func NewUser(spec akov2.AtlasDatabaseUserSpec, projectID, password string) *User {
```
s-urbaniak, line 21: nit: avoid shallow-copies, this produces unnecessary memory pressure and doesn't help with nested maps and slices. I vote for simply:
```suggestion
func NewUser(spec *akov2.AtlasDatabaseUserSpec, projectID, password string) *User {
```
and a godoc whether `spec` is being mutated signalling the caller potentially has to `spec.DeepCopy` first. Note that DeepCopying should always be done on the consumer side, never here. This is the convention used in the Kubernetes source code base and other operators as well.

pkg/controller/atlasdatabaseuser/databaseuser.go
```diff
@@ -218,52 +208,59 @@ func filterScopeDeployments(user akov2.AtlasDatabaseUser, allDeploymentsInProjec
...
 	}
 	return passwordsChanged, nil
 }
 
-// TODO move to a separate utils (reuse from deployments)
-func userMatchesSpec(log *zap.SugaredLogger, atlasSpec *mongodbatlas.DatabaseUser, operatorSpec akov2.AtlasDatabaseUserSpec) (bool, error) {
-	userMerged := mongodbatlas.DatabaseUser{}
-	if err := compat.JSONCopy(&userMerged, atlasSpec); err != nil {
+func userMatchesSpec(log *zap.SugaredLogger, atlasUsername, operatorUser *akov2.AtlasDatabaseUserSpec) (bool, error) {
+	operatorCopy, err := dbuser.Normalize(operatorUser.DeepCopy())
```
s-urbaniak, line 228: DeepCopy'ing here this is actually fine 👍 We're deep copying as an internal implementation detail for normalization. Great job! 🎉

pkg/controller/atlasdatabaseuser/databaseuser.go
```diff
@@ -218,52 +208,59 @@ func filterScopeDeployments(user akov2.AtlasDatabaseUser, allDeploymentsInProjec
...
-	if d != "" {
-		log.Debugf("Users differs from spec: %s", d)
+	if atlasUsername.OIDCAuthType != operatorCopy.OIDCAuthType {
+		log.Infof("OIDCAuthType differs from spec: %q <> %q\n", atlasUsername.OIDCAuthType, operatorCopy.OIDCAuthType)
+		return false, nil
 	}
-
-	return d == "", nil
+	if atlasUsername.AWSIAMType != operatorCopy.AWSIAMType {
+		log.Infof("AWSIAMType differs from spec: %q <> %q\n", atlasUsername.AWSIAMType, operatorCopy.AWSIAMType)
```
s-urbaniak, line 250: nit: instead of logging here, return a `diff string` and the caller can decide whether this needs logging, event emission, increasing a "diff" counter, etc. etc.

internal/translayer/deployment/translation.go
```diff
@@ -0,0 +1,96 @@
...
+	SrvConnURL       string
+	PrivateURL       string
+	SrvPrivateURL    string
+	Serverless       bool
+	PrivateEndpoints []PrivateEndpoint
+}
+
+type PrivateEndpoint struct {
+	URL      string
+	SrvURL   string
```
s-urbaniak, line 17: nit: spell out `Srv` as that abbreviation is not ubiquitous.

internal/translayer/dbuser/translation.go
s-urbaniak: yes! 👏

internal/translayer/client.go
```diff
@@ -0,0 +1,29 @@
+package translayer
```
s-urbaniak, line 1: suggestion: `translayer` implies two terms, "translation" and "layer". The  latter is somewhat obvious as every package is a layer. How about just `translation`? That naming is pretty idiomatic and immediately implies what is done here, namely translating things.

internal/translayer/dbuser/translation.go
josvazg: Will fix and document

internal/translayer/dbuser/translation.go
```diff
@@ -0,0 +1,183 @@
...
+	return &User{AtlasDatabaseUserSpec: spec, ProjectID: projectID, Password: password}
+}
+
+func Normalize(spec *akov2.AtlasDatabaseUserSpec) (*akov2.AtlasDatabaseUserSpec, error) {
```
josvazg, line 25: Add godoc on making a copy before calling, as this makes changes on input

pkg/controller/atlasdatabaseuser/databaseuser.go
josvazg: will fix

internal/translayer/client.go
josvazg: coming on next PR update

internal/translayer/dbuser/translation.go
josvazg: coming next

internal/translayer/deployment/translation.go
josvazg: done next

pkg/controller/atlasdatabaseuser/databaseuser.go
josvazg: done

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -123,18 +124,16 @@ func (r *AtlasDatabaseUserReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		workflowCtx.SetConditionFromResult(api.DatabaseUserReadyType, result)
 
 		return result.ReconcileResult(), nil
 	}
-	workflowCtx.OrgID = orgID
-	workflowCtx.Client = atlasClient
 
 	// Setting protection flag to static false because ownership detection is disabled.
-	owner, err := customresource.IsOwner(databaseUser, false, customresource.IsResourceManagedByOperator, managedByAtlas(ctx, atlasClient, project.ID(), log))
+	owner, err := customresource.IsOwner(databaseUser, false, customresource.IsResourceManagedByOperator, managedByAtlas(ctx, dus, project.ID(), log))
```
helderjs, line 136: Shouldn't we wait for the removal of the ownership detection code to be merged?

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -282,28 +286,32 @@ func (r *AtlasDatabaseUserReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-func managedByAtlas(ctx context.Context, atlasClient *mongodbatlas.Client, projectID string, log *zap.SugaredLogger) customresource.AtlasChecker {
+func managedByAtlas(ctx context.Context, dus dbuser.AtlasUsersService, projectID string, log *zap.SugaredLogger) customresource.AtlasChecker {
```
helderjs, line 289: Shouldn't we wait for the removal of the ownership detection code to be merged?

scripts/int_local.sh
```diff
@@ -16,4 +16,4 @@ export MCLI_PUBLIC_API_KEY="${public_key}"
 export MCLI_PRIVATE_API_KEY="${private_key}"
 export MCLI_ORG_ID="${org_id}"
 
-AKO_INT_TEST=1 ginkgo --race --label-filter="${label}" --timeout 80m -v ./test/int -coverprofile cover.out
\ No newline at end of file
+AKO_INT_TEST=1 ginkgo --race --label-filter="${label}" --timeout 80m -v ./test/int/... -coverprofile cover.out
```
helderjs, line 19: Expanding all packages might be problematic once the go files inside `clusterwide` folder define the same package `int` on its parent. Tests will collapse with 2 test suites running at the same time.

helderjs (changes requested):
LGTM overall, I believe we should cover the translation layer more with tests, especially the conversion functions.

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -123,18 +124,16 @@ func (r *AtlasDatabaseUserReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 
-	atlasClient, orgID, err := r.AtlasProvider.Client(ctx, project.ConnectionSecretObjectKey(), log)
+	dus, err := dbuser.NewAtlasDatabaseUsersService(ctx, r.AtlasProvider, project.ConnectionSecretObjectKey(), log)
 	if err != nil {
 		result = workflow.Terminate(workflow.AtlasAPIAccessNotConfigured, err.Error())
 		workflowCtx.SetConditionFromResult(api.DatabaseUserReadyType, result)
 
 		return result.ReconcileResult(), nil
 	}
-	workflowCtx.OrgID = orgID
```
helderjs, line 133: we use `workflow.Context` everywhere and while DBUser doesn't need to be aware of orgID, I'm concerned that leaving the context incomplete might lead to bugs in the feature. Maybe worth adding a comment saying that it won't be always available once it will probably be moved out of context

helderjs (changes requested):

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: I can add a comment indeed. I removed as I do not think it is used or required, will double check it thought.

Also `workflow.Context` is a bag of things and not all controllers need everything in such a bag... IMO we should get rid of it when we can.

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: Yep, that makes sense.

I will rebase then.

scripts/int_local.sh
josvazg: should I user a new line instead?
```suggestion
AKO_INT_TEST=1 ginkgo --race --label-filter="${label}" --timeout 80m -v ./test/int -coverprofile cover.out
AKO_INT_TEST=1 ginkgo --race --label-filter="${label}" --timeout 80m -v ./test/int/clusterwide... -coverprofile cover.out

```

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: I believe next update is rebased already.

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: rebased

scripts/int_local.sh
josvazg: Done

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: Note added, BTW going forward is org id is used, it should be part of the internal type addendums, like the secret or project id.

internal/translation/dbuser/conversion.go
```diff
@@ -0,0 +1,185 @@
...
+	ProjectID string
+}
+
+// NewUser wraps a Kubernetes Atlas User Spec pointer augmenting it with projectID and password.
+func NewUser(spec *akov2.AtlasDatabaseUserSpec, projectID, password string) *User {
+	return &User{AtlasDatabaseUserSpec: spec, ProjectID: projectID, Password: password}
+}
+
+// Normalize modifies the spec and returns it back. Clone to avoid unwanted changes.
+func Normalize(spec *akov2.AtlasDatabaseUserSpec) (*akov2.AtlasDatabaseUserSpec, error) {
```
s-urbaniak, line 27: normalization is not complete. we have the following nested slices which need sorting:
- labels
- roles
- scopes

josvazg: let me fix that

josvazg: BTW note that ordering was already being done in the **fromAtlas** conversion direction. We can reuse that.

The fix will make `Normalize` an unexported implementation detail function which will be executed when data comes `fromAtlas` and when a `NewUser` is created (from the Kubernetes spec).

Also the diff code is moved into the translation layer conversion code, so it forces the user to use normalized internal user types.

internal/cmp/normalize.go
```diff
@@ -25,11 +24,11 @@ func SemanticEqual[T Normalizer[T]](this, that T) (bool, error) {
 	return reflect.DeepEqual(thisResult, thatResult), nil
 }
 
-func NormalizeSlice[S ~[]E, E any](slice S, cmp func(a, b E) int) S {
+func NormalizeSlice[S ~[]E, E any](slice S, less func(i, j int) bool) S {
```
s-urbaniak, line 27: I made the `cmp` semantics consciously and would like to stay with that. The `less` semantics are deprecated in the Go SDK:
```
$ go doc sort.SliceStable
...
    Note: in many situations, the newer slices.SortStableFunc function is more
    ergonomic and runs faster.
```

s-urbaniak:
```
$ go doc slices.SortStableFunc
package slices // import "slices"

func SortStableFunc[S ~[]E, E any](x S, cmp func(a, b E) int)
    SortStableFunc sorts the slice x while keeping the original order of equal
    elements, using cmp to compare elements in the same way as SortFunc.
```

pkg/controller/connectionsecret/connectionsecrets.go
```diff
@@ -178,6 +143,34 @@ func RemoveStaleSecretsByUserName(ctx context.Context, k8sClient client.Client,
...
+				PvtConnURL:      pe.URL,
+				PvtSrvConnURL:   pe.ServerURL,
+				PvtShardConnURL: pe.ShardURL,
+			})
+		}
+	}
+}
+
+// FillPrivateConnStrings fills private conn urls from connection strings
+// TODO: remove once all usages move over to FillPrivateConns instead
```
s-urbaniak, line 172: is there a ticket for this? if not, please create and attach `TODO(CLOUDP-...)`

internal/translation/dbuser/conversion.go
```diff
@@ -0,0 +1,230 @@
...
+	})
+	if user.DeleteAfterDate != "" { // enforce date format
+		operatorDeleteDate, err := timeutil.ParseISO8601(user.DeleteAfterDate)
+		if err != nil {
+			return nil, err
+		}
+		user.DeleteAfterDate = timeutil.FormatISO8601(operatorDeleteDate)
+	}
+	// Ensure comparisons succeed on default value
+	// TODO: Check if this is required after updating the Helm Chart CRDs
```
s-urbaniak, line 84: same as below: please create a ticket and attach `TODO(CLOUDP-....`)

scripts/int_local.sh
```diff
@@ -16,4 +16,5 @@ export MCLI_PUBLIC_API_KEY="${public_key}"
 export MCLI_PRIVATE_API_KEY="${private_key}"
 export MCLI_ORG_ID="${org_id}"
 
-AKO_INT_TEST=1 ginkgo --race --label-filter="${label}" --timeout 80m -v ./test/int -coverprofile cover.out
\ No newline at end of file
+AKO_INT_TEST=1 ginkgo --race --label-filter="${label}" --timeout 80m -v ./test/int -coverprofile cover.out
+AKO_INT_TEST=1 ginkgo --race --label-filter="${label}" --timeout 80m -v ./test/int/clusterwide... -coverprofile cover.out
```
helderjs, line 20: Maybe we don't care about it here, but wouldn't the second run override the coverage output of the previous one?

internal/cmp/normalize.go
josvazg: ok, but the new semantics are pretty difficult for ordering several chained fields... maybe I can convert the fields to string and use a string comparison.

internal/translation/dbuser/conversion.go
josvazg: I can act on this now, the release is done, so the new helm chart already takes this into account. Should be able to remove.

pkg/controller/connectionsecret/connectionsecrets.go
josvazg: will double check if this still applies

internal/cmp/normalize.go
s-urbaniak: @josvazg for `less`-semantics the `cmp` function has to return <0.

josvazg: yes, but also 0 is equal and >0 if greater. I will think this carefuly

scripts/int_local.sh
josvazg: actually can we do merge to one line again but be more specific? I will try

internal/translation/dbuser/conversion.go
josvazg: next update will test this removal

pkg/controller/connectionsecret/connectionsecrets.go
josvazg: next change refers to CLOUDP-253951

scripts/int_local.sh
josvazg: merged and tested

internal/translation/deployment/conversion.go
```diff
@@ -0,0 +1,96 @@
...
+			Serverless:       false,
+			PrivateEndpoints: fillClusterPrivateEndpoints(c.ConnectionStrings.GetPrivateEndpoint()),
+		})
+	}
+	return conns
+}
+
+func fillClusterPrivateEndpoints(cpeList []admin.ClusterDescriptionConnectionStringsPrivateEndpoint) []PrivateEndpoint {
+	pes := []PrivateEndpoint{}
+	if len(cpeList) == 0 {
```
s-urbaniak, line 39: not a blocker: this is not necessary

internal/translation/deployment/conversion.go
```diff
@@ -0,0 +1,96 @@
...
+			Serverless:       true,
+			PrivateEndpoints: fillServerlessPrivateEndpoints(s.ConnectionStrings.GetPrivateEndpoint()),
+		})
+	}
+	return conns
+}
+
+func fillServerlessPrivateEndpoints(cpeList []admin.ServerlessConnectionStringsPrivateEndpointList) []PrivateEndpoint {
+	pes := []PrivateEndpoint{}
+	if len(cpeList) == 0 {
```
s-urbaniak, line 68: not a blocker: this feels inconsistent, some methods have this, some don't.

internal/translation/deployment/deployment.go
```diff
@@ -0,0 +1,93 @@
...
+
+func NewProductionAtlasDeployments(clusterService admin.ClustersApi, serverlessAPI admin.ServerlessInstancesApi) *ProductionAtlasDeployments {
+	return &ProductionAtlasDeployments{clustersAPI: clusterService, serverlessAPI: serverlessAPI}
+}
+
+func (ds *ProductionAtlasDeployments) ListClusterNames(ctx context.Context, projectID string) ([]string, error) {
+	var deploymentNames []string
+	clusters, _, err := ds.clustersAPI.ListClusters(ctx, projectID).Execute()
+	if err != nil {
+		return nil, err
```
s-urbaniak, line 44: nit: we should be consistent with error wrapping. desirably, this should be catched by linting.

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -236,14 +244,11 @@ func (r *AtlasDatabaseUserReconciler) handleDeletion(
...
 
-	_, err := atlasClient.DatabaseUsers.Delete(ctx, dbUser.Spec.DatabaseName, project.ID(), dbUser.Spec.Username)
-	if err != nil {
-		var apiError *mongodbatlas.ErrorResponse
-		if errors.As(err, &apiError) && apiError.ErrorCode != atlas.UsernameNotFound {
-			return true, workflow.Terminate(workflow.DatabaseUserNotDeletedInAtlas, err.Error())
-		}
-
+	err := dus.Delete(ctx, dbUser.Spec.DatabaseName, project.ID(), dbUser.Spec.Username)
+	if errors.Is(err, dbuser.ErrorNotFound) {
```
s-urbaniak, line 248: we have zero unit test coverage in `atlasdatabaseuser_controller.go`. As we're touching pieces of code here, I'd like to see a unit test asserting we don't change behavior with this contribution.

pkg/controller/atlasdatabaseuser/databaseuser.go
```diff
@@ -6,25 +6,25 @@ import (
...
-	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
-	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlasdeployment"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/connectionsecret"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/workflow"
 )
 
-func (r *AtlasDatabaseUserReconciler) ensureDatabaseUser(ctx *workflow.Context, project akov2.AtlasProject, dbUser akov2.AtlasDatabaseUser) workflow.Result {
-	apiUser, err := dbUser.ToAtlas(ctx.Context, r.Client)
+func (r *AtlasDatabaseUserReconciler) ensureDatabaseUser(ctx *workflow.Context, dus dbuser.AtlasUsersService, ds deployment.AtlasDeploymentsService, project akov2.AtlasProject, dbUser akov2.AtlasDatabaseUser) workflow.Result {
+	password, err := dbUser.ReadPassword(ctx.Context, r.Client)
```
s-urbaniak, line 23: same comment here: this changes existing code while not having any unit test coverage.

pkg/controller/atlasdatabaseuser/databaseuser.go
```diff
@@ -113,27 +113,25 @@ func performUpdateInAtlas(ctx *workflow.Context, k8sClient client.Client, projec
...
-			log.Debugw("User doesn't exist. Create new user", "apiUser", apiUser)
-			if _, _, err = ctx.Client.DatabaseUsers.Create(ctx.Context, project.ID(), apiUser); err != nil {
-				return workflow.Terminate(workflow.DatabaseUserNotCreatedInAtlas, err.Error())
-			}
-			ctx.EnsureStatusOption(status.AtlasDatabaseUserPasswordVersion(currentPasswordResourceVersion))
-
-			ctx.Log.Infow("Created Atlas Database User", "name", dbUser.Spec.Username)
-			return retryAfterUpdate
-		} else {
+	au, err := dus.Get(ctx.Context, dbUser.Spec.DatabaseName, project.ID(), dbUser.Spec.Username)
```
s-urbaniak, line 116: same comment here: we change code heavily in this file while this whole code block is not unit tested.

pkg/controller/connectionsecret/connectionsecrets.go
```diff
@@ -12,75 +12,40 @@ import (
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/deployment"
 	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/workflow"
 )
 
 const ConnectionSecretsEnsuredEvent = "ConnectionSecretsEnsured"
 
-func CreateOrUpdateConnectionSecrets(ctx *workflow.Context, k8sClient client.Client, recorder record.EventRecorder, project akov2.AtlasProject, dbUser akov2.AtlasDatabaseUser) workflow.Result {
-	advancedDeployments, _, err := ctx.Client.AdvancedClusters.List(ctx.Context, project.ID(), &mongodbatlas.ListOptions{})
+func CreateOrUpdateConnectionSecrets(ctx *workflow.Context, k8sClient client.Client, ds deployment.AtlasDeploymentsService, recorder record.EventRecorder, project akov2.AtlasProject, dbUser akov2.AtlasDatabaseUser) workflow.Result {
```
s-urbaniak, line 22: same comment here: we're changing logic heavily without unit testing.

s-urbaniak (changes requested):
overall looks good to me, however I have concerns about missing unit tests in changed controller code.

internal/translation/deployment/conversion.go
josvazg: will revisit this

internal/translation/deployment/conversion.go
josvazg: will revisit

internal/translation/deployment/deployment.go
josvazg: will wrap

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: unit tests will come in next update

josvazg:
Will draft this until I rebase and add the unit tests and other suggested fixes

internal/translation/deployment/conversion.go
josvazg: Fixed on next update

internal/translation/deployment/conversion.go
josvazg: Fixed on next update

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: Adding full reconciliation unit testing is too complex at this point. Instead I will just unit test `handleDeletion` here.

pkg/controller/atlasdatabaseuser/databaseuser.go
josvazg: Ok, as explained before, will unit test `ensureDatabaseUser` specifically.

pkg/controller/atlasdatabaseuser/databaseuser.go
josvazg: This will be covered by `ensureDatabaseUser` unit tests.

internal/translation/dbuser/conversion.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// NewUser wraps a Kubernetes Atlas User Spec pointer augmenting it with projectID and password.
+func NewUser(spec *akov2.AtlasDatabaseUserSpec, projectID, password string) (*User, error) {
+	return normalize(&User{AtlasDatabaseUserSpec: spec, ProjectID: projectID, Password: password})
+}
+
+// DiffSpecs returns all differences found in the user Spec fields or a spec user and an atlas user.
+// Non Spec fields are not compared. Inputs are dbuser.User so they are normalized
+func DiffSpecs(specUser, atlasUser *User) []string {
```
s-urbaniak, line 30: This looks very verbose and implies lots of boiler plate code for every type. Do we have to implement this for every type?

Given we only check for
```
	if len(diffs) > 0 {
		return true, nil
	}
```
what is the benefit vs. a `reflect.DeepEquals`?

internal/translation/dbuser/conversion.go
```diff
@@ -0,0 +1,221 @@
+package dbuser
```
s-urbaniak, line 1: this package is missing unit tests.

internal/translation/deployment/deployment.go
```diff
@@ -0,0 +1,94 @@
...
+	ClusterExists(ctx context.Context, projectID, clusterName string) (bool, error)
+	DeploymentIsReady(ctx context.Context, projectID, deploymentName string) (bool, error)
+}
+
+type ProductionAtlasDeployments struct {
+	clustersAPI   admin.ClustersApi
+	serverlessAPI admin.ServerlessInstancesApi
+}
+
+func NewAtlasDeploymentsService(ctx context.Context, provider atlas.Provider, secretRef *types.NamespacedName, log *zap.SugaredLogger) (*ProductionAtlasDeployments, error) {
```
s-urbaniak, line 29: this is missing unit tests.

internal/translation/dbuser/conversion.go
josvazg: Precise debug info, this tells you exactly what was the first difference found and can save potentially a lot of debugging time later.
If we can achieve the same, or better like full diff reports, with less effort I am ll in.

internal/translation/dbuser/conversion.go
josvazg: I can add conversion unit tests, but the contract tests are the ones that should cover the package fully.

internal/translation/deployment/deployment.go
josvazg: Will add unit test for this one. I think it might not be covered by contract tests later.

josvazg:
Lesson learned:
This PR is bigger and took way longer than I thought it would or should. Here is what I should instead have done in hindsight:
1. TDD the translation layer alone in isolation, do not connect it yet.
2. Have another branch on top experimenting with the usage from the rest of the code, keep it as draft only.
3. Post the translation layer PR alone in the internal package, unused but fully tested with both unit tests and contract tests.
4. Merge translation later.
5. Rebase the branch using it as a new PR including unit tests and post that.

Even though has more steps it would have been easier to debug and review, so the whole process would have been faster in comparison IMHO.

internal/translation/dbuser/conversion.go
josvazg: uhm , actually I was wrong, I forgot this is so precise that it lists ALL diffs found, something a mere boolean will not do.

internal/translation/deployment/deployment.go
josvazg: done

internal/translation/dbuser/conversion.go
josvazg: Done

s-urbaniak (approved):

internal/translation/dbuser/conversion.go
s-urbaniak: but why do we need the diff?

s-urbaniak (changes requested):
sec, i still have doubts about https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1543/files/ceee31a0830619806fedee060e1ccddfb456cd2e#r1666294140

internal/translation/dbuser/conversion.go
s-urbaniak: my point is, in no place we evaluate the diff. we only check for the boolean result here: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1543/files#diff-733b0371c07644c49998b08f8fbbdfd0827a48cf7c717754a972296f3b289f24R220

s-urbaniak: My concern here is that this is a case of YAGNI. Precise diff was used in `cmp.Diff`, it is very inefficient and actually not used. If we want debug info, we can still add a debug log entry, serialized to json or yaml and print the diff.

The total cost of ownership in maintaining and mandating to maintain an exact diff is very high. If we avoid using `cmp.Diff` in the future this contribution mandates it to maintain it manually which is very verbose code.

josvazg: will make it a simple boolean equal and the debug could be after the fact and only on `false`

s-urbaniak (approved):

roothorp (approved):

PR #1546, opened 2024-04-25
Semantic equals implementation

igor-karpukhin:
### All Submissions:

This PR adds generic comparator and normalizer


pkg/api/v1/alert_configurations.go
```diff
@@ -27,6 +30,15 @@ type AlertConfiguration struct {
 	MetricThreshold *MetricThreshold `json:"metricThreshold,omitempty"`
 }
 
+func (in AlertConfiguration) Key() string {
```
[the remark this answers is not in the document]
igor-karpukhin: We can remove that, but let's do that later. This change won't affect our current JSONCopy mechanism for AlertConfigurations.

pkg/api/v1/alert_configurations.go
```diff
@@ -208,6 +228,31 @@ type Notification struct {
...
+		in.OpsGenieRegion + "|" +
+		in.OrgName + "|" +
+		in.ServiceKeyRef.Key() + "|" +
+		strconv.FormatBool(admin.GetOrDefault(in.SMSEnabled, false)) + "|" +
+		in.TeamID + "|" +
+		in.TeamName + "|" +
+		in.TypeName + "|" +
+		in.Username + "|" +
+		in.VictorOpsSecretRef.Key() + "|" +
+		"[" + strings.Join(in.Roles, ",") + "]"
```
josvazg, line 253: Why not use a hash code for all the strings instead? does it need to be readable? I to not think will ever be (readable)

s-urbaniak: hashes are not sortable which is the point of this function

igor-karpukhin: This is just an example of implementation. It has no impact on the current logic

pkg/api/v1/alert_configurations.go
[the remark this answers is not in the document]
s-urbaniak: sounds good!

igor-karpukhin:
@josvazg @s-urbaniak if your comments were resolved, could you please approve this PR so I could continue working on the Search Indexes implementation?

.licenses-gomod.sha256
```diff
@@ -1 +1 @@
-100644 d537760138665c24225e4f5725c5bd80775033a6 go.mod
+100644 60d2b0501dda29d0d258c6a0f401c95854606d5c go.mod
```
josvazg, line 1: no changes to licenses? I guess not or tests would have failed.

pkg/api/v1/alert_configurations.go
s-urbaniak: yes, we need sortable keys for normalization

s-urbaniak (approved):

internal/cmp/normalize.go
```diff
@@ -0,0 +1,104 @@
...
+	return false, -1
+}
+
+func PermuteOrder(data any, r *rand.Rand) {
+	traverse(data, func(slice reflect.Value) {
+		r.Shuffle(slice.Len(), func(i, j int) {
+			reflect.Swapper(slice.Interface())(i, j)
+		})
+	})
+}
```
josvazg, line 66: This one is only needed for test code right?

pkg/api/v1/alert_configurations.go
s-urbaniak: oh, you indeed have a point about being stable. as long as that invariant is maintained, then yes, stable hashes are good enough

josvazg (approved):
LGTM

I think we could move a function away to test code, but rest looks good

internal/cmp/normalize.go
s-urbaniak: potentially but similar as hashmaps in golang it could be a good practice to shuffle nested slices by default

helderjs (approved):
LGTM, but we still need cover it with more tests right?

internal/cmp/normalize.go
igor-karpukhin: Yes, this one is only used for testing. I'll try to clear that in a separate PR.

igor-karpukhin:
> LGTM, but we still need cover it with more tests right?

@helderjs You mean to test functions that perform sorting and comparison?

PR #1566, opened 2024-05-08
Make run eases local interactive tests

josvazg:
Using `make run RUN_YAML=my-k8s-ako-sample-here.yml` you get:
- A local kind cluster to run...
- A freshly compiled manager with latest changes in the current terminal process.
- The `RUN_YAML` file applied after the CRDs are installed and the manager is started, so it will try to enforce the given state.
No need for ad-hoc scripts to get the same.

It relies on:
- `make install-crds` to install the latest versions of the CRDs.
- `make install-credentials` to create the root Atlas Operator credentials secret off your `MCLI_` env vars. *Remember to have those setup in advance in your terminal*.
  - The `set-namespace` rule ensures the namespace set by `$(OPERATOR_NAMESPACE)` is in place.

If `RUN_YAML` is not passed in, `make run` will only launch the manager binary without any initial Kubernetes definitions in place. You should go to another terminal to apply Kubernetes YAMLs.

You can call `make run ...` several times in a row but, depending on your testing state, you might find issues with leftover state in Kubernetes or Atlas. To wipe prior Kubernetes state you can do 'make stop-kind` and `make run` will relaunch kind on its own. 

### All Submissions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,6 +101,10 @@ GOMOD_SHA := $(shell git ls-files -s go.mod | awk '{print $$1" "$$2" "$$4}')
 LICENSES_GOMOD_SHA_FILE := .licenses-gomod.sha256
 GOMOD_LICENSES_SHA := $(shell cat $(LICENSES_GOMOD_SHA_FILE))
 
+OPERATOR_NAMESPACE=atlas-operator
+OPERATOR_POD_NAME=mongodb-atlas-operator
+RUN_YAML= # Set to the YAML to run when calling make run
+
 .DEFAULT_GOAL := help
 .PHONY: help
 help: ## Show this help screen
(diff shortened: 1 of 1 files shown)


helderjs:
It's a simple and smart way to approach the problem, great idea!
Why not include a clean-up, even as simple as stop kind, to avoid conflicts with leftovers? Do you think is better to deal with it manually?

josvazg:
> It's a simple and smart way to approach the problem, great idea! Why not include a clean-up, even as simple as stop kind, to avoid conflicts with leftovers? Do you think is better to deal with it manually?

We already have `make stop-kind` which effectively wipes it all clean, at least from the Kubernetes side of things. I use it all the time.

helderjs (approved):

PR #1576, opened 2024-05-09
CLOUDP-248533: Fix X509Type as a enum

josvazg:
### All Submissions:


josvazg:
Bad news here, I thought this change introduced some errors that I had to find and fix. But it is worse actually, introduces new flakiness. It either creates such flakiness somehow or increases the likelihood of an issue that was already there was did not see so flaky before.

josvazg:
Today flakiness might have gone away locally and in the CI, not sure how to proceed. Will wait for tests to complete.

josvazg:
Only serverless failed, which is a well known flaky from before so, maybe we can merge this already.

josvazg:
related to CLOUDP-248533

helderjs (approved):

PR #1581, opened 2024-05-13
[skunkworks] centralize indexers

s-urbaniak:
This is a proposal to centralise indexing logic.

Currently, indexers are loosely coupled, mostly based on the fact that controller-runtime does not provider any abstraction for it. This PR introduces a lightweight interface `Indexer` which is being used for global registration and client-go mocks.

### DIFF
--- a/pkg/indexer/indexer.go
+++ b/pkg/indexer/indexer.go
@@ -1,0 +1,40 @@
+package indexer
+
+import (
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+type Indexer interface {
+	Object() client.Object
+	Name() string
+	Keys(client.Object) []string
+}
+
+// RegisterAll registers all known indexers to the given manager.
+// It uses the given logger to create a new named "indexer" logger,
+// passing that to each indexer.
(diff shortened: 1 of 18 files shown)


pkg/indexer/indexer.go
```diff
@@ -0,0 +1,40 @@
...
+type Indexer interface {
+	Object() client.Object
+	Name() string
+	Keys(client.Object) []string
+}
+
+// RegisterAll registers all known indexers to the given manager.
+// It uses the given logger to create a new named "indexer" logger,
+// passing that to each indexer.
+func RegisterAll(ctx context.Context, mgr manager.Manager, logger *zap.Logger) error {
```
josvazg, line 21: I wonder if this is the most ergonomic way. I mean we will need to remember to add the indexers here when we create new ones. Maybe it is not a problem.

s-urbaniak: not sure either 🤷 do you have any other suggestions?

josvazg (approved):
LGTM

pkg/indexer/indexer.go
josvazg: Was thinking about self registration, using inits, but it might be worse. Let's go for this for now.

s-urbaniak: `init` is considered an anti-pattern generally. classic dependency-injection, as done here seems a better alternative.

s-urbaniak: i.e. `init` can have nasty side-effects, potentially invoked twice if imported under different packages.

PR #1583, opened 2024-05-13
Support empty fields for new SDK

josvazg:
New SDK empty fields are expected to be nullified in the JSON data. Using just `pointer.MakePtr()` will just make those fields pointers to an empty string, which the API interprets as NOT empty.

This fix has been proven necessary by PoC at #1543 to avoid such confusion with the API.

### All Submissions:

### DIFF
--- a/internal/pointer/pointer.go
+++ b/internal/pointer/pointer.go
@@ -27,3 +27,12 @@ func GetOrDefault[T any](ptr *T, defaultValue T) T {
 func MakePtr[T any](value T) *T {
 	return &value
 }
+
+// MakePtrOrNil returns a pointer only when value is not empty.
+// Otherwise Atlas versioned API interprets a pointer to an empty value as not empty.
+func MakePtrOrNil[T comparable](value T) *T {
+	if value == *new(T) {
+		return nil
+	}
+	return &value
+}


josvazg:
@helderjs did you encounter this already in prior API SDK upgrades? are we using a different solution maybe?

helderjs (approved):
I'm okay with this new util function but I think we should not use it as the way to build correct data to Atlas, and in most of the cases we won't need it. Many endpoints behave differently according to the data sent to them, with that, I believe is better to build data according with the related type we are dealing with, for example, fo db user:

```
u := NewCloudDatabaseUser("my-db", "project-id", "my-username")
if atlasDbUser.Password != "" {
  u.SetPassword(atlasDbUser.Password)
}
```
I think this approach is better than direct assign all fields and expect we are building the right data payload.

internal/pointer/pointer.go
```diff
@@ -27,3 +27,23 @@ func GetOrDefault[T any](ptr *T, defaultValue T) T {
...
+// The API complaint if the db user password pointer goes to an empty value and
+// another auth such as X509Type is set. A nil must be sent instead to avoid issues
+//
+// https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Database-Users/operation/createDatabaseUser
+// X509type is NOT set as required in the API docs, the API complaints when it
+// is pointing to an empty value
+//
+// Generally, empty value, or more precisely, pointer to empty value is not the
+// same as unset for the API, so return nil instead of pointer to empty string
+func MakePtrOrNilOnEmpty(value string) *string {
```
s-urbaniak, line 44: I am leaving the discussion about the usefulness of this discussion to you @josvazg and @helderjs but generally I'd prefer a generic function here:
```
func MakePtrOrNil[T comparable](value T) *T {
	if value == *new(T) {
		return nil
	}
	return &value
}
```

This works with any comparable type, see https://go.dev/play/p/72xuwRshnrw.

internal/pointer/pointer.go
```diff
@@ -27,3 +27,23 @@ func GetOrDefault[T any](ptr *T, defaultValue T) T {
...
+// This creates problems in many fields, for example:
+// The API complaint if the db user password pointer goes to an empty value and
+// another auth such as X509Type is set. A nil must be sent instead to avoid issues
+//
+// https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Database-Users/operation/createDatabaseUser
+// X509type is NOT set as required in the API docs, the API complaints when it
+// is pointing to an empty value
+//
+// Generally, empty value, or more precisely, pointer to empty value is not the
+// same as unset for the API, so return nil instead of pointer to empty string
```
s-urbaniak, line 43: please make the godoc shorter, this is way too long to grasp.

internal/pointer/pointer.go
josvazg: Nice! Will use that instead

internal/pointer/pointer.go
josvazg: will do

s-urbaniak (approved):

PR #1589, opened 2024-05-14
[skunkworks] Auditing contract testing

josvazg:
These is the sample "foundations" PR of a new feature. It does NOT add anything that cannot be merged ASAP. Will hold it as it is a PoC.

It includes:
- Auditing translation layer at `internal/translayer/auditing`.
- Contract tests for the translation layer at `test/contract/auditing`.

Screenshot 2024-05-15 at 09 03 46

Infrastructure that had to be added as a one off, if this approach is followed, next PRs like will not need it:
- Contract tests workflow and wiring from cloud tests:
  - `.github/workflows/contract-tests.yml` & `.github/workflows/cloud-tests.yml`
- Added a rule for contract tests to the `Makefile`.
- Client building function for translation layer services at `internal/translater/client.go`.
- Launcher infrastructure for contract tests at `test/helper/launcher`.
  - Comes from earlier PoC #1487.
- A minor improvement for skipping tests at `test/helper/control/enable.go`.
  - And its new usages.

[Engineering Proposal: Translation Layer](https://docs.google.com/document/d/1b5Sw6-WxylDIvhZL8ZeZ0_tpGygqdtF_hluXmqo33E8/)

### All Submissions:


internal/translayer/auditing/translate.go
helderjs, line 1: On the DBUser PoC, this file is called `conversion.go`. I think we should be consistent here.

internal/translayer/auditing/translate.go
```diff
@@ -0,0 +1,56 @@
...
+	None          AuditingConfigType = "NONE"
+	FilterBuilder AuditingConfigType = "FILTER_BUILDER"
+	FilterJSON    AuditingConfigType = "FILTER_JSON"
+)
+
+type Auditing struct {
+	Enabled                   bool
+	AuditAuthorizationSuccess bool
+	ConfigurationType         AuditingConfigType
+	AuditFilter               string
```
helderjs, line 23: In the CRD is possible to have a formatted representation of the JSON, but I think we should check if Atlas returns this in the same format we set, otherwise, we need normalization.
we would fail comparing `{"a":"b"}` with
```
{
  "a": "b"
}
```

internal/translayer/client.go
helderjs, line 1: This will conflict with DBUser PoC

helderjs (changes requested):
The setup of the env, using Kind and the operator itself to prepare things for the test doesn't make me happy. Don't you think that using the Atlas client directly is a better idea? Like this, the contract test would have zero dependencies to be run.

internal/translayer/auditing/translate.go
josvazg: I will rename to `conversion.go` and also the package from `translayer` to `translation` so it will be:
`internal/translation/auditing/conversion.go`

Thanks for the heads up!

internal/translayer/auditing/translate.go
josvazg: Right! problem is, the Spec is added in the next PR. So not sure if this is the right place to fix, as the type changes quite a lot in this case.

internal/translayer/client.go
josvazg: First one wins, should be the same code. I will make take care of rebasing properly, but both PRs need it as they are right now.
Alternatively I could push the prerequisites on a previous PR.

josvazg:
> The setup of the env, using Kind and the operator itself to prepare things for the test doesn't make me happy. Don't you think that using the Atlas client directly is a better idea? Like this, the contract test would have zero dependencies to be run.

This way we are dogfooding, we depend on us.

I did a try without the operator and it got quite complex. Note we need not only provide a project, or a project and a deployment, but arbitrary setups, which the operator can do just by passing it a YAML input file.

I think my old attempt is there closed somewhere.

josvazg:
@helderjs I found my old attempt, see that is ugly as hell: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1438

Do you have something else in mind?

internal/translayer/auditing/translate.go
josvazg: Definitely this is fixed in the next PR #1594, we use this:

```go
	// AuditFilter contains the JSON/YAML definition of the audit logging filter
	AuditFilter *apiextensionsv1.JSON `json:"auditFilter,omitempty"`
```

internal/translayer/auditing/translate.go
josvazg: done in next update

helderjs (approved):

roothorp (approved):

internal/translation/auditing/conversion.go
```diff
@@ -0,0 +1,56 @@
+package auditing
```
s-urbaniak, line 1:
```suggestion
package audit
```
shorter?

internal/translation/client.go
```diff
@@ -0,0 +1,20 @@
...
+	"fmt"
+
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
+)
+
+func NewVersionedClient(ctx context.Context, provider atlas.Provider, secretRef *types.NamespacedName, log *zap.SugaredLogger) (*admin.APIClient, error) {
```
s-urbaniak, line 14: general nit that applies everywhere: please add godoc to public methods.

test/contract/auditing/auditing_test.go
```diff
@@ -0,0 +1,144 @@
...
+		return
+	}
+	l := launcher.NewFromEnv(testVersion)
+	l.Launch(
+		testYml,
+		launcher.WaitReady("atlasprojects/my-project", 30*time.Second))
+	if !control.Enabled("SKIP_CLEANUP") { // allow to reuse Atlas resources for local tests
+		defer l.Cleanup()
+	}
+	m.Run()
```
s-urbaniak, line 39: if you use `testing.M` you must `os.Exit` with some exit code otherwise tests will never yield exit code 1, even if tests fail. As per godoc:
```
$ go doc testing
...
TestMain runs in the main goroutine and can do whatever
setup and teardown is necessary around a call to m.Run. m.Run will return an
exit code that may be passed to os.Exit. If TestMain returns, the test wrapper
will pass the result of m.Run to os.Exit itself.
...
$ go doc testing.M.Run
package testing // import "testing"

func (m *M) Run() (code int)
    Run runs the tests. It returns an exit code to pass to os.Exit.
```
so you at least need:
```suggestion
	os.Exit(m.Run())
```

test/contract/auditing/auditing_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+func TestMain(m *testing.M) {
+	if !control.Enabled("AKO_CONTRACT_TEST") {
+		log.Printf("Skipping contract test as AKO_CONTRACT_TEST is unset")
+		return
+	}
+	l := launcher.NewFromEnv(testVersion)
+	l.Launch(
+		testYml,
+		launcher.WaitReady("atlasprojects/my-project", 30*time.Second))
```
s-urbaniak, line 35: nit: 30 seconds might be quite aggressive

test/helper/launcher/launcher.go
```diff
@@ -0,0 +1,257 @@
...
+	LauncherTestInstall = "test-ako"
+
+	HelmRepoURL = "https://mongodb.github.io/helm-charts"
+
+	RepoRef = "mongodb"
+
+	OperatorChart = "mongodb-atlas-operator"
+
+	AtlasURI = "https://cloud-qa.mongodb.com"
+
```
s-urbaniak, line 25: nit: let's remove the superfluous newlines between consts.

test/helper/launcher/launcher.go
```diff
@@ -0,0 +1,257 @@
...
+	return &Launcher{credentials: creds, version: version, cmdOutput: cmdOutput, appliedYAMLs: []string{}}
+}
+
+// NewFromEnv creates an operator Launcher using defaults and credentials from env vars
+func NewFromEnv(version string) *Launcher {
+	return NewLauncher(credentialsFromEnv(), version, true)
+}
+
+// MustSetEnv sets the env var value given, or panics if the env var is not set
+func MustSetEnv(envvar string) string {
```
s-urbaniak, line 51: the method name `MustSetEnv` is confusing. As per Go idioms `Set` always implies an imperative action being performed, i.e `MustSetThisThing` which will panic if setting `ThisThing` fails. I suggest to rename to:
```suggestion
func MustLookupEnv(envvar string) string {
```

test/helper/launcher/launcher.go
```diff
@@ -0,0 +1,257 @@
...
+	if !ok {
+		panic(fmt.Errorf("environment variable %s not set", envvar))
+	}
+	return value
+}
+
+// Launch will try to launch the operator and apply the given YAML for it to handle
+func (l *Launcher) Launch(yml string, waitCfg *WaitConfig) error {
+	if err := l.ensureK8sCluster(); err != nil {
+		return err
```
s-urbaniak, line 62: return `fmt.Errorf("... %w", err)` everywhere in this function.

test/helper/launcher/launcher.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// Cleanup related Launcher test resources, merely wipe the kind cluster
+func (l *Launcher) Cleanup() error {
+	for _, yml := range l.appliedYAMLs {
+		l.kubeDelete(yml)
+	}
+	if l.clearOperator {
+		if err := l.uninstall(); err != nil {
+			return err
```
s-urbaniak, line 88: return `fmt.Errorf("... %w", err)` everywhere in this function.

test/helper/launcher/launcher.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	return l.silentRun("kubectl", "label", "secret", AtlasSecretName, "atlas.mongodb.com/type=credentials")
+}
+
+func (l *Launcher) kubeDeleteAtlasSecret() error {
+	return l.silentRun("kubectl", "delete", "secret", AtlasSecretName)
+}
+
+func (l *Launcher) kubeApply(yml string) error {
+	return l.pipeRun(yml, "kubectl", "apply", "-f", "-")
```
s-urbaniak, line 207: nit: hiding this behind small function masks implementation details of the underlying kubectl command, i.e. it is now known in 1. what namespace `kubeApply` operates in and 2. what command line arguments are used. For clarity my taste would be to inline this otherwise I see the danger of an over-abstraction over kubectl and lack of clarity for the reader what happens here.

test/helper/launcher/launcher_test.go
```diff
@@ -0,0 +1,35 @@
...
+	testVersion = "2.1.0"
+)
+
+func TestLaunch(t *testing.T) {
+	control.SkipTestUnless(t, "AKO_LAUNCHER_TEST")
+
+	l := launcher.NewFromEnv(testVersion)
+	assert.NoError(t, l.Launch(
+		testYml,
+		launcher.WaitReady("atlasprojects/my-project", 30*time.Second)))
```
s-urbaniak, line 29: as above, this timeout looks pretty aggressive.

test/helper/launcher/wait.go
```diff
@@ -0,0 +1,28 @@
...
+func WaitReady(target string, timeout time.Duration) *WaitConfig {
+	return &WaitConfig{Condition: "condition=Ready", Target: target, Timeout: timeout}
+}
+
+func (cfg *WaitConfig) waitArgs() []string {
+	args := []string{"wait", fmt.Sprintf("--for=%s", cfg.Condition), cfg.Target}
+	if cfg.Timeout != 0 {
+		args = append(args, fmt.Sprintf("--timeout=%v", cfg.Timeout))
+	}
+	return args
```
s-urbaniak, line 27: hmmmmm, not sure we should go into the business of building an abstraction framework over `kubectl` invocations. This is not the focus of the challenge at hand.

Not a blocker but I am sceptical.

internal/translation/auditing/auditing.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
+)
+
+type AtlasAuditService interface {
+	Get(ctx context.Context, projectID string) (*Auditing, error)
+	Set(ctx context.Context, projectID string, auditing *Auditing) error
+}
+
+type ProductionAtlasAudit struct {
```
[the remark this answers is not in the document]
josvazg: Agreed, I will revisit the naming. Probably apply the suggestion as is.

internal/translation/auditing/auditing.go
```diff
@@ -0,0 +1,47 @@
...
+
+func NewService(ctx context.Context, provider atlas.Provider, secretRef *types.NamespacedName, log *zap.SugaredLogger) (*ProductionAtlasAudit, error) {
+	client, err := translation.NewVersionedClient(ctx, provider, secretRef, log)
+	if err != nil {
+		return nil, err
+	}
+	return NewProductionAtlasAudit(client.AuditingApi), nil
+}
+
+func NewProductionAtlasAudit(api admin.AuditingApi) *ProductionAtlasAudit {
```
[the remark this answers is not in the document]
josvazg: I disagree, it was you on another PR comment that made me realise the API had **not** to be embedded but instead properly encapsulated. We do not want anyone calling the API directly by mistake, only the public service API endpoints should be usable. And that means both the internal must be not exported and a constructor function is needed.

internal/translation/auditing/conversion.go
```diff
@@ -0,0 +1,56 @@
...
+
+type AuditingConfigType string
+
+const (
+	None          AuditingConfigType = "NONE"
+	FilterBuilder AuditingConfigType = "FILTER_BUILDER"
+	FilterJSON    AuditingConfigType = "FILTER_JSON"
+)
+
+type Auditing struct {
```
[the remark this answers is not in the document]
josvazg: Will rename as suggested.

internal/translation/auditing/auditing.go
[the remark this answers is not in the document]
s-urbaniak: > We do not want anyone calling the API directly by mistake

indeed, that is a good argument

internal/translation/auditing/conversion.go
josvazg: As `audit.AuditLog` for the interface and `audit.AuditLogService` for the struct?

BTW I usually think the other way around, `...service` for the abstract interface while the other seems to be more appropriate for the struct. Will think about it, might be a remnant from my old Java days.

test/helper/launcher/launcher.go
josvazg: Why are we arguing about an internal implementation detail? This is not a public API.

The idea is we could replace the launcher with something that does not use the Operator or kubectl directly. The Interface we care about is the one seen from the contract test code.

test/helper/launcher/wait.go
josvazg: Again, this is not part of the public API of the launcher. We are in the business of taking a YAML and making it real in Atlas. This is one way of doing it, there can be other better ones we can figure out later. This unblocks us for now to progress, but can be replaced later as needed.

s-urbaniak: > this is not part of the public API

ok, as long as this stays private, I am fine.

test/helper/launcher/launcher.go
s-urbaniak: as per comment below, agreed

internal/translation/auditing/conversion.go
s-urbaniak: > As audit.AuditLog for the interface and audit.AuditLogService for the struct?

no, `audit.AuditLogService` for the interface and `audit.AtlasAuditLogService` for the struct.

internal/translation/auditing/conversion.go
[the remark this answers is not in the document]
josvazg: coming in next update

test/contract/auditing/auditing_test.go
josvazg: Coming next

test/contract/auditing/auditing_test.go
josvazg: coming next

test/helper/launcher/launcher.go
josvazg: coming next

test/helper/launcher/launcher.go
josvazg: coming next

test/helper/launcher/launcher_test.go
josvazg: fix coming

josvazg:
@s-urbaniak I believe I applied all your feedback.

s-urbaniak (approved):

josvazg:
Related to CLOUDP-242997

PR #1591, opened 2024-05-15
CLOUDP-244166: [skunkworks] refactor backup schedules to use indexers

s-urbaniak:
This is a first iteration on converting existing resource watcher implementations in favour of indexers.

Here, we are converting backup schedules to trigger events on deployments via indexers rather than via the deprecated resource watcher.


pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -518,6 +522,43 @@ func (r *AtlasDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+func (r *AtlasDeploymentReconciler) findDeploymentsForBackupSchedule(ctx context.Context, obj client.Object) []reconcile.Request {
```
s-urbaniak, line 525: this is missing a unit test

PR #1606, opened 2024-05-22
Fix teams e2e test isolation

josvazg:
The `teams` `e2e` test was frequently flaky. We also noticed it probably had isolation issues, as it tended to fail consistently if another `teams` test was running concurrently, like it happens on nightlies.

This fix managed to get a couple of concurrent teams test to succeed by using separate team test names. Note teams are  resources at the org level, so projects might fight each other when not using unique names.

This might not be the only source of flakiness in this test, but this PR seems a good first step.

### All Submissions:


s-urbaniak (approved):
lgtm, i am fine with skipping ticket creation here 👍

PR #1608, opened 2024-05-22
CLOUDP-246611: Wire context in the integration tests

helderjs:
### All Submissions:

### DIFF
--- a/test/int/integration_suite_test.go
+++ b/test/int/integration_suite_test.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	ctrzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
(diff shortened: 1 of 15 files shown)


s-urbaniak (approved):
This is outstanding! 👏 🎉 🙇

s-urbaniak:
One comment. As per https://onsi.github.io/ginkgo/ we can directly wire the context in all ginkgo specs transparently, i.e.:

```
It("can save books", func(ctx SpecContext) {
      book := &books.Book{
        Title:  "Les Miserables",
        Author: "Victor Hugo",
        Pages:  2783,
      }

      Expect(libraryClient.SaveBook(ctx, book)).To(Succeed())
      Expect(libraryClient.ListBooks(ctx)).To(ContainElement(book))
})
```

> when such a node is detected Ginkgo will automatically supply a SpecContext object. This SpecContext object satisfies the context.Context interface and can be used anywhere a context.Context object is used. When a spec times out or is interrupted by the user (see below) Ginkgo will cancel the SpecContext to signal to the spec that it is time to exit. In the case above, it is assumed that libraryClient knows how to return once ctx is cancelled.

test/int/integration_suite_test.go
```diff
@@ -71,27 +72,23 @@ const (
...
 
 	// These variables are initialized once per each node
+	ctx         context.Context
 	k8sClient   client.Client
 	atlasClient *admin.APIClient
 
 	// These variables are per each test and are changed by each BeforeRun
-	namespace         corev1.Namespace
-	cfg               *rest.Config
+	testNamespace     corev1.Namespace
```
josvazg, line 83: Would this not be a concurrency issue. I do not think the namespace can be shared across concurrent tests.

test/int/integration_suite_test.go
```diff
@@ -71,27 +72,23 @@ const (
 var (
 	// This variable is "global" - as is visible only on the first ginkgo node
 	testEnv *envtest.Environment
+	cfg     *rest.Config
```
josvazg, line 75: What is this config about? Is this config immutable and the same for all tests, even concurrent ones?

test/int/integration_suite_test.go
```diff
@@ -183,159 +197,162 @@ func defaultTimeouts() {
...
+			ObjectDeletionProtection:    deletionProtection,
+			SubObjectDeletionProtection: false,
+		}).SetupWithManager(k8sManager)
 		Expect(err).ToNot(HaveOccurred())
-	}()
+	})
 
-	return &namespace, managerCancelFunc
-}
+	By("Setting up database user controller", func() {
```
josvazg, line 279: nit

I would not have refactor these with such care. All this code needs to die soon.

There should be a single `setupAKOManager` under `pkg/` that can be called from **main**, **integration**, **integration wide** and **e2e** tests all the same.

hopefully next time we touch here is to remove code.

josvazg (commented):
Just have a few questions, probably nothing, but would like to understand before I approve

josvazg:
Oops! I just checked the CI expecting to see the usual e2e suspects there red, but the most of integration tests are red now. I believe this change introduces a problem.

helderjs:
The task was done in another [PR](https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1611) keeping the context defined in the ticket.

PR #1609, opened 2024-05-23
Refactor searchindex translation layer

josvazg:
### All Submissions:


josvazg:
The refactor has messed up the unit tests due to the initialisation of the service. I need to figure out a way to do teh initialisation that works for production and test and is acceptably neat.

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -96,7 +97,8 @@ func (r *AtlasDeploymentReconciler) advancedDeploymentIdle(
 		}
 	}
 
-	result := handleSearchIndexes(ctx, r.Client, deployment, project.ID())
+	searchService := searchindex.NewProductionAtlasSearch(ctx.SdkClient.AtlasSearchApi)
```
helderjs, line 100: Why doesn't this service depend on the client provider as the other?

pkg/controller/atlasdeployment/searchindex_test.go
```diff
@@ -57,32 +59,25 @@ func Test_searchIndexReconciler(t *testing.T) {
...
-		mockSearchAPI.EXPECT().
-			CreateAtlasSearchIndexExecute(admin.CreateAtlasSearchIndexApiRequest{ApiService: mockSearchAPI}).
-			Return(&admin.ClusterSearchIndex{
-				CollectionName: "",
-				Database:       "",
-				IndexID:        pointer.MakePtr("testID"),
-				Name:           "",
-				Status:         pointer.MakePtr("NOT STARTED"),
-			}, &http.Response{StatusCode: http.StatusCreated}, nil)
+		fakeAtlasSearch := &fake.FakeAtlasSearch{
```
helderjs, line 62: Why did you add this custom mock? what's the problem with using the SDK mock?

pkg/controller/atlasdeployment/advanced_deployment_test.go
```diff
@@ -242,69 +239,3 @@ func TestDbUserBelongsToProjects(t *testing.T) {
 		assert.True(t, dbUserBelongsToProject(dbUser, project))
 	})
 }
-
-func TestAdvancedDeploymentIdle(t *testing.T) {
-	t.Run("should not handle search nodes in Atlas Gov", func(t *testing.T) {
```
helderjs, line 247: Why was this test removed?

josvazg: We looked at it for a while, because the migrate broke this test. @s-urbaniak noticed the test is not at the right level of abstraction and can not possibly test what the aim of the test says. It just assumes that calling the idle state and getting an OK workflow means what the test expects, which might not be true.

Testing what this test aimed to test required a refactor or mock outside of the scope of this migration change.

pkg/controller/atlasdeployment/searchindex_test.go
josvazg: Having the translation layer means we can chose to test more easily at the translation layer interface.

Testing at the API level is not only more verbose and awkward, but also redundant in most cases, because the translation layer logic is covered by contract tests and conversion unit tests.

Still, we might want to use the SDK mocks at times. For instance, this migration translated tests that were validating logic that was moved within the translation layer, and for those tests the SDK mocks needed to be used as the test had to reach inside said translation layer. For the rest of the test that just assume a certain reply by the API but which validate code at the reconciliation logic, the simpler translation layer interface is preferred.

pkg/controller/atlasdeployment/advanced_deployment.go
josvazg: uhm, let me fix that

pkg/controller/atlasdeployment/searchindex_test.go
helderjs: The SDK mocks still have a lot to improve, but I'm afraid that running away from it on our unit tests might lead to confusion on how to mock API calls.

pkg/controller/atlasdeployment/advanced_deployment.go
josvazg: Wait I get it now. This is using the provider indirectly, because it uses `ctx.SdkClient` which comes from the provider here:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/pkg/controller/atlasdeployment/atlasdeployment_controller.go#L153

```go
	atlasSdkClient, _, err := r.AtlasProvider.SdkClient(workflowCtx.Context, project.ConnectionSecretObjectKey(), log)
	if err != nil {
		result := workflow.Terminate(workflow.AtlasAPIAccessNotConfigured, err.Error())
		workflowCtx.SetConditionFromResult(api.DeploymentReadyType, result)
	}
	workflowCtx.SdkClient = atlasSdkClient

```

BTW I think we should revisit this. There is **no** reason to compute this dependency per reconcile call.

Instead the reconciler could be set up with whatever interface implementations needed... which basically will mean killing the provider or at least using it at reconciliation build time, not at reconcile runtime.

josvazg: BTW I do agree that the naming `NewProductionAtlasSearch` is probably not the best and has started this confusion. But again, I do not think we should be calling this dependency injections here on each call.

helderjs: That would be great, but how are you going to get the credentials, usually defined per project, to set up the client?

pkg/controller/atlasdeployment/searchindex_test.go
josvazg: There should not be confusion, we now have a layer between the reconciliation logic and the API calls. Unit tests should not try to test 2 layers at once when they can avoid it, or unless there is a very good reason for it:

The layers we have are:

1. **Reconcile logic** - Unit tests itself against the translation layer interface, preferably. Probably always on new code.
2. **Translation layer** - Unit tests itself (when needed) against the SDK mocks. Still most conversion unit tests may not need to mock the SDK and Contract Tests do not mock the SDK at all. But if we need to unit test a full API call for whatever reason, like some tricky logic within that is not just conversions, the SDK mocks are the ones to use here.
3. **SDK**

helderjs: The proper way would be to mock the translation layer service, but you are here adding a custom mock that does the same as the SDK mock. I see no reason to replace the SDK mock if you're still faking the internal calls abstracted by the translation layer. IMO, you either mock translation service or use SDK mocks

josvazg: This new fake is for the translation layer service interface:

```go
type FakeAtlasSearch struct {
	GetIndexFn    func(ctx context.Context, projectID, clusterName, indexName, indexID string) (*searchindex.SearchIndex, error)
	CreateIndexFn func(ctx context.Context, projectID, clusterName string, index *searchindex.SearchIndex) (*searchindex.SearchIndex, error)
	DeleteIndexFn func(ctx context.Context, projectID, clusterName, indexID string) error
	UpdateIndexFn func(ctx context.Context, projectID, clusterName string, index *searchindex.SearchIndex) (*searchindex.SearchIndex, error)
}
```

This is NOT the same as the Atlas SDK API call.

For instance, here we are mocking:
```go
CreateIndex(context.Context, string, string, *searchindex.SearchIndex) (*searchindex.SearchIndex, error)
```
Instead of:
```go
CreateAtlasSearchIndex(context.Context, string, string, *admin.ClusterSearchIndex) (admin.CreateAtlasSearchIndexApiRequest)
+
CreateAtlasSearchIndexExecute(admin.CreateAtlasSearchIndexApiRequest)
```

It is not the same. It is a thin layer, but it is a layer.

internal/translation/searchindex/fake/fake.go
```diff
@@ -0,0 +1,42 @@
+package fake
+
+import (
+	"context"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/searchindex"
+)
+
+type FakeAtlasSearch struct {
+	GetIndexFn    func(ctx context.Context, projectID, clusterName, indexName, indexID string) (*searchindex.SearchIndex, error)
```
s-urbaniak, line 10:
```suggestion
	GetIndexFunc    func(ctx context.Context, projectID, clusterName, indexName, indexID string) (*searchindex.SearchIndex, error)
```
The `Func` infix is idiomatic in Go, let's stick to that.

helderjs (approved):

internal/translation/searchindex/fake/fake.go
```diff
@@ -0,0 +1,42 @@
...
+
+type FakeAtlasSearch struct {
+	GetIndexFunc    func(ctx context.Context, projectID, clusterName, indexName, indexID string) (*searchindex.SearchIndex, error)
+	CreateIndexFunc func(ctx context.Context, projectID, clusterName string, index *searchindex.SearchIndex) (*searchindex.SearchIndex, error)
+	DeleteIndexFunc func(ctx context.Context, projectID, clusterName, indexID string) error
+	UpdateIndexFunc func(ctx context.Context, projectID, clusterName string, index *searchindex.SearchIndex) (*searchindex.SearchIndex, error)
+}
+
+func (fas *FakeAtlasSearch) GetIndex(ctx context.Context, projectID, clusterName, indexName, indexID string) (*searchindex.SearchIndex, error) {
+	if fas.GetIndexFunc == nil {
```
s-urbaniak, line 17: this nil check is superfluous. It will panic anyways in the next line.

s-urbaniak: (if `GetIndexFunc` is nil)

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,109 @@
...
+	DeleteIndex(ctx context.Context, projectID, clusterName, indexID string) error
+	UpdateIndex(ctx context.Context, projectID, clusterName string, index *SearchIndex) (*SearchIndex, error)
+}
+
+type ProductionAtlasSearch struct {
+	searchAPI admin.AtlasSearchApi
+}
+
+func NewAtlasDatabaseUsersService(ctx context.Context, provider atlas.Provider, secretRef *types.NamespacedName, log *zap.SugaredLogger) (*ProductionAtlasSearch, error) {
+	client, err := translation.NewVersionedClient(ctx, provider, secretRef, log)
```
s-urbaniak, line 34: so each service maintains its own `client` instance? Is this efficient?

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,109 @@
...
+	atlasIndex, err := index.toAtlas()
+	if err != nil {
+		return nil, err
+	}
+	resp, httpResp, err := si.searchAPI.CreateAtlasSearchIndex(ctx, projectID, clusterName, atlasIndex).Execute()
+	if err != nil || httpResp.StatusCode != http.StatusCreated && httpResp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to create index: %w, status: %d", err, httpResp.StatusCode)
+	}
+	if resp == nil {
+		return nil, fmt.Errorf("returned an empty index as a result of creation")
```
s-urbaniak, line 75: there is no formatting here, just
```suggestion
		return nil, errors.New("returned an empty index as a result of creation")
```

this should've been captured by linting 🤔

s-urbaniak (changes requested):
a couple of comments, else looking great 👍

internal/translation/searchindex/fake/fake.go
josvazg: I like to have explicit error messages, but ok

internal/translation/searchindex/searchIndexsvc.go
josvazg: The other option is to call the `NewProductionAtlasSearch` ... which BTW i think we agreed to rename on other translation layer PRs. I will revisit that and see if we can do away with this function or not.

internal/translation/searchindex/fake/fake.go
josvazg: fixed

internal/translation/searchindex/searchIndexsvc.go
josvazg: It was not even used in this case, so I removed. I also renamed the implementing type, no production prefix now.

internal/translation/searchindex/searchIndexsvc.go
josvazg: done

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		return nil, err
+	}
+	if resp == nil {
+		return nil, fmt.Errorf("received an empty index. Index: %s(%s). Status code: %d, E: %w",
+			indexName, indexID, httpResp.StatusCode, err)
+	}
+	stateInAtlas, err := fromAtlas(*resp)
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert index to AKO. Index: %s(%s, E: %w",
```
s-urbaniak, line 46:
```suggestion
		return nil, fmt.Errorf("unable to convert index to AKO. Index: %s(%s, E: %w",
```

s-urbaniak, line 46:
```suggestion
		return nil, fmt.Errorf("unable to convert index: %s(%s): %w",
```

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+func (si *SearchIndexes) GetIndex(ctx context.Context, projectID, clusterName, indexName, indexID string) (*SearchIndex, error) {
+	resp, httpResp, err := si.searchAPI.GetAtlasSearchIndex(ctx, projectID, clusterName, indexID).Execute()
+	if err != nil {
+		if httpResp.StatusCode == http.StatusNotFound {
+			return nil, errors.Join(err, ErrNotFound)
+		}
+		return nil, err
+	}
+	if resp == nil {
+		return nil, fmt.Errorf("received an empty index. Index: %s(%s). Status code: %d, E: %w",
```
s-urbaniak, line 41:
```suggestion
		return nil, fmt.Errorf("received an empty index %s(%s), status code %d: %w",
```

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (si *SearchIndexes) CreateIndex(ctx context.Context, projectID, clusterName string, index *SearchIndex) (*SearchIndex, error) {
+	atlasIndex, err := index.toAtlas()
+	if err != nil {
+		return nil, err
+	}
+	resp, httpResp, err := si.searchAPI.CreateAtlasSearchIndex(ctx, projectID, clusterName, atlasIndex).Execute()
+	if err != nil || httpResp.StatusCode != http.StatusCreated && httpResp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to create index: %w, status: %d", err, httpResp.StatusCode)
```
s-urbaniak, line 59:
```suggestion
		return nil, fmt.Errorf("failed to create index, status code %d: %w", httpResp.StatusCode, err)
```

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	if resp == nil {
+		return nil, errors.New("returned an empty index as a result of creation")
+	}
+	akoIndex, err := fromAtlas(*resp)
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert index to AKO: %w", err)
```
s-urbaniak, line 66:
```suggestion
		return nil, fmt.Errorf("unable to convert index: %w", err)
```

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+	atlasIndex, err := index.toAtlas()
+	if err != nil {
+		return nil, err
+	}
+	resp, httpResp, err := si.searchAPI.CreateAtlasSearchIndex(ctx, projectID, clusterName, atlasIndex).Execute()
+	if err != nil || httpResp.StatusCode != http.StatusCreated && httpResp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to create index: %w, status: %d", err, httpResp.StatusCode)
+	}
+	if resp == nil {
+		return nil, errors.New("returned an empty index as a result of creation")
```
s-urbaniak, line 62:
```suggestion
		return nil, errors.New("empty response when creating index.")
```

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert index to AKO: %w", err)
+	}
+	return akoIndex, nil
+}
+
+func (si *SearchIndexes) DeleteIndex(ctx context.Context, projectID, clusterName, indexID string) error {
+	_, resp, err := si.searchAPI.DeleteAtlasSearchIndex(ctx, projectID, clusterName, indexID).Execute()
+	if resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusNotFound || err != nil {
+		return fmt.Errorf("failed to delete index: %w, status: %d", err, resp.StatusCode)
```
s-urbaniak, line 74:
```suggestion
		return fmt.Errorf("error deleting index, status code %d: %w", resp.StatusCode, err)
```

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return nil
+}
+
+func (si *SearchIndexes) UpdateIndex(ctx context.Context, projectID, clusterName string, index *SearchIndex) (*SearchIndex, error) {
+	atlasIndex, err := index.toAtlas()
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert index to AKO: %w", err)
```
s-urbaniak, line 82:
```suggestion
		return nil, fmt.Errorf("error converting index: %w", err)
```

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	resp, httpResp, err := si.searchAPI.UpdateAtlasSearchIndex(ctx, projectID, clusterName, index.GetID(), atlasIndex).Execute()
+	if httpResp.StatusCode != http.StatusCreated && httpResp.StatusCode != http.StatusOK || err != nil {
+		return nil, fmt.Errorf("failed to update index: %w, status: %d", err, httpResp.StatusCode)
```
s-urbaniak, line 86:
```suggestion
		return nil, fmt.Errorf("error updating index, status code %d: %w", httpResp.StatusCode, err)
```

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	if resp == nil {
+		return nil, fmt.Errorf("update returned an empty index: %w", err)
+	}
+	akoIndex, err := fromAtlas(*resp)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert updated index to AKO: %w", err)
```
s-urbaniak, line 93:
```suggestion
		return nil, fmt.Errorf("error converting index: %w", err)
```

s-urbaniak (changes requested):
some final improvements to error messages.

helderjs (approved):
I'd only move mocks to `internal/mocks/translation` :see_no_evil: . I'd also think about use mockery instead of manually add these mocks, the same way is done for the SDK

internal/translation/searchindex/searchIndexsvc.go
```diff
@@ -0,0 +1,96 @@
...
+	atlasIndex, err := index.toAtlas()
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert index to AKO: %w", err)
+	}
+	resp, httpResp, err := si.searchAPI.UpdateAtlasSearchIndex(ctx, projectID, clusterName, index.GetID(), atlasIndex).Execute()
+	if httpResp.StatusCode != http.StatusCreated && httpResp.StatusCode != http.StatusOK || err != nil {
+		return nil, fmt.Errorf("failed to update index: %w, status: %d", err, httpResp.StatusCode)
+	}
+	if resp == nil {
+		return nil, fmt.Errorf("update returned an empty index: %w", err)
```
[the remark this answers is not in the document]
josvazg: wait, this is an update, not a create operation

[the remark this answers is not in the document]
s-urbaniak: indeed :)

s-urbaniak (approved):

josvazg:
> I'd only move mocks to `internal/mocks/translation` 🙈 . I'd also think about use mockery instead of manually add these mocks, the same way is done for the SDK

Lets make a final decision, I volunteer to make the refactors whatever the decision is finally.

PR #1610, opened 2024-05-24
[DO NOT MERGE] Translation Layer Template

josvazg:
This is an example template that can help understand the gist of a translation layer, or even be used as a starting point to implement a translation layer for a new resource or refactor an existing one.

Translation layer public interface:
- An resource interface with the operations the reconciliation would need to do on the resource.
  - All arguments and return types are internal and or Kubernetes CRDs.
- A way to create the internal version of the resource from the CRD and some attachments, like secrets, etc.
- A way to normalize the resource internal type defined in Kubernetes to allow it to be compered with the same type converted from Atlas.
- Optionally, the comparison function of internal resource types might also be defined in the translation layer.

Key features:
- The translation layer completely encapsulates the Atlas API interactions:
  - No **Atlas API calls are exposed elsewhere** in the code.
  - No **Atlas API data types are used or imported anywhere else** in the code. Nothing else imports the Atlas SDK.
- The translation layer is an **internal package**, not exposed to operator consumers.
  - Adding a translation layer does not affect the rest of the codebase util it is used.
- The translation layer allows for easy Contract Testing of each of the API endpoints encapsulated.
  - This enables easy Atlas behavior verification up front, even before the reconcilation state machine is coded.
- Other than that, unit testing of the type conversions, normalization and comparisons is also to be expected within this layer.

### All Submissions:

### DIFF
--- a/internal/translation/resource/fake/fake.go
+++ b/internal/translation/resource/fake/fake.go
@@ -1,0 +1,42 @@
+package fake
+
+import (
+	"context"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/resource"
+)
+
+type FakeResource struct {
+	GetResourceFn    func(ctx context.Context, id string) (*resource.Resource, error)
+	CreateResourceFn func(ctx context.Context, resource *resource.Resource) (*resource.Resource, error)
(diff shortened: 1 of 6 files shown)


josvazg:
Do not think we need this template anymore, we already have plenty of translation layers by different author merged in.

PR #1621, opened 2024-05-29
CLOUDP-250918: Add test to reproduce issue #1515

### LINKED ISSUE #1515: AtlasDatabaseUser - message - unable to list: test because of unknown namespace for the cache
Have a version of operator 1.7.1 and decide to upgrade to the latest in cluster. 
Create local env 
 
1.  k8s - by docker desktop v 1.25.4
2. operator v1.7.1 
3.  Add AtlasDeployment and AtlasDatabaseUser 
4.  Upgrade to v2.2.0 ( helm upgrade crd then upgrade operator ) 
5.  Fix AtlasDeployment 
6. Check logs of operator get error aka 

```{"level":"INFO","time":"2024-04-16T12:12:14.543Z","msg":"Status update","atlasdatabaseuser":"test/operator-upgrade-test","lastCondition":{"type":"DatabaseUserReady","status":"False","lastTransitionTime":null,"reason":"DatabaseUserStaleConnectionSecrets","message":"unable to list: test because of unknown namespace for the cache"}}```

**What did you expect?**
After all step operator just should work as expected

**What happened instead?**
AtlasDatabaseUser status always in False state 

**Operator Information**
 - 1.7.1 -> 2.2.0
 
 - Docker Desktop
 - 1.25.4

**Additional context**
Try to figure out why AtlasDatabaseUser CRD failed. 
It's created proper secrets and creates users in AtlasUI but CRD itself always in Ready - False state 

`status:
    conditions:
    - lastTransitionTime: "2024-04-16T12:03:17Z"
      status: "False"
      type: Ready
    - lastTransitionTime: "2024-04-16T11:44:08Z"
      status: "True"
      type: ResourceVersionIsValid
    - lastTransitionTime: "2024-04-16T11:44:08Z"
      status: "True"
      type: ValidationSucceeded
    - lastTransitionTime: "2024-04-16T12:03:18Z"
      message: 'unable to list: test because of unknown namespace for the cache'
      reason: DatabaseUserStaleConnectionSecrets
      status: "False"
      type: DatabaseUserReady`

If possible, please include:

`{"level":"DEBUG","time":"2024-04-16T12:17:12.709Z","msg":"Ensured connection Secret up-to-date","atlasdatabaseuser":"test/operator-upgrade-test","secretname":"HIDDEN"}
{"level":"INFO","time":"2024-04-16T12:17:12.709Z","msg":"Status update","atlasdatabaseuser":"test/operator-upgrade-test-","lastCondition":{"type":"DatabaseUserReady","status":"False","lastTransitionTime":null,"reason":"DatabaseUserStaleConnectionSecrets","message":"unable to list: test because of unknown namespace for the cache"}}
`

### PR DESCRIPTION
**Pull request by josvazg**
This is a bugfix for namespace watching filtering setup in both `main` and the `e2e` tests. The bug was discovered by issue #1515. The mistake came to be a mixture of controller runtime changing the manager setup and the operator code having a special case for watching a single namespace.

The fix consists of:
1. A new set of tests that detect the issue:
  - Test the cache gets, list and reconciles to access 3 namespaces when:
    - No namespaces were configured, meaning all 3 namespaces are accessible.
    - One namespace was configured, meaning only that namespace is accessible.
    - Two namespaces were configured, meaning two are accesible and the other is not.
2. The actual fix in `main` and `e2e` code:
  - Do not treat the single namespace as a special case.
  - Do not even get a "default namespace" in config at all.
  - Remove any default list of namespaces to watch, so that config sets it up clean.
  - Ensure the global predicate defaults to All Namespaces when no watched namespaces were setup.
  - Remove the `e2e` logic coming from a reflection on for the `WATCH_NAMESPACE` environment variable in `main`. It was not only useless but also introducing part of the bug in tests.

Note the reconcile test cases need to be run on separate processes because they are thread unsafe otherwise. See [CLOUDP-251767: Race conditions within e2e helper code](https://jira.mongodb.org/browse/CLOUDP-251767) for more info, but the gist of it is a AddToScheme being run on each test Kubernetes client initialisation, which clashes with the client event watcher.

Close detail ( T )
Cloud Services
Race conditions within e2e helper code
 
closes #1515 

### All Submissions:

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -102,7 +102,7 @@ func main() {
 	syncPeriod := time.Hour * 3
 
 	var cacheFunc cache.NewCacheFunc
-	if len(config.WatchedNamespaces) > 1 {
+	if len(config.WatchedNamespaces) > 0 {
 		var namespaces []string
 		for ns := range config.WatchedNamespaces {
 			namespaces = append(namespaces, ns)
@@ -124,8 +124,7 @@ func main() {
 			Port: 9443,
 		}),
 		Cache: cache.Options{
-			DefaultNamespaces: map[string]cache.Config{config.Namespace: {}},
-			SyncPeriod:        &syncPeriod,
+			SyncPeriod: &syncPeriod,
 		},
 		HealthProbeBindAddress: config.ProbeAddr,
 		LeaderElection:         config.EnableLeaderElection,
@@ -142,9 +141,10 @@ func main() {
 	// globalPredicates should be used for general controller Predicates
 	// that should be applied to all controllers in order to limit the
 	// resources they receive events for.
+	predicateNamespaces := controller.NamespacesForGlobalPredicate(config.WatchedNamespaces)
 	globalPredicates := []predicate.Predicate{
-		watch.CommonPredicates(),                                  // ignore spurious changes. status changes etc.
-		watch.SelectNamespacesPredicate(config.WatchedNamespaces), // select only desired namespaces
+		watch.CommonPredicates(),                             // ignore spurious changes. status changes etc.
+		watch.SelectNamespacesPredicate(predicateNamespaces), // select only desired namespaces
 	}
 
 	atlasProvider := atlas.NewProductionProvider(config.AtlasDomain, config.GlobalAPISecret, mgr.GetClient())
@@ -294,7 +294,6 @@ type Config struct {
 	AtlasDomain                 string
 	EnableLeaderElection        bool
 	MetricsAddr                 string
-	Namespace                   string
 	WatchedNamespaces           map[string]bool
 	ProbeAddr                   string
 	GlobalAPISecret             client.ObjectKey
@@ -336,14 +335,12 @@ func parseConfiguration() Config {
 	// dev note: we pass the watched namespace as the env variable to use the Kubernetes Downward API. Unfortunately
 	// there is no way to use it for container arguments
 	watchedNamespace := os.Getenv("WATCH_NAMESPACE")
-	config.WatchedNamespaces = make(map[string]bool)
-	for _, namespace := range strings.Split(watchedNamespace, ",") {
-		namespace = strings.TrimSpace(namespace)
-		config.WatchedNamespaces[namespace] = true
-	}
-
-	if len(config.WatchedNamespaces) == 1 {
-		config.Namespace = watchedNamespace
+	if strings.TrimSpace(watchedNamespace) != "" {
+		config.WatchedNamespaces = make(map[string]bool)
+		for _, namespace := range strings.Split(watchedNamespace, ",") {
+			namespace = strings.TrimSpace(namespace)
+			config.WatchedNamespaces[namespace] = true
+		}
 	}
 
 	configureDeletionProtection(&config)
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -141,7 +141,7 @@ func main() {
 	// globalPredicates should be used for general controller Predicates
 	// that should be applied to all controllers in order to limit the
 	// resources they receive events for.
-	predicateNamespaces := controller.NamespacesForGlobalPredicate(config.WatchedNamespaces)
+	predicateNamespaces := controller.NamespacesOrAllPredicate(config.WatchedNamespaces)
 	globalPredicates := []predicate.Predicate{
 		watch.CommonPredicates(),                             // ignore spurious changes. status changes etc.
 		watch.SelectNamespacesPredicate(predicateNamespaces), // select only desired namespaces
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -334,8 +334,8 @@ func parseConfiguration() Config {
 
 	// dev note: we pass the watched namespace as the env variable to use the Kubernetes Downward API. Unfortunately
 	// there is no way to use it for container arguments
-	watchedNamespace := os.Getenv("WATCH_NAMESPACE")
-	if strings.TrimSpace(watchedNamespace) != "" {
+	watchedNamespace := strings.TrimSpace(os.Getenv("WATCH_NAMESPACE"))
+	if watchedNamespace != "" {
 		config.WatchedNamespaces = make(map[string]bool)
 		for _, namespace := range strings.Split(watchedNamespace, ",") {
 			namespace = strings.TrimSpace(namespace)
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -28,3 +28,11 @@ func CustomLabelSelectorCacheBuilder(obj client.Object, labelsSelector labels.Se
 		return cache.New(config, opts)
 	}
 }
+
+func NamespacesForGlobalPredicate(namespaceMap map[string]bool) map[string]bool {
+	if len(namespaceMap) > 0 {
+		return namespaceMap
+	}
+	// if no namespaces where specified it must check all namespaces
+	return map[string]bool{cache.AllNamespaces: true}
+}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -29,7 +29,7 @@ func CustomLabelSelectorCacheBuilder(obj client.Object, labelsSelector labels.Se
 	}
 }
 
-func NamespacesForGlobalPredicate(namespaceMap map[string]bool) map[string]bool {
+func NamespacesOrAllPredicate(namespaceMap map[string]bool) map[string]bool {
 	if len(namespaceMap) > 0 {
 		return namespaceMap
 	}
--- a/test/e2e/cache_watch_test.go
+++ b/test/e2e/cache_watch_test.go
@@ -1,0 +1,308 @@
+package e2e_test
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: cmd/manager/main.go [hunk 1]
```diff
@@ -102,7 +102,7 @@ func main() {
 	syncPeriod := time.Hour * 3
 
 	var cacheFunc cache.NewCacheFunc
-	if len(config.WatchedNamespaces) > 1 {
+	if len(config.WatchedNamespaces) > 0 {
```
josvazg, line 105: I believe this was a different bug, it was only applying atched namespaces when more than 1 was set.

test/helper/e2e/k8s/operator.go
```diff
@@ -72,7 +72,7 @@ func BuildManager(initCfg *Config) (manager.Manager, error) {
 	logger.Info("starting manager", zap.Any("config", config))
 
 	var cacheFunc cache.NewCacheFunc
-	if len(config.WatchedNamespaces) > 1 {
+	if len(config.WatchedNamespaces) > 0 {
```
josvazg, line 75: Same error was duplicated in e2e tests. I think this was done on purpose and made sense probably on the previous version of controller runtime, before the manual migration that brought up the bug.

cmd/manager/main.go
josvazg: it is actually all related

pkg/controller/controller.go
```diff
@@ -28,3 +28,11 @@ func CustomLabelSelectorCacheBuilder(obj client.Object, labelsSelector labels.Se
 		return cache.New(config, opts)
 	}
 }
+
+func NamespacesForGlobalPredicate(namespaceMap map[string]bool) map[string]bool {
```
s-urbaniak, line 32:
```suggestion
func NamespacesOrAllPredicate(namespaceMap map[string]bool) map[string]bool {
```

s-urbaniak:
@josvazg i have too many suggestions here, let me submit a review commit instead.

s-urbaniak:
@josvazg I encoded my set of feedback into this commit on top of this PR: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1621/commits/d6d92d1f2de3e55c438151d9e9047e33aaaf256e

In my eyes this commit has a couple of benefits:

1. It dependency-injects the ginkgo `context.Context` everywhere (no `context.Background()`)
2. It establishes a concrete test case structure for the tabel driven ginkgo test, shared across all test cases
3. It encodes the fixture (what namespaces to create, what namespaces to watch, what namespaces we expect to be found) with concrete names rather than arrays of bools and numerical indexes which makes it easier for the reader to reason about the test fixtures and assertions. Note that the expectations here are encoding using "generate names" which are then mapped on the randomly generated names by kubernetes.
4. It replaces the `mgrHandle` with a simple `stop()` function, reducing the amount of types to be created here
5. it leverages the `GenerateName` feature of kubernetes so we don't have to randomly create names all over the place

test/e2e/cache_watch_test.go
```diff
@@ -21,32 +22,38 @@ import (
...
+	wantToFind      []string
+}
+
 var _ = Describe("Kubernetes cache watch test:", Label("cache-watch"), func() {
 	DescribeTable("Cache gets", Label("watch-gets"),
-		func(name string, namespaceIndexesToWatch []int, expectedNamespaceFound []bool) {
+		func(ctx context.Context, testCase *tc) {
 			testData := model.DataProvider(
-				fmt.Sprintf("cache-%s", name),
+				fmt.Sprintf("cache-%s", CurrentSpecReport().LeafNodeLabels[0]),
```
s-urbaniak, line 37: in case you're wondering what this returns: `CurrentSpecReport().LeafNodeLabels[0]` returns `"watch-gets"` in this case which is the leaf label used for this table driven test.

test/e2e/cache_watch_test.go
```diff
@@ -112,45 +145,47 @@ var _ = Describe("Kubernetes cache watch test:", Label("cache-watch"), func() {
...
-						project.Name = utils.RandomName(fmt.Sprintf("test-project-%s-ns%d", name, i))
-						project.Namespace = ns
+						project.Name = ""
+						project.GenerateName = "test-project-" + CurrentSpecReport().LeafNodeLabels[0]
+						project.Namespace = ns.GetName()
 						project.Spec.Name = "" // Set empty name to force validation error
-						projectNames[i] = project.Name
-
 						Expect(testData.K8SClient.Create(context.Background(), project)).ToNot(HaveOccurred())
+						projectNames[i] = project.Name
```
s-urbaniak, line 175: Note: we retrieve the project name after creation, after Kubernetes picking a random name that we initialised with `GenerateName` a couple of lines above.

test/e2e/cache_watch_test.go
```diff
@@ -112,45 +145,47 @@ var _ = Describe("Kubernetes cache watch test:", Label("cache-watch"), func() {
...
 					})
 
 					prj := &akov2.AtlasProject{}
 					By("Wait project to be present in Kubernetes", func() {
 						Eventually(func(g Gomega) bool {
 							return g.Expect(
-								testData.K8SClient.Get(testData.Context,
```
s-urbaniak: let's not continue using `testData.Context` (I removed it in my suggested commit). It doesn't allow layering of contexts which is a crucial feature to use (i.e. nested cancellation).

test/e2e/cache_watch_test.go
```diff
@@ -186,123 +221,140 @@ var _ = Describe("Reconciles test:", func() {
...
+		mgr manager.Manager
+	)
+
+	wantToWatchSet := sets.NewString(wantToWatch...)
+	watchedNamespaces := make(map[string]bool)
+	for _, ns := range namespaces {
+		if wantToWatchSet.Has(ns.GetGenerateName()) {
+			watchedNamespaces[ns.GetName()] = true
+		}
+	}
```
s-urbaniak, line 301: note: I am not convinced we need this here. I reimplemented it as it was there before but I think we can simply `managerConfig.WatchedNamespaces = wantToWatch` and call it a day. Feel free to remove this.

test/e2e/cache_watch_test.go
```diff
@@ -55,36 +62,50 @@ var _ = Describe("Kubernetes cache watch test:", Label("cache-watch"), func() {
...
-		Entry("Two namespaces when only those two are configured", Label("gets-two"),
-			"gets-two", []int{0, 1}, []bool{true, true, false}),
+
+		Entry("From all namespaces when no namespace config is set", Label("gets-all"), &tc{
+			namespaces:      []string{"ns1", "ns2", "ns3"},
+			watchNamespaces: nil,
+			wantToFind:      []string{"ns1", "ns2", "ns3"},
+		}),
+
+		Entry("One namespace when only one namespace is configured", Label("gets-all"), &tc{
```
josvazg, line 72: This should be labeled `gets-one`

test/e2e/cache_watch_test.go
```diff
@@ -55,36 +62,50 @@ var _ = Describe("Kubernetes cache watch test:", Label("cache-watch"), func() {
...
+			namespaces:      []string{"ns1", "ns2", "ns3"},
+			watchNamespaces: []string{"ns1"},
+			wantToFind:      []string{"ns1"},
+		}),
+
+		Entry("Two namespaces when only those two are configured", Label("gets-all"), &tc{
```
josvazg, line 78: And this `gets-two`

josvazg:
> @josvazg I encoded my set of feedback into this commit on top of this PR: [d9ead9f](https://github.com/mongodb/mongodb-atlas-kubernetes/commit/d9ead9f112760222782855d21c31eb155e5540a6)
> 
> In my eyes this commit has a couple of benefits:
> 
> 1. It dependency-injects the ginkgo `context.Context` everywhere (no `context.Background()`) and does not use the context in the testData struct. The former is not an optimal construct as it does not allow you to layer contexts in the call stack.
> 2. It establishes a concrete test case structure for the tabel driven ginkgo test, shared across all test cases
> 3. It encodes the fixture (what namespaces to create, what namespaces to watch, what namespaces we expect to be found) with concrete names rather than arrays of bools and numerical indexes which makes it easier for the reader to reason about the test fixtures and assertions. Note that the expectations here are encoding using "generate names" which are then mapped on the randomly generated names by kubernetes.
> 4. It replaces the `mgrHandle` with a simple `stop()` function, reducing the amount of types to be created here
> 5. it leverages the `GenerateName` feature of kubernetes so we don't have to randomly create names all over the place

Thanks, this looks better and more readable!

test/e2e/cache_watch_test.go
s-urbaniak: he, copy-pasta, do you mind to fix?

test/e2e/cache_watch_test.go
s-urbaniak: indeed, copy-pasta as well, do you mind to fix?

cmd/manager/main.go
```diff
@@ -336,14 +335,12 @@ func parseConfiguration() Config {
...
 	watchedNamespace := os.Getenv("WATCH_NAMESPACE")
-	config.WatchedNamespaces = make(map[string]bool)
-	for _, namespace := range strings.Split(watchedNamespace, ",") {
-		namespace = strings.TrimSpace(namespace)
-		config.WatchedNamespaces[namespace] = true
-	}
-
-	if len(config.WatchedNamespaces) == 1 {
-		config.Namespace = watchedNamespace
+	if strings.TrimSpace(watchedNamespace) != "" {
```
s-urbaniak, line 338: very nice sanitization 👍

s-urbaniak: maybe we can put the trimming in line 338, wdyt?

s-urbaniak: (this being just a super small nit to be ignored at will)

s-urbaniak (approved):

test/e2e/cache_watch_test.go
josvazg: Done

test/e2e/cache_watch_test.go
josvazg: Done

helderjs (approved):

PR #1631, opened 2024-06-05
Upgrade to Go 1.22.4 to fix security vulnerability

josvazg:
### All Submissions:


go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/mongodb/mongodb-atlas-kubernetes/v2
 
 go 1.22.0
 
-toolchain go1.22.3
+toolchain go1.22.4
```
helderjs, line 5: What about Dockerfile? why not include the tooling in the update?

helderjs (changes requested):

go.mod
helderjs: this is one more reason to do, no? We "skip" the magic and ensure tooling explicitly reflects what it should be

josvazg: I think this magic is done on purpose by the Go team. Basically means you do not need to wait for whatever distribution channel to get the latest go tool chain version. You do not need to wait for the docker container with go 1.22.4 to be ready, you do not need the GitHub CI ubuntu image to have the latest go tool chain ready, your nix packages or your internal CI system to be up to date.

josvazg: I updated the tools' toolchain the same way.

helderjs (approved):

PR #1633, opened 2024-06-07
CLOUDP-251873: Fix 2.3.0 SBOMs and generation script

josvazg:
This fixes a bug in the SBOM generation when the host machine does not match the arch from the `docker sbom` requested. Eg:

`image has unexpected architecture "arm64", which differs from the user specified architecture "amd64"`

The only workaround I found was to be explicit about the platform image SHA, so it will not complain and just compute the SBOM properly.

This PR also fixed version 2.3.0 SBOMs affected by this issue earlier.

### All Submissions:


josvazg:
CLOUDP-251873

helderjs (approved):

slaskawi (commented):
LGTM!

PR #1651, opened 2024-06-13
CLOUDP-228589: Migrate AuditLog to new Atlas SDK

helderjs:
### All Submissions:

### DIFF
--- a/internal/mocks/translation/audit.go
+++ b/internal/mocks/translation/audit.go
@@ -1,0 +1,19 @@
+package translation
+
+import (
+	"context"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/audit"
+)
+
+type AuditLogMock struct {
+	GetFunc func(projectID string) (*audit.AuditConfig, error)
+	SetFunc func(projectID string, auditing *audit.AuditConfig) error
+}
+
+func (c *AuditLogMock) Get(_ context.Context, projectID string) (*audit.AuditConfig, error) {
+	return c.GetFunc(projectID)
+}
+func (c *AuditLogMock) Set(_ context.Context, projectID string, auditing *audit.AuditConfig) error {
+	return c.SetFunc(projectID, auditing)
+}
--- a/internal/translation/audit/audit.go
+++ b/internal/translation/audit/audit.go
@@ -5,11 +5,6 @@ import (
 	"fmt"
 
 	"go.mongodb.org/atlas-sdk/v20231115008/admin"
-	"go.uber.org/zap"
-	"k8s.io/apimachinery/pkg/types"
-
-	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation"
-	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
 )
 
 // AuditLogService is the interface exposed by this translation layer over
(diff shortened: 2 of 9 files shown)


pkg/controller/workflow/reason.go
```diff
@@ -38,7 +38,7 @@ const (
 	ProjectNetworkPeerIsNotReadyInAtlas        ConditionReason = "ProjectNetworkPeerIsNotReadyInAtlas"
 	ProjectEncryptionAtRestReady               ConditionReason = "ProjectEncryptionAtRestReady"
 	ProjectCloudIntegrationsIsNotReadyInAtlas  ConditionReason = "ProjectCloudIntegrationsIsNotReadyInAtlas"
-	ProjectAuditingReady                       ConditionReason = "ProjectAuditingReady"
+	ProjectAuditLogsNotConfigured              ConditionReason = "ProjectAuditLogsNotConfigured"
```
s-urbaniak, line 41: does this end up in status conditions? If so, this is effectively an API change.

s-urbaniak:
one concern regarding status conditions, else great! 👍

internal/mocks/translation/audit.go
```diff
@@ -0,0 +1,19 @@
+package translation
```
josvazg, line 1: In other PR (or PRs) I put thin under the fake folder.

We should be consistent, so lets ask for others votes here:
@s-urbaniak &  @roothorp WDYT?

`internal/resource/fake`
OR
`internal/mocks/translation` (all mocks in a single package it seems)

internal/translation/audit/audit.go
```diff
@@ -24,15 +19,6 @@ type AuditLog struct {
 	auditAPI admin.AuditingApi
 }
 
-// NewAuditLogService creates an AuditLog from credentials and the atlas provider
```
josvazg, line 27: If we do this everywhere we might be able to remove the code at the `translation` parent package.

Maybe we should pin a reminder for that in a Jira. But it does not affect this PR.

pkg/controller/atlasproject/auditing.go
```diff
@@ -3,96 +3,87 @@ package atlasproject
...
-	}
-
-	if isAuditingEmpty(project.Spec.Auditing) {
-		workflowCtx.UnsetCondition(api.AuditingReadyType)
-		return workflow.OK()
-	}
-
-	workflowCtx.SetConditionTrue(api.AuditingReadyType)
-	return workflow.OK()
+type auditController struct {
```
josvazg, line 12: Great!

pkg/controller/workflow/reason.go
helderjs: It's not a new status, just a more meaningful reason. Is it still a breaking change?

josvazg: +1

josvazg:
LGTM, waiting on that API change issue.

About the mocks voting, we can merge and fix afterwards if needed, whatever the decision is.

pkg/controller/workflow/reason.go
s-urbaniak: This is removing `ProjectAuditingReady` in favor of `ProjectAuditLogsNotConfigured`.
If users rely on `ProjectAuditingReady` in their GitOps pipeline, they will be broken after this change.
Hence, we cannot simply remove reasons from status conditions, they are meant to be machine-readable.
`message` though can be improved.

internal/mocks/translation/audit.go
s-urbaniak: this is a nuance, leaving the final call to you @josvazg and @helderjs

pkg/controller/atlasproject/auditing.go
s-urbaniak: indeed, this is a slick, small, and neat little state machine done here 😍

pkg/controller/atlasproject/auditing.go
```diff
@@ -3,96 +3,87 @@ package atlasproject
...
+		return a.configure(auditInAKO, isUnset)
+	}
 
-func prepareAuditingSpec(spec *akov2.Auditing) *mongodbatlas.Auditing {
-	if isAuditingEmpty(spec) {
-		return &mongodbatlas.Auditing{
-			Enabled: pointer.MakePtr(false),
-		}
+	if a.project.Spec.Auditing == nil {
+		return a.unmanage()
```
s-urbaniak, line 33: nit: this transition is not handled in unit tests.

s-urbaniak (changes requested):
two concerns:
- missing unit test controller
- when I execute the contract test locally, I actually get a JSON_FILTER config type by default

pkg/controller/atlasproject/auditing.go
helderjs: Added

s-urbaniak (approved):

PR #1667, opened 2024-06-28
CLOUDP-227273: Translation layer on Project IP Access List + state machine

helderjs:
### All Submissions:


internal/translation/ipaccesslist/conversion.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (i *IPAccessEntry) ID() string {
+	if i.CIDR != "" {
+		return i.CIDR
+	}
+
+	return i.AWSSecurityGroup
+}
+
+func (i *IPAccessEntry) IsExpired() bool {
```
s-urbaniak, line 30: for better testability:
```suggestion
func (i *IPAccessEntry) IsExpired(now time.Time) bool {
```

internal/translation/ipaccesslist/conversion.go
```diff
@@ -0,0 +1,192 @@
...
+			Mask: net.IPv4Mask(255, 255, 255, 255),
+		}
+		return ipNet.String(), nil
+	}
+
+	if cidr != "" {
+		var err error
+		_, parsedNet, err := net.ParseCIDR(cidr)
+		if err != nil {
+			return "", fmt.Errorf("cidr %s is invalid", cidr)
```
s-urbaniak, line 185: return the error as well

s-urbaniak (approved):
this is fantastic 👏

internal/translation/ipaccesslist/conversion.go
```diff
@@ -0,0 +1,192 @@
...
+	for ix, entry := range i {
+		if !entry.IsExpired() {
+			entries[ix] = entry
+		}
+	}
+
+	return entries
+}
+
+func (i IPAccessEntries) GetExpired() IPAccessEntries {
```
s-urbaniak, line 51:
```suggestion
func (i IPAccessEntries) ByExpired(expired bool) IPAccessEntries {
```

internal/translation/ipaccesslist/conversion.go
```diff
@@ -0,0 +1,192 @@
...
+	if i.DeleteAfterDate == nil {
+		return false
+	}
+
+	return i.DeleteAfterDate.Before(time.Now())
+}
+
+type IPAccessEntries map[string]*IPAccessEntry
+
+func (i IPAccessEntries) GetActives() IPAccessEntries {
```
s-urbaniak, line 40: this method can be removed in favor of a single `ByExpired(expired bool)` method (see below)

internal/translation/ipaccesslist/conversion.go
```diff
@@ -0,0 +1,192 @@
...
+	for ix, entry := range i {
+		if entry.IsExpired() {
+			entries[ix] = entry
+		}
+	}
+
+	return entries
+}
+
+func ToAKO(ipAccessEntries IPAccessEntries) []project.IPAccessList {
```
s-urbaniak, line 62: I believe the convention elsewhere was `fromAtlas` and `toAtlas`. Let's stick to a common scheme.

s-urbaniak (changes requested):
a couple of nits, else lgtm

internal/translation/ipaccesslist/conversion.go
s-urbaniak: `func FromInternal(ipAccessEntries IPAccessEntries) []project.IPAccessList {`

s-urbaniak: @helderjs do you mind to add the "source signifier" convention into the translation layer proposal, including the rationale (aka the "why").

internal/translation/ipaccesslist/conversion.go
josvazg: Nit pick I would call is `isExpired(at time.Time)` were usually it **at** is **now**: `.IsExpired(time.now())`

internal/translation/ipaccesslist/conversion.go
josvazg: I see the value on the boolean arg, but the `GetByExpired` name sounds weird to me. Not sure what that means. `GetByStatus(expired bool)` seems clearer.

internal/translation/ipaccesslist/conversion.go
josvazg: Yes, I thought we would only use 'toAtlas` and `fromAtlas` everywhere.

We also intend to overuse the AKO acronym in our code a lot.

pkg/controller/atlasproject/ipaccess_list.go
```diff
@@ -1,207 +1,133 @@
...
+			}
+		}
 	}
 
-	desiredMap := map[string]project.IPAccessList{}
-	for _, item := range desired {
-		desiredMap[genIPAccessListKey(item)] = item
+	if isUnset {
+		return i.unmanage()
 	}
```
josvazg, line 65: I find it confusing this `isUnset -> unmanage` is both here and in the main `reconcile` body. Can you elaborate on why is that needed?

josvazg (commented):
LGTM overall. Just a question and no additional concerns apart from those by @s-urbaniak

pkg/controller/atlasproject/ipaccess_list.go
helderjs: This ensures status conditions are removed when config is removed from the manifest

internal/translation/ipaccesslist/conversion.go
```diff
@@ -0,0 +1,182 @@
...
+	for ix, entry := range i {
+		if entry.IsExpired(time.Now()) == expired {
+			entries[ix] = entry
+		}
+	}
+
+	return entries
+}
+
+func FromInternal(ipAccessEntries IPAccessEntries) []project.IPAccessList {
```
josvazg, line 52: I was not understanding why we needed this, but then I saw we are changing the shape of the IP access list from an array to a set. That is fine, I guess it simplifies something later like checking for no duplicates.

Still most other "internal" types augment information that was not there in the spec, such as the project id or a password value.

To avoid the need for this `FromInternal` you could use the composition trick again:

```go
type IPAccessEntries struct {
  List []project.IPAccessList
  Set map[string]*IPAccessEntry
}
```

When passing the type you just need to use the `.List` instead of calling `FromInternal`. It is the original list, so no bugs possible.

josvazg: I saw later that this is used after a filtering call, so the trick will not work.

josvazg (approved):

s-urbaniak (approved):

PR #1675, opened 2024-07-03
[DO NOT MERGE] Audit conversion tests + method refactor

filipcirtog:
### All Submissions:

### DIFF
--- a/internal/translation/audit/conversion_test.go
+++ b/internal/translation/audit/conversion_test.go
@@ -100,23 +100,23 @@ func TestConversionContructor(t *testing.T) {
 	}
 }
 
-func TestToAtlas(t *testing.T) {
+func TestToAndFromAtlas(t *testing.T) {
 	TrueBool := true
 	FalseBool := false
 	EmptyAudit := "{}"
 	testCases := []struct {
 		title          string
-		input          *AuditConfig
-		expectedOutput *admin.AuditLog
+		kubernetesSide *AuditConfig
+		atlasSide      *admin.AuditLog
 	}{
 		{
 			title: "Just enabled",
-			input: NewAuditConfig(
+			kubernetesSide: NewAuditConfig(
 				&akov2.Auditing{
 					Enabled: true,
 				},
 			),
-			expectedOutput: &admin.AuditLog{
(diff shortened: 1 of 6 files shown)


internal/translation/audit/conversion_test.go
```diff
@@ -0,0 +1,220 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+)
+
+func TestConversionContructor(t *testing.T) {
```
josvazg, line 12:
```suggestion
func TestConversionConstructor(t *testing.T) {
```

Or maybe even better:

```suggestion
func TestNewAuditConfig(t *testing.T) {
```

internal/translation/audit/conversion_test.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	for _, tc := range testCases {
+		t.Run(tc.title, func(t *testing.T) {
+			actualResult := NewAuditConfig(tc.input)
+			assert.Equal(t, tc.expectedOutput, actualResult)
+		})
+	}
+}
+
+func TestToAndFromAtlas(t *testing.T) {
```
josvazg, line 103: This seems like a weird mix, I would just have a `TestToAtlas` and a separate `TestFromAtlas` with only internal types or atlas types respectively.

Alternatively one thing you could do is to have a single `TestToAndFromAtlas` table with only the internal type as inputs, then in the test you do `ToAtlas` and the result you pass to `FromAtlas` and finally compare the final result with the original input, which should match.

josvazg (commented):
Looking good, just a comment on some test table that seems a bit mixed up

josvazg:
Please rebase on today's latest main to avoid the Go vulnerability issue in the linter:
https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/9777360588/job/26991884113

filipcirtog:
A new PR was created for the same issue.

PR #1684, opened 2024-07-10
CLOUDP-260438: Fix redacted integration comparisons

josvazg:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasproject/integrations.go
+++ b/pkg/controller/atlasproject/integrations.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"reflect"
 
 	"go.mongodb.org/atlas/mongodbatlas"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/atlasproject/integrations.go
```diff
@@ -148,41 +147,21 @@ func (r *AtlasProjectReconciler) checkIntegrationsReady(ctx *workflow.Context, n
...
-	}
-
-	charCount := 4
-	str := *strPtr
-	if len(str) <= charCount {
-		return
-	}
-
-	*strPtr = str[len(str)-charCount:]
+func areIntegrationsEqual(_, _ *aliasThirdPartyIntegration) bool {
```
slaskawi, line 159: It's probably a matter of taste but I would extract this into an interface and provide an "AlwaysApplyComparer" or something along these lines. Just to get this stop-the-bleeding patch code out of the sight. 

Feel free to ignore this suggestion if you think otherwise.

josvazg: Not sure is worth it. We are going to actually write quite some new code once we need to actually compare. We just do now know what that code is now, so having interfaces seems premature. For instance, if we go for implicit evaluation these interfaces break, because we need input from 3 sources, Atlas, Spec and annotations.

s-urbaniak (approved):
lgtm but I'd like to leave a final review to @helderjs and/or @igor-karpukhin as well.

igor-karpukhin (approved):

helderjs (approved):

PR #1702, opened 2024-07-17
CLOUDP-228591: Migrate Custom Roles to SDK

roothorp:
(also add translation layer)

### All Submissions:


roothorp:
WRT Linter error, from the [gosec README](https://github.com/securego/gosec?tab=readme-ov-file#available-rules):
> G601: Implicit memory aliasing of items from a range statement (only for Go 1.21 or lower)

I'm pretty sure we can therefore ignore this (with `//nosec`), but it suggests that gosec and/or golangci hasn't been updated.

filipcirtog (commented):
Looks great to me! Combining the implementation of the translation layer with the switch from the deprecated go-client to the SDK in a single PR was a clever and efficient move!

pkg/controller/atlasproject/custom_roles.go
```diff
@@ -6,27 +6,44 @@ import (
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/customroles"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api"
 	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1/status"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/workflow"
 )
 
+type roleController struct {
+	ctx     *workflow.Context
+	project *akov2.AtlasProject // TODO: Should this be a translation layer type?
```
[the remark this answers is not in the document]
roothorp: Good catch!

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/pointer"
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+)
+
+type CustomRole struct {
+	*akov2.CustomRole
+}
```
josvazg, line 11: Why wrapping if we are not adding anything?

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,127 @@
...
+				Db:         admin.GetOrDefault(resource.Database, ""),
+				Cluster:    admin.GetOrDefault(resource.Cluster, false),
+			})
+		}
+		actions = append(actions, admin.DatabasePrivilegeAction{
+			Action:    action.Name,
+			Resources: &resources,
+		})
+	}
+	atlas.SetActions(actions)
```
josvazg, line 42: Nitpick, I would move this to a sub-function, but maybe that is just me:

```suggestion
	atlas.SetActions(toAtlasActions(role.Actions))
```

and separately:
```go
func toAtlasActions(actions []akov2.Action) *admin.UserCustomDBRole {
actions := make([]admin.DatabasePrivilegeAction, 0, len(role.Actions))
	for _, action := range role.Actions {
		resources := make([]admin.DatabasePermittedNamespaceResource, 0, len(action.Resources))
		for _, resource := range action.Resources {
			if resource.Cluster != nil && !*resource.Cluster {
				resource.Cluster = nil
			}
			resources = append(resources, admin.DatabasePermittedNamespaceResource{
				Collection: admin.GetOrDefault(resource.Collection, ""),
				Db:         admin.GetOrDefault(resource.Database, ""),
				Cluster:    admin.GetOrDefault(resource.Cluster, false),
			})
		}
		actions = append(actions, admin.DatabasePrivilegeAction{
			Action:    action.Name,
			Resources: &resources,
		})
	}
	return actions
}
```

But again, that is a nitpick and maybe personal preference.

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,127 @@
...
+
+	inheritedRoles := make([]admin.DatabaseInheritedRole, 0, len(role.InheritedRoles))
```
josvazg, line 44: Same nitpick for a separate `toAtlasInheritedRoles(role.InheritedRoles)`

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	atlas.SetInheritedRoles(inheritedRoles)
+
+	return atlas
+}
+
+func toAtlasUpdate(role *CustomRole) *admin.UpdateCustomDBRole {
+	atlas := admin.NewUpdateCustomDBRoleWithDefaults()
+
+	actions := make([]admin.DatabasePrivilegeAction, 0, len(role.Actions))
```
josvazg, line 59: Same nitpick for separate `fromAtlasActions(role.Actions)`

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	atlas.SetInheritedRoles(inheritedRoles)
+
+	return atlas
+}
+
+func fromAtlas(role *admin.UserCustomDBRole) *CustomRole {
+	inheritedRoles := make([]akov2.Role, 0, len(*role.InheritedRoles))
+
+	for _, atlasInheritedRole := range *role.InheritedRoles {
```
josvazg, line 94: Same nitpick for separate `fromAtlasActions/InheritedRoles`

.mockery.yaml
```diff
@@ -9,3 +9,4 @@ packages:
   github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/ipaccesslist:
   github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/dbuser:
   github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/deployment:
+  github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/customroles:
```
josvazg, line 12: The mocks are missing, you will need to run mockery manually for now, we have no integration in the Makefile for that yet.

internal/translation/customroles/custom_roles.go
```diff
@@ -0,0 +1,63 @@
...
+func (s *CustomRoles) List(ctx context.Context, projectID string) ([]CustomRole, error) {
+	atlasRoles, _, err := s.roleAPI.ListCustomDatabaseRoles(ctx, projectID).Execute()
+	if err != nil {
+		return nil, fmt.Errorf("failed to list custom roles from Atlas: %w", err)
+	}
+
+	customRoles := make([]CustomRole, len(atlasRoles))
+
+	for i, r := range atlasRoles {
+		customRoles[i] = *fromAtlas(&r)
```
josvazg, line 44: I would avoid the struct copy operation and also avoid the linter issue at the same time probably.

```suggestion
	customRoles := make([]*CustomRole, len(atlasRoles))

	for i, r := range atlasRoles {
		customRoles[i] = fromAtlas(&r)
```

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,127 @@
...
+			Db:   inheritedRole.Database,
+			Role: inheritedRole.Name,
+		})
+	}
+	atlas.SetInheritedRoles(inheritedRoles)
+
+	return atlas
+}
+
+func fromAtlas(role *admin.UserCustomDBRole) *CustomRole {
```
josvazg, line 91: Given how it is used later maybe it is better to return a by value directly:
```suggestion
func fromAtlas(role *admin.UserCustomDBRole) CustomRole {
```

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,127 @@
...
+		})
+	}
+
+	return &CustomRole{
+		CustomRole: &akov2.CustomRole{
+			Name:           role.RoleName,
+			InheritedRoles: inheritedRoles,
+			Actions:        actions,
+		},
+	}
```
josvazg, line 126: By value:

```suggestion
	return CustomRole{
		CustomRole: &akov2.CustomRole{
			Name:           role.RoleName,
			InheritedRoles: inheritedRoles,
			Actions:        actions,
		},
	}
```

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,127 @@
...
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+)
+
+type CustomRole struct {
+	*akov2.CustomRole
+}
+
+func NewCustomRole(role *akov2.CustomRole) *CustomRole {
+	return &CustomRole{
```
josvazg, line 14:
```suggestion
func NewCustomRole(role *akov2.CustomRole) CustomRole {
	return CustomRole{
```

josvazg (commented):
LGTM over all, just check out those returns by value. They will probably fix things including the linter.
Nitpicks not important, just style

internal/translation/customroles/conversion.go
roothorp: TBH I was just following what was done in the Auditing translation layer. I suppose this gives us flexibility to add more fields in future if we ever see the need? I can change to use the akov2 type if preferred.

internal/translation/customroles/conversion.go
roothorp: I've separated these out into their own functions.

internal/translation/customroles/conversion.go
josvazg: Ok, let's leave it for regularity and expandability then.

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,108 @@
...
+	return atlas
+}
+
+func toAtlasUpdate(role *CustomRole) *admin.UpdateCustomDBRole {
+	atlas := admin.NewUpdateCustomDBRoleWithDefaults()
+	atlas.SetActions(toAtlasActions(role))
+	atlas.SetInheritedRoles(toAtlasInheritedRoles(role))
+
+	return atlas
+}
```
josvazg, line 35: Looking neat now! 🎉

internal/translation/customroles/conversion.go
```diff
@@ -0,0 +1,108 @@
+package customroles
```
josvazg, line 1: Conversions can and should probably be fully unit tested.

pkg/controller/atlasproject/custom_roles_test.go
```diff
@@ -1,210 +1,194 @@
...
+			roleAPI: func() *mockadmin.CustomDatabaseRolesApi {
+				roleAPI := mockadmin.NewCustomDatabaseRolesApi(t)
+				roleAPI.EXPECT().ListCustomDatabaseRoles(context.Background(), "").
+					Return(admin.ListCustomDatabaseRolesApiRequest{ApiService: roleAPI})
+				roleAPI.EXPECT().ListCustomDatabaseRolesExecute(mock.Anything).
+					Return(
+						[]admin.UserCustomDBRole{},
+						&http.Response{},
+						nil,
+					)
```
josvazg, line 68: These mocks are on the SDK library, which is fine here. But if we had added contract testing, which the translation layer enables, these would be a bit redundant and we could mock at the translation layer interface itself `List` instead of `ListCustomDatabaseRoles`.

BTW, I see we added the mockery mock support for the custom role service, are we using it at all here?
If not I would remove until we do (YAGNI)

pkg/controller/atlasproject/custom_roles_test.go
```diff
@@ -1,210 +1,194 @@
...
 import (
+	"context"
+	"net/http"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"go.uber.org/zap"
+	"github.com/stretchr/testify/mock"
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+	"go.mongodb.org/atlas-sdk/v20231115008/mockadmin"
```
josvazg, line 11: @roothorp do you think we should go already with the 2024 version here?
Or it might be better to complete the migration on the same one everywhere and then sweep upgrade in one or a few PRs.
cc @s-urbaniak

s-urbaniak: i suggest to sync this with the DRI @helderjs

s-urbaniak (approved):
looks great on green 👍

pkg/controller/atlasproject/custom_roles_test.go
helderjs: Latest version contains significant changes and migration requires a separate PR

Issue #1689, opened 2024-07-12
AtlasDatabaseUser is not reconciled periodically - manual changes are not overridden until pod restart


legal90:
**What did you do to encounter the bug?**

1. Create any `AtlasDatabaseUser` resource
2. Wait for the operator controller to reconcile/create a new database user. Log example:
  > {"level":"INFO","time":"2024-07-12T14:12:13.215Z","msg":"Status update","atlasdatabaseuser":"my-namespace/my-dbuser","lastCondition":{"type":"Ready","status":"True","lastTransitionTime":null}}
3. Do manual changes to that user via Atlas Web UI. For example, update the roles or just delete that user.

Now the declarative configuration in k8s, `AtlasDatabaseUser` , does not match the real state of the object in Atlas.

**What did you expect?**
The controller should periodically reconcile the object which it controls via `AtlasDatabaseUser` resource - 
it should read the actual state from Atlas API, compare it with the desired state declared in `AtlasDatabaseUser` resource and reconcile it - update the 

**What happened instead?**
Operator does nothing to reconcile the DB user in Atlas - changes made manually in UI are never overridden. I waited for > 4 hours - no reaction. There are no any new log message in the controller's log, which points to the fact that operator doesn't re-read the object state from API.

The reconciliation only happens when you **restart the controller pod** or update the `AtlasDatabaseUser` resource in k8s. 

**Operator Information**
 - Operator Version: 2.3.1

**Kubernetes Cluster Information**
 - Distribution: GKE
 - Version v1.29.1-gke.1589018

**Additional context**
YAML definition:
```
apiVersion: atlas.mongodb.com/v1
kind: AtlasDatabaseUser
metadata:
  labels:
  name: my-dbuser
  namespace: my-namespace
spec:
  databaseName: admin
  passwordSecretRef:
    name: my-dbuser
  projectRef:
    name: my-atlas-project
    namespace: my-namespace
  roles:
    - databaseName: admin
      roleName: atlasAdmin
  username: dbUserTest
```

josvazg:
This is expected up to a point. The default sync period in the Operator is 3 hours, so you should have seen the change after that time.

What happens here is the Operator thinks it has done its work, the Atlas state and the Kubernetes definition matches and so the next check will be in 3 hours.

Note there is no way for the operator to know the UI or any other agent with access to Altas made some changes. It can only learn about next time it tries to reconcile.

[...]

PR #1707, opened 2024-07-18
CLOUDP-261715: Enabled resync for Atlas custom resources

igor-karpukhin:
### All Submissions:

* [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if there is one). Closes https://github.com/mongodb/mongodb-atlas-kubernetes/issues/1689

### DIFF
--- a/test/int/syncperiod_test.go
+++ b/test/int/syncperiod_test.go
@@ -86,7 +86,7 @@ var _ = Describe("Sync Period test", Label("int", "sync-period"), func() {
 			time.Sleep(syncInterval * 2)
 		})
 
-		By("Project generation should be different", func() {
+		By("Project resource version should be different", func() {
 			var currentResourceVersion string
 			Eventually(func(g Gomega) bool {
 				if !resources.ReadAtlasResource(context.Background(), k8sClient, createdProject) {
--- a/test/int/syncperiod_test.go
+++ b/test/int/syncperiod_test.go
@@ -1,0 +1,102 @@
+package int
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/kube"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api"
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/test/helper/conditions"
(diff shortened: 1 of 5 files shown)


s-urbaniak (changes requested):
This is not enough:
1. Please add a resource version predicate on all Secret watches. Otherwise we will have double reconciliations in the face of resyncs.
2. Please make the resync interval configurable in the builder and add an e2e test.

s-urbaniak:
re 2.: for now, we shouldn't expose the knob to main just yet but just for the e2e test.

igor-karpukhin:
@s-urbaniak the sync interval is already configurable. The main just doesn't set it and it's set up to the default value of 3 hours

igor-karpukhin:
> Otherwise we will have double reconciliations in the face of resyncs.
I just checked that locally, and it doesn't double the reconciliations. It doubles them if we actually remove the predicated

test/int/syncperiod_test.go
```diff
@@ -0,0 +1,102 @@
...
+				previousResourceVersion = createdProject.ResourceVersion
+				return true
+			}).WithTimeout(10 * time.Second).WithPolling(2 * time.Second).Should(BeTrue())
+		})
+
+		By(fmt.Sprintf("Should wait for at least %f seconds", (syncInterval*2).Seconds()), func() {
+			time.Sleep(syncInterval * 2)
+		})
+
+		By("Project generation should be different", func() {
```
s-urbaniak, line 89:
```suggestion
		By("Project resource version should be different", func() {
```

s-urbaniak (commented):
one nit, else lgtm 👏

test/int/syncperiod_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api"
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/test/helper/conditions"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/test/helper/events"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/test/helper/resources"
+)
+
+var _ = Describe("Sync Period test", Label("int", "sync-period"), func() {
+	const interval = time.Second * 2
+	const syncInterval = 40 * time.Second
```
josvazg, line 22: nit pick
would 10s be enough to test the same?

josvazg (approved):

test/int/syncperiod_test.go
s-urbaniak: As a side note: i am working on a pure unit-test approach for this. I don't think we need a full blown e2e test to assert resyncs. I will submit once ready.

s-urbaniak (approved):

PR #1708, opened 2024-07-22
CLOUDP-244167: atlasdatabaseuser: use indexer instead of resourcewatcher

s-urbaniak:
### All Submissions:


josvazg (approved):
LGTM, ok to merge when CI is also green

pkg/indexer/atlasdatabaseusersecrets.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"go.uber.org/zap"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+)
+
+const (
+	AtlasDatabaseUserBySecretsIndex = "atlasdatabaseuser.spec.passwordSecret"
```
helderjs, line 11: dbUser belongs to a project, should we add an index for it too?
Deployments would benefit it when creating connection secret, it goes through all DBUsers to match the ones that belong to the same project in order to create the respective connection secret

helderjs (commented):
Overall LGTM, just raised a question once index for DBUser is being introduced.

helderjs (approved):

pkg/indexer/atlasdatabaseusersecrets.go
s-urbaniak: yes, that makes sense. the database user controller reads the project so we should add an index on it and let the atlas database user reconciler watch projects as well.

This is not part of this PR though, this mechanically replaces resource watcher with indexers.

PR #1715, opened 2024-07-23
CLOUDP-261435: Broken GOV tests - change AWS region

igor-karpukhin:
### All Submissions:

### DIFF
--- a/pkg/controller/connectionsecret/connectionsecrets.go
+++ b/pkg/controller/connectionsecret/connectionsecrets.go
@@ -196,13 +196,12 @@ func FillPrivateConnStrings(connStrings *mongodbatlas.ConnectionStrings, data *C
 }
 
 func GetAllServerless(ctx *workflow.Context, projectID string) ([]*mongodbatlas.Cluster, error) {
+	if IsCloudGovDomain(ctx) {
+		return make([]*mongodbatlas.Cluster, 0), nil
+	}
 	serverless, _, err := ctx.Client.ServerlessInstances.List(ctx.Context, projectID, nil)
 	if err != nil {
-		if !IsCloudGovDomain(ctx) {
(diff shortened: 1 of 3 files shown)


pkg/controller/connectionsecret/connectionsecrets.go
```diff
@@ -215,6 +214,7 @@ func IsCloudGovDomain(ctx *workflow.Context) bool {
 		"cloud-qa.mongodbgov.com",
 	}
 
+	fmt.Printf("DBG: DOMAIN: %s", ctx.Client.BaseURL.Host)
```
s-urbaniak, line 217: let's remove this

pkg/controller/connectionsecret/connectionsecrets.go
```diff
@@ -196,13 +196,12 @@ func FillPrivateConnStrings(connStrings *mongodbatlas.ConnectionStrings, data *C
 }
 
 func GetAllServerless(ctx *workflow.Context, projectID string) ([]*mongodbatlas.Cluster, error) {
+	if IsCloudGovDomain(ctx) {
+		return make([]*mongodbatlas.Cluster, 0), nil
+	}
```
s-urbaniak, line 201: is this a separate issue? I think this solves a separate issue, namely that serverless is not supported at all in cloud gov. Hence, I would like to see this in a separate contribution with a separate ticket.

igor-karpukhin: I wouldn't say it is a fix, more like an optimisation. There is no need to call "GetServerlessDeployments" if we're running on cloud-gov.

pkg/controller/connectionsecret/connectionsecrets.go
igor-karpukhin: Sure. This is just to see if we getting in there in the CI/CD

PR #1718, opened 2024-07-23
CLOUDP-261435: Fixed atlas gov

igor-karpukhin:
### All Submissions:

### DIFF
--- a/internal/translation/deployment/deployment.go
+++ b/internal/translation/deployment/deployment.go
@@ -24,18 +24,19 @@ type AtlasDeploymentsService interface {
 type ProductionAtlasDeployments struct {
 	clustersAPI   admin.ClustersApi
 	serverlessAPI admin.ServerlessInstancesApi
+	isGov         bool
 }
 
-func NewAtlasDeploymentsService(ctx context.Context, provider atlas.Provider, secretRef *types.NamespacedName, log *zap.SugaredLogger) (*ProductionAtlasDeployments, error) {
+func NewAtlasDeploymentsService(ctx context.Context, provider atlas.Provider, secretRef *types.NamespacedName, log *zap.SugaredLogger, isGov bool) (*ProductionAtlasDeployments, error) {
 	client, err := translation.NewVersionedClient(ctx, provider, secretRef, log)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create versioned client: %w", err)
 	}
-	return NewProductionAtlasDeployments(client.ClustersApi, client.ServerlessInstancesApi), nil
+	return NewProductionAtlasDeployments(client.ClustersApi, client.ServerlessInstancesApi, isGov), nil
 }
 
-func NewProductionAtlasDeployments(clusterService admin.ClustersApi, serverlessAPI admin.ServerlessInstancesApi) *ProductionAtlasDeployments {
-	return &ProductionAtlasDeployments{clustersAPI: clusterService, serverlessAPI: serverlessAPI}
+func NewProductionAtlasDeployments(clusterService admin.ClustersApi, serverlessAPI admin.ServerlessInstancesApi, isGov bool) *ProductionAtlasDeployments {
+	return &ProductionAtlasDeployments{clustersAPI: clusterService, serverlessAPI: serverlessAPI, isGov: isGov}
 }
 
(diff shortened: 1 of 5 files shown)


internal/translation/deployment/deployment.go
```diff
@@ -16,7 +16,7 @@ import (
 
 type AtlasDeploymentsService interface {
 	ListClusterNames(ctx context.Context, projectID string) ([]string, error)
-	ListDeploymentConnections(ctx context.Context, projectID string) ([]Connection, error)
+	ListDeploymentConnections(ctx context.Context, projectID string, isGov bool) ([]Connection, error)
```
josvazg, line 19: I do not think this change is needed here. The implementation can remember on its own if it is gov or not. It is a object field, rather than a parameter.

internal/translation/deployment/deployment.go
```diff
@@ -57,15 +57,19 @@ func (ds *ProductionAtlasDeployments) ListClusterNames(ctx context.Context, proj
...
+func (ds *ProductionAtlasDeployments) ListDeploymentConnections(ctx context.Context, projectID string, isGov bool) ([]Connection, error) {
 	clusters, _, err := ds.clustersAPI.ListClusters(ctx, projectID).Execute()
 	if err != nil {
 		return nil, fmt.Errorf("failed to list clusters for project %s: %w", projectID, err)
 	}
 	clusterConns := clustersToConnections(clusters.GetResults())
 
-	serverless, _, err := ds.serverlessAPI.ListServerlessInstances(ctx, projectID).Execute()
-	if err != nil {
+	if isGov {
```
josvazg, line 67: A field here would be better:
```suggestion
	if ds.isGov {
```

internal/translation/deployment/deployment.go
igor-karpukhin: Yes, good idea. Let me quickly refactor this

josvazg (approved):

pkg/controller/atlasdatabaseuser/databaseuser_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/mocks/atlas"
```
s-urbaniak, line 9: package ordering

internal/translation/deployment/deployment.go
```diff
@@ -64,8 +65,12 @@ func (ds *ProductionAtlasDeployments) ListDeploymentConnections(ctx context.Cont
 	}
 	clusterConns := clustersToConnections(clusters.GetResults())
 
-	serverless, _, err := ds.serverlessAPI.ListServerlessInstances(ctx, projectID).Execute()
-	if err != nil {
+	if ds.isGov {
+		return clusterConns, nil
+	}
```
s-urbaniak, line 70: is this case governed by unit tests?

igor-karpukhin: I don't think we have unit test for the ProductionAtlasDeployments

pkg/controller/atlasdatabaseuser/databaseuser_test.go
igor-karpukhin: What's wrong with it?

s-urbaniak: it should be in line 24

internal/translation/deployment/deployment.go
s-urbaniak: in that case, please add a unit test here to have coverage.

igor-karpukhin: Unit-test added

helderjs (approved):
LGTM. I'm okay with the same client handling both Clusters and Serverless but I prefer the translation layer to not treat them as they are the same. However, I think this should be tackled in a separate PR.

s-urbaniak (approved):

igor-karpukhin:
> LGTM. I'm okay with the same client handling both Clusters and Serverless but I prefer the translation layer to not treat them as they are the same. However, I think this should be tackled in a separate PR.

Yep, we can refactor this in the separate PR if needed

PR #1737, opened 2024-07-31
Tweak linting to be faster

josvazg:
Turns out golingci is made slow by `goimports` alone. Most other linters, including `gofmt` take below 50 seconds to run, typically even less than 45 seconds. We can also run `goimports` separately anyway.

### All Submissions:


josvazg:
While it is noticeable in local dev, it does not help in the CI

PR #1758, opened 2024-08-08
CLOUDP-263321: Unified development environment (Devbox)

filipcirtog:
### All Submissions:


docs/dev/start-guide.md
```diff
@@ -1,37 +1,39 @@
...
+For VSCode
+1. Install the extension [[Nix Environment Selector](https://marketplace.visualstudio.com/items?itemName=arrterian.nix-env-selector)] 
+2. Restart the IDE and press (Control + Shift + P) and search for Nix-Env: Select Environment command.
+3. Choose the Nix environment from the repo (shell.nix) and wait for it to build.
+4. Start using VSCode with everything installed.
 # How-To
 ## Run integration tests
+
+**IMPORTANT: Please ensure you are in a Nix environment when running any make targets**
+
```
igor-karpukhin, line 32: I assume this also needs to be updated to use Devbox?

filipcirtog: yes. thanks

.github/actions/build-push-image/action.yaml
```diff
@@ -69,28 +74,53 @@ runs:
...
       with:
         platforms: ${{ inputs.platforms }}
+        
+    # Devbox setup steps
+    - name: Install devbox
+      uses: jetify-com/devbox-install-action@v0.11.0
+
+    - name: Download Go build dependencies (Devbox)
+      run: devbox run -- go mod download
+      if: ${{ inputs.devbox-build == 'true' }}
```
josvazg, line 84: Nit pick, as this IF controls whether the rest runs or not I would place it higher up. I would go for `name` then `if` always:

```yaml
 - name: Some name ...
    if: ${{ inputs.devbox-build == 'true' }}
    ...
```

.github/workflows/openshift-upgrade-test.yaml
```diff
@@ -48,40 +48,24 @@ jobs:
...
-
-          wget https://github.com/mikefarah/yq/releases/download/v4.27.2/yq_linux_amd64 -O yq -q
-          chmod +x yq && sudo mv yq /usr/local/bin/yq
-
-          opm version
-          oc version
-          operator-sdk version
       - name: Install Go dependencies
-        run: go install golang.org/x/tools/cmd/goimports@latest
+        run: devbox run -- 'go install golang.org/x/tools/cmd/goimports@latest'
```
josvazg, line 64: Why can't this be delegated to **devbox** to install `goimports` rather than install explicit here?

.github/workflows/rebuild-released-images.yaml
```diff
@@ -78,40 +77,54 @@ jobs:
...
+          fi
+      - name: Check signing supported
+        id: check-signing-support
+        run: |
+          if test -f "./scripts/sign-multiarch.sh"; then
+            echo "sign=true" >> $GITHUB_OUTPUT
+          else
+            echo "sign=false" >> $GITHUB_OUTPUT
+          fi
+      - name: Check for devbox
```
josvazg, line 96: Should we make a note of the first version using Devbox?

In fact we should probably mark all features so we can remove support as the old versions stop being supported and rebuilt.

It does not have to be in this PR though, we could make a note to tackle this separately. It is a small maintenance annotation.

.github/workflows/openshift-upgrade-test.yaml
filipcirtog: I must have missed it! Will update that! Thanks!

.github/workflows/rebuild-released-images.yaml
josvazg: I wonder if it would be easier to just refactor this by build method:
- Version up to X supports build method 1 that means old **Dockerfile**.
- Version X-Y means build method 2 that uses **fast.Dockerfile without devbox**.
- Version Y and onward means **fast.Dockerfile with devbox**.

Still this is just food for thought and I think it is best done in a separate PR and probably separate Jira.

.github/workflows/release-post-merge.yml
```diff
@@ -195,15 +191,15 @@ jobs:
...
+          devbox run -- 'make verify IMG="${{ env.IMAGE_REPOSITORY }}:${{ steps.tag.outputs.version }}" SIGNATURE_REPO=${{ env.IMAGE_REPOSITORY }}'
+          devbox run -- 'make verify IMG="quay.io/${{ env.IMAGE_REPOSITORY }}:${{ steps.tag.outputs.version }}" SIGNATURE_REPO=${{ env.IMAGE_REPOSITORY }}'
+          devbox run -- 'make verify IMG="quay.io/${{ env.IMAGE_REPOSITORY }}:${{ steps.tag.outputs.version }}-certified" SIGNATURE_REPO=${{ env.IMAGE_REPOSITORY }}'
+          devbox run -- 'make verify IMG="${{ env.IMAGE_REPOSITORY }}:${{ steps.tag.outputs.version }}" SIGNATURE_REPO=mongodb/signatures'
+          devbox run -- 'make verify IMG="quay.io/${{ env.IMAGE_REPOSITORY }}:${{ steps.tag.outputs.version }}-certified" SIGNATURE_REPO=mongodb/signatures'
       - name: Create configuration package
         run: |
-          set -x
-          tar czvf atlas-operator-all-in-one-${{ steps.tag.outputs.version }}.tar.gz -C deploy all-in-one.yaml
+          devbox run -- 'set -x'
```
josvazg, line 201: Why do we need this `set -x` here?

If we are unsure I prefer we leave it as is for now.

josvazg (approved):
LGTM with a few small questions

.github/actions/build-push-image/action.yaml
filipcirtog: Updated. Thanks!

.github/workflows/rebuild-released-images.yaml
filipcirtog: We're in a rather complex situation with so many 'if' conditions in this workflow. It might be worth exploring whether those  can be delegated to their respective workflows, allowing for more automation. However, things should become simpler once we deprecate the old versions still relying on the previous Dockerfile but I do see a benefit in versioning those and being more explicing why we do that.

josvazg: yeah, but still I think that should be a different effort Jira and PR wise

.github/workflows/devbox-update.yml
```diff
@@ -0,0 +1,48 @@
+name: Weekly Devbox Update
```
filipcirtog, line 1: @josvazg would you please take a look here?

josvazg: LGTM, were you able to test it?

filipcirtog: yes. https://github.com/filipcirtog/mongodb-atlas-kubernetes/pull/51

flakes/k8s-controller-tools/flake.nix
```diff
@@ -0,0 +1,56 @@
+{
```
josvazg, line 1: We can also contribute upstream like this:
https://github.com/NixOS/nixpkgs/pull/288288

But this already unblocks us regardless.

josvazg (approved):

flakes/k8s-controller-tools/flake.nix
s-urbaniak: I strongly advise to contribute upstream, yes.

flakes/k8s-controller-tools/version.patch
```diff
@@ -0,0 +1,24 @@
...
++var version string
++
+ // Version returns the version of the main module
+ func Version() string {
+-	info, ok := debug.ReadBuildInfo()
+-	if !ok || info == nil || info.Main.Version == "" {
+-		// binary has not been built with module support or doesn't contain a version.
+-		return "(unknown)"
+-	}
+-	return info.Main.Version
```
s-urbaniak, line 18: did you include this patch manually? What was the process?

flakes/k8s-controller-tools/flake.nix
josvazg: I will give that a try next. The change itself is small, but the barriers to first contribution might not be so trivial. Let's see.

flakes/k8s-controller-tools/version.patch
josvazg: This is the same approach as the original nixos package:
https://github.com/NixOS/nixpkgs/tree/nixos-unstable/pkgs/development/tools/kubernetes-controller-tools
You have the `default.nix` file that we converted here to a flake, and then you reference from the flake code the patch like this:

```nix
  patches = [ ./version.patch ];
 ```
 
 Being in the git repo, the nix tool has this file available locally.
 
 BTW I lost quite some time with this. I wrongly assumed this patch was only needed for the older 0.14 version and did not look what it was actually doing.
 
 Then I realised that nix was producing a binary with version `(devel)` which can be explained looking here:
 
 https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/version/version.go#L37
 
 Took me quite a long time to realise the following:
 - The `controller-tools` use Go's module info to infer the binary version.
 - Go modules have a version as modules, but the main module is always `devel`, it does not have a version.
 - This means that when you install with `go get ...@X.Y.Z` the module will be labeled as version `X.Y.Z` but if you compile it, it will be always `devel`.
 - I suppose this is to protect the module from appearing somewhere with the same version but different code.
 - Both the nix package and our flake now patch the version code to be able to set the version in the usual `ldflags -X` way, but overriding a variable within the binary at compile time.

File: flakes/k8s-controller-tools/flake.nix [hunk 9]
```diff
@@ -0,0 +1,56 @@
...
+                version = "0.16.1";
+
+                src = pkgs.fetchFromGitHub {
+                    owner = "kubernetes-sigs";
+                    repo = "controller-tools";
+                    rev = "v${version}";
+                    sha256 = "0czfaaaxs7v7a7hy3cy83ph10r2lk4hlfn75whzp2njms5krvxh4";
+                };
+
+                k8s-controller-tools = pkgs.buildGoModule rec {
```
josvazg, line 22: `buildGoModule` ifs the nix function that does most of the magic of building the actual go program:
https://nixos.org/manual/nixpkgs/stable/#ssec-language-go

About patches:
https://nixos.org/manual/nixpkgs/stable/#var-stdenv-patches

flakes/k8s-controller-tools/flake.nix
josvazg: Upstream contribution in NixOS packages repo:
[PR #338425 kubernetes-controller-tools: 0.14.0 -> 0.16.1](https://github.com/NixOS/nixpkgs/pull/338425)

s-urbaniak:
This is looking green enough 💚 I think we're good for final approval to merge this! Let's discuss in our tomorrow's standup.

.github/actions/build-push-image/action.yaml
```diff
@@ -49,18 +49,23 @@ inputs:
   tags:
     description: A list of tags applied to the image
     required: true
+  devbox-build:
```
helderjs, line 52: Do we still need this parameter? I understand it was good for testing and evaluating the tool but IMO, now it only adds complexity to the action. I'd say we can remove it and stick to devbox only

devbox.json
```diff
@@ -0,0 +1,26 @@
...
+    "git@latest",
+    "gettext@latest",
+    "wget@latest",
+    "cosign@latest",
+    "kubernetes-helm@latest",
+    "govulncheck@latest",
+    "gotools@latest",
+    "go-licenses@latest",
+    "ginkgo@latest",
+    "operator-sdk@latest",
```
helderjs, line 21: I think we should pin the version of operator sdk

devbox.json
```diff
@@ -0,0 +1,26 @@
+{
+  "packages": [
+    "yq-go@latest",
+    "kubebuilder@latest",
```
helderjs, line 4: I think we should pin the version of kubebuilder

helderjs (approved):
LGTM, just left few suggestions

.github/actions/build-push-image/action.yaml
josvazg: I agree, we can remove. I can handle it.

devbox.json
josvazg: Agreed, seems lock is at `4.1.1`

on the other hand, anything we pin we need to revisit every now and then. It would be good to discuss how we make sure we do not forget to maintain the pinned versions on the longer run.

devbox.json
josvazg: Ok, `1.36.1` seems to be the one.

Same caveat, we need a process for revisiting pinned version, they cannot stay pinned for ever.

josvazg:
Saw some weird things on tests. These tests were skipped:
- `helm-update`
- `long-run`
- `deletion-protection`

.github/workflows/devbox-update.yml
```diff
@@ -0,0 +1,63 @@
...
+        git commit -m "Weekly devbox dependencies update"
+        git push origin ${{ steps.generate_branch.outputs.BRANCH_NAME }}  # Push branch
+
+    - name: Create Pull Request
+      if: steps.check_changes.outputs.CHANGES == 'true'
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        PR_URL=$(curl -X POST -H "Authorization: token $GITHUB_TOKEN" \
+        -d '{"title":"Weekly Devbox Update","body":"This PR contains the weekly dependencies update for Devbox.","head":"${{ steps.generate_branch.outputs.BRANCH_NAME }}","base":"origin/CLOUDP-263321/unified-development-environment-devbox"}' \
```
filipcirtog, line 60: Hi there! Just a quick note: the base should be changed to `main`instead of `origin/CLOUDP-263321/unified-development-environment-devbox`. I initially thought this workflow would trigger a weekly PR before being merged, but I was mistaken. It turns out the weekly workflows are only triggered by the main branch.

docs/dev/start-guide.md
```diff
@@ -1,37 +1,60 @@
...
 4. Clone the project to your workspace (note, that this doesn't need to be `GOPATH` as the project uses Go Modules)
 5. Copy the default Github Actions settings for local run: `cp .actrc.local.sample .actrc`
 6. Copy the default Github Actions environment for local run: `cp dotenv.sample .env`
 7. Update the .actrc - specify your Atlas connectivity data (orgId, keys)
 8. Build and deploy the Operator into the K8s cluster: `make deploy`
-9. Create an AtlasProject: `kubectl apply -f config/samples/atlas_v1_atlasproject.yaml` (note, that Atlas connection secrets are
- created during running `make deploy`)
+9. Create an AtlasProject: `kubectl apply -f config/samples/atlas_v1_atlasproject.yaml` (note, that Atlas connection secrets are created during running `make deploy`)
 10. Create an AtlasDeployment: `kubectl apply -f config/samples/atlas_v1_atlasdeployment.yaml`
 11. Create an AtlasDatabaseUser: `kubectl apply -f config/samples/atlas_v1_atlasdatabaseuser.yaml`
```
cveticm, line 19: Minor: There are two item 4s, readjust numbering here

flakes/k8s-controller-tools/version.patch
```diff
@@ -0,0 +1,24 @@
...
++	_ = debug.ReadBuildInfo
++	return version
+ }
+ 
+ // Print prints the main module version on stdout.
```
cveticm, line 23: This comment seems to be from a missing or removed func. Does this function need to be added or can this comment be removed?

josvazg:
> Saw some weird things on tests. These tests were skipped:
> 
> * `helm-update`
> * `long-run`
> * `deletion-protection`
> 
> Local test on the same branch does trigger for `long-run`, not of the other 2 tests.

#1809

flakes/k8s-controller-tools/version.patch
josvazg: Not really, this is a patch that I copied from here:
https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/tools/kubernetes-controller-tools/version.patch

This is patching the source code from here:
https://github.com/kubernetes-sigs/controller-tools/blob/v0.16.1/pkg/version/version.go#L25
Lines 25-35 are from the patch. We should not touch those
Note the changes are preceded by `+` and `-` lines and the rest are reference context code around those that **must** match for the patch to apply.

josvazg:
Failures in #1809 confirm the test issues predate this change. So will separate both fixes completely

josvazg:
The `long-run` test not running issue was introduced here, but there are several other issues to fix before right now on separate PRs.

cveticm (approved):
LGTM 👍

docs/dev/start-guide.md
```diff
@@ -1,37 +1,60 @@
...
-6. Copy the default Github Actions environment for local run: `cp dotenv.sample .env`
-7. Update the .actrc - specify your Atlas connectivity data (orgId, keys)
-8. Build and deploy the Operator into the K8s cluster: `make deploy`
-9. Create an AtlasProject: `kubectl apply -f config/samples/atlas_v1_atlasproject.yaml` (note, that Atlas connection secrets are
- created during running `make deploy`)
-10. Create an AtlasDeployment: `kubectl apply -f config/samples/atlas_v1_atlasdeployment.yaml`
-11. Create an AtlasDatabaseUser: `kubectl apply -f config/samples/atlas_v1_atlasdatabaseuser.yaml`
+1. Install devbox by following the instructions provided on the official [Jetify website](https://www.jetify.com/devbox).
+2. Run the Devbox shell (by entering ```devbox shell``` in the bash to initialize an environment with pre-installed dependencies). For more details, refer to the devbox.json file.
+3. Install optional dependencies
```
roothorp, line 4: Can many of these steps not be automated via devbox? Having this as close to a one-click 'install' would be awesome.

roothorp (approved):

docs/dev/start-guide.md
josvazg: We could, but I will probably do it as a follow up.

Probably kind can be added to devbox, so it does not need separate install. And we could run some scripts at entering devbox shell.

PR #1769, opened 2024-08-13
CLOUDP-266544: DatabaseUser independent CRD

helderjs:
### All Submissions:


config/crd/bases/atlas.mongodb.com_atlasdatabaseusers.yaml
```diff
@@ -54,6 +54,20 @@ spec:
             description: AtlasDatabaseUserSpec defines the desired state of Database
               User in Atlas
             properties:
+              atlasRef:
+                description: AtlasRef holds the Atlas project ID the user belongs
+                  to
+                properties:
+                  credentials:
+                    description: Credentials is the name of the secret that holds
```
igor-karpukhin, line 62: Are we assume credentials will always be on the same namespace as the user? I think it should be a `ResourceRefNamespaced` (just like connection secret) type according to the SPIKE

igor-karpukhin (changes requested):
Looks good in general. But I thought we wanted to make resources to reference Atlas API credentials, right? In that case, the resources that is going to be handled separately needs the API keys that are stored in a K8S secret that may be located in a different namespace.

config/crd/bases/atlas.mongodb.com_atlasdatabaseusers.yaml
```diff
@@ -200,10 +214,13 @@ spec:
...
                 type: string
             required:
-            - projectRef
             - roles
             - username
             type: object
+            x-kubernetes-validations:
+            - message: cannot define both atlasRef and projectRef at same time
+              rule: 'has(self.atlasRef) && !has(self.projectRef) ? true : !has(self.atlasRef)
+                && has(self.projectRef)'
```
s-urbaniak, line 223: I am missing a unit test for this CEL.

As mentioned to @igor-karpukhin please be inspired by https://github.com/s-urbaniak/mongodb-atlas-kubernetes/commit/8b0af16a92c2809242f4c790191861111c7be79b#diff-86f3f6d950e0b76bc3ef5858a2280676cb73ead100b087d78270c714747a5249R134 to add one here.

pkg/api/v1/externalreference.go
```diff
@@ -0,0 +1,8 @@
+package v1
+
+type ExternalProjectReference struct {
+	// ID is the Atlas project ID
+	ID string `json:"id"`
```
s-urbaniak, line 5: is this required? if so, please add a kubebuilder annotation.

pkg/indexer/atlasdatabaseuserprojects.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil
+	}
+
+	if user.Spec.Project != nil && user.Spec.Project.Name != "" {
+		return []string{user.Spec.Project.GetObject(user.Namespace).String()}
+	}
+
+	if user.Spec.AtlasRef != nil && user.Spec.AtlasRef.ID != "" {
+		return []string{user.Spec.AtlasRef.ID}
+	}
```
s-urbaniak, line 45: This is mixing two independent indexes with different keys: 1. using AKO project names as keys pointing to database users and 2. one using Atlas Project IDs pointing to database users. For the latter the index name `atlasdatabaseuser.spec.project` is not correct.

Please create a dedicated index with a correct name for atlas IDs.

If we want to replace boilerplate we could:
a) rename this index as `atlasdatabaseuser.spec.project` is not correct
b) remove the precedence of project name keys over atlas project IDs pointing to database users. That precedence is not obvious. If it is necessary it should be documented and tested.
c) Prefix each separate key, i.e. using `spec.project/fooNamespace/fooProject` and `spec.atlasRef/123456..`

Having said that I still very much prefer a separate index as it is one.

config/crd/bases/atlas.mongodb.com_atlasdatabaseusers.yaml
josvazg: I disagree, it does not seem sensible security wise to be grabbing credentials from some other namespace. Specially in this case that credentials are per CRD they can be locally provisioned.

josvazg: Actually, I have reconsidered this while talking to @helderjs.

Seems the general recommendation is that all references should be without the namespace, with a single exception being the secrets themselves.

This is because, normal resources should not depend or reach out to other namespaces, as that would have security concerns and undermines the actual reason for having namespaces in the first place. Still, in the case of secrets, it makes sense to actually allow users of a namespace to consume secrets even when they are not allowed to create or update them themselves.

So I guess yes, we should also allow namespaces for these credentials here.

pkg/api/v1/externalreference.go
josvazg: I will take this off from @helderjs and will add the kubebuilder annotations here

config/crd/bases/atlas.mongodb.com_atlasdatabaseusers.yaml
s-urbaniak: > it does not seem sensible security wise to be grabbing credentials from some other namespace.

agreed

josvazg: Reconsidering this:
> Still, in the case of secrets, it makes sense to actually allow users of a namespace to consume secrets even when they are not allowed to create or update them themselves.

[Implementing Anti-patterns: Kubernetes Cross-namespace Resource Ownership - Tom Coufal, Red Hat](https://www.youtube.com/watch?v=iWz5AAbbT-c)

We still can be Kubernetes best practices compliant and allow for managers of secrets and consumers to be different:
- Just use a namespaced used for admins with an RBAC to create or update secrets in the namespace.
- Use a different namespaced user withan RBAC for regular devs to just consume the secrets in the namespace.
- Should a secret be needed in several namespaces? just provision it on all of them. Still, if it needs to be in different namespaces maybe it should NOT be the same exact credential, it might help security to use different credentials for different dev teams or end users.

igor-karpukhin: Resolved as won't do

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -51,9 +51,12 @@ const (
 )
 
 // AtlasDatabaseUserSpec defines the desired state of Database User in Atlas
+// +kubebuilder:validation:XValidation:rule="has(self.atlasRef) && !has(self.projectRef) ? true : !has(self.atlasRef) && has(self.projectRef)",message="cannot define both atlasRef and projectRef at same time"
```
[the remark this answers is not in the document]
igor-karpukhin: nit: it can be shortened a bit: 
`// +kubebuilder:validation:XValidation:rule="(!has(self.atlasRef) && has(self.projectRef)) || (has(self.atlasRef) && !has(self.projectRef))`

pkg/api/v1/externalreference.go
josvazg: This is back now on @helderjs hands

josvazg (approved):
LGTM

pkg/api/v1/atlasdatabaseuser_types_test.go
```diff
@@ -0,0 +1,157 @@
...
+		"no project reference is set": {
+			dbUser: &AtlasDatabaseUser{
+				Spec: AtlasDatabaseUserSpec{},
+			},
+			expectedErrors: []string{"spec: Invalid value: \"object\": must define only one project reference through externalProjectRef or projectRef"},
+		},
+		"both project references are set": {
+			dbUser: &AtlasDatabaseUser{
+				Spec: AtlasDatabaseUserSpec{
+					Project: &common.ResourceRefNamespaced{
```
s-urbaniak, line 32: Can we add a test case where have a valid project reference only?

pkg/api/v1/atlasdatabaseuser_types_test.go
```diff
@@ -0,0 +1,157 @@
...
+			},
+			expectedErrors: []string{
+				"spec: Invalid value: \"object\": must define only one project reference through externalProjectRef or projectRef",
+				"spec: Invalid value: \"object\": must define a local connection secret when referencing an external project",
+			},
+		},
+		"external project references is set": {
+			dbUser: &AtlasDatabaseUser{
+				Spec: AtlasDatabaseUserSpec{
+					ExternalProjectRef: &ExternalProjectReference{
```
s-urbaniak, line 48: let's add a test case where we have an external project ref and a local connection secret.

s-urbaniak (approved):
looks great 🎉

igor-karpukhin (approved):
Great job! 👍

PR #1773, opened 2024-08-16
CLOUDP-267957: Deployment independent resource

helderjs:
### All Submissions:


config/crd/bases/atlas.mongodb.com_atlasdeployments.yaml
```diff
@@ -55,6 +55,18 @@ spec:
               AtlasDeploymentSpec defines the desired state of AtlasDeployment
               Only one of DeploymentSpec, AdvancedDeploymentSpec and ServerlessSpec should be defined
             properties:
+              atlasRef:
+                properties:
+                  credentials:
+                    description: Credentials is the name of the secret that holds
+                      the Atlas credentials
+                    type: string
```
igor-karpukhin, line 63: Same thing as with DB User: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1769#pullrequestreview-2242716924 it should be a reference to the secret including a namespace

igor-karpukhin (changes requested):

config/crd/bases/atlas.mongodb.com_atlasdeployments.yaml
```diff
@@ -816,9 +828,11 @@ spec:
                 - name
                 - providerSettings
                 type: object
-            required:
-            - projectRef
             type: object
+            x-kubernetes-validations:
+            - message: cannot define both atlasRef and projectRef at same time
```
s-urbaniak, line 833: see https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1769/files#r1721226434

config/crd/bases/atlas.mongodb.com_atlasdeployments.yaml
igor-karpukhin: Resolved as won't do

josvazg:
Converted to draft, as there are changes going before this one that will require heavy rebasing.

pkg/api/v1/atlasdatabaseuser_types_test.go
```diff
@@ -1,33 +1,22 @@
-package v1
+package v1 // nolint: dupl
```
josvazg, line 1: We did not need this before, now we do. Maybe the linter has a point and we should revisit our code?

helderjs: It complains about test cases, which are only data and don't make sense to be generalized. The generic part which actually tests and asserts data is already decoupled.

pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -172,26 +160,88 @@ func (r *AtlasDeploymentReconciler) Reconcile(context context.Context, req ctrl.
...
 			clusterDeployment = deploymentInAtlas.(*deployment.Cluster)
 		}
-		return r.handleAdvancedDeployment(ctx, deploymentInAKO.(*deployment.Cluster), clusterDeployment)
+		return r.handleAdvancedDeployment(workflowCtx, deploymentInAKO.(*deployment.Cluster), clusterDeployment)
 	}
 
 	return workflow.OK().ReconcileResult(), nil
 }
 
+func (r *AtlasDeploymentReconciler) getProjectFromAtlas(ctx *workflow.Context, atlasDeployment *akov2.AtlasDeployment) (*project.Project, error) {
```
s-urbaniak, line 183: is this the very same method as from the database users PR? If so, does it make sense to somehow centralize this method?

helderjs: The internal services built and set are distinct from controller to controller. Also, deployment specifically still needs the old Atlas client which needs to be initialized. I think duplication here is better than the complex function that will try to cover every case on every controller.

s-urbaniak:
@helderjs isn't this missing watching local secrets similar to what the database user reconciler does? https://github.com/mongodb/mongodb-atlas-kubernetes/blob/c56374f3570e44818d6e7393dfd500e7808bf0f0/pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go#L217-L227

Otherwise I don't see how the controller would react to local secret changes.

pkg/api/v1/project_reference_cel_test.go
```diff
@@ -0,0 +1,60 @@
...
+			require.Equal(t, len(tt.expectedErrors), len(errs))
+
+			for i, err := range errs {
+				assert.Equal(t, tt.expectedErrors[i], err.Error())
+			}
+		})
+	}
+}
+
+func assertExternalProjectReferenceConnectionSecret(t *testing.T, crdPath string, tests projectReferenceTestCase) {
```
josvazg, line 40: Am I missing something or is this the same code as the other function with a different name?

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -1090,3 +1093,334 @@ func TestFindDeploymentsForBackupSchedule(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			testScheme := runtime.NewScheme()
+			assert.NoError(t, akov2.AddToScheme(testScheme))
+			assert.NoError(t, corev1.AddToScheme(testScheme))
```
josvazg, line 1214: These could happen once and outside of the loop and sub test right?

josvazg (commented):
Minor comments on my end, waiting on replies to @s-urbaniak

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
helderjs: Yes, but it would be a minor optimization. We follow this pattern in every test we have.

pkg/controller/atlasdeployment/atlasdeployment_controller.go
josvazg: Wait even better, setup the client(s) and services first of all in a method:
```go
func (r *AtlasDeploymentReconciler) setupDeploymentClientServices(ctx *workflow.Context, credentialsName, namespace string) error
```

After this the reconciler only needs to call `r.projectService.GetProject`

On the other codepath you cal the `setupDeploymentClientServices` first jus the same, then call a functions to resolve the project from Kubernetes. A function that can be shared across controllers.

The only thing that needs to be per reconciler is the clients and services setup from credentials.

pkg/indexer/indexer.go
```diff
@@ -33,8 +33,7 @@ func RegisterAll(ctx context.Context, mgr manager.Manager, logger *zap.Logger) e
 		NewAtlasFederatedAuthBySecretsIndexer(logger),
 		NewAtlasDatabaseUserBySecretsIndexer(logger),
 		NewAtlasDatabaseUserByCredentialIndexer(logger),
-		NewAtlasDatabaseUserByProjectsRefIndexer(logger),
-		NewAtlasDatabaseUserByExternalProjectsRefIndexer(logger),
```
josvazg, line 37: I do not understand why we remove this.

The fact that the deployment can be managed independently does not change the fact that db users can also be managed independently and will need their indexes.

josvazg (changes requested):
I do not understand why we are removing indexes for another independently managed resource. Please clarify if that is not a mistake.

AFAIU this will break the case of using the db users alone, they lost their indexers.

pkg/indexer/indexer.go
josvazg: Wait I was confused, these are not the ones affecting the Database User reconciler. Those are on lines 34 and 35, so it is all fine.

Sorry for the noise.

josvazg (approved):
LGTM

I still think we could improve the project setting bit with smaller and more reusable functions, but that is a nitpick. It could also be done as a follow up and fixing both deployments and users in the same was and in a single PR.

pkg/indexer/indexer.go
josvazg: those do not affect the db users, I see that now

s-urbaniak (approved):

pkg/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -226,3 +204,33 @@ func (r *AtlasDeploymentReconciler) ensureAdvancedOptions(ctx *workflow.Context,
 
 	return nil
 }
+
+func (r *AtlasDeploymentReconciler) getProjectDatabaseUsers(ctx context.Context, projectID string) ([]*akov2.AtlasDatabaseUser, error) {
+	databaseUsers := &akov2.AtlasDatabaseUserList{}
+	err := r.Client.List(ctx, databaseUsers, &client.ListOptions{})
```
s-urbaniak, line 210: @helderjs @josvazg question: This will list all users in Kubernetes rather than caching them. What was wrong with the indexers/cache?

helderjs: Without going through all users and filtering them manually, the deployment would not create the secrets for all the users under their project. Basically, independent deployment would update independent users, linked would update only linked. In order to keep everything in sync we have to do that.

s-urbaniak: Sorry, I don't understand. The indexers are being updated every time a DB user resource is observed. What loss of information would we have if using a cache?

igor-karpukhin (approved):
LGTM 👍

PR #1782, opened 2024-08-22
CLOUDP-266544: Support local resource credentials

josvazg:
- Add common infrastructure for adding local per CRD credentials at `pkg/api/credentials.go`.
- Use the new infra from `AtlasDatabaseUser`'s CRD and Controller.
- Solve whether the credentials are taken from the current resource, the project or render nil.

Testing:
- Unit tests on new `ComputeSecret`.
- Conversion test fix: credentials must be ignored on comparisons between the Atlas and Kubernetes Spec.
- Users e2e test: Second user takes locally defined credentials. Also if local credentials do not exist, reconciliation fails as expected.

### All Submissions:


config/crd/bases/atlas.mongodb.com_atlasdatabaseusers.yaml
```diff
@@ -66,6 +66,24 @@ spec:
...
+                  name:
+                    default: ""
+                    description: |-
+                      Name of the referent.
+                      This field is effectively required, but due to backwards compatibility is
+                      allowed to be empty. Instances of this type with an empty value here are
+                      almost certainly wrong.
+                      TODO: Add other useful fields. apiVersion, kind, uid?
+                      More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                      TODO: Drop `kubebuilder:default` when controller-gen doesn't need it https://github.com/kubernetes-sigs/kubebuilder/issues/3896.
```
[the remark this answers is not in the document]
josvazg: Not mine though, I think they are auto generated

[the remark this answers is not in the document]
s-urbaniak: controller-runtime, for some reason, believes this is a top-level struct. This should be fixed.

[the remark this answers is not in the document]
s-urbaniak: lol, this comes straight from the `LocalObjectReference` annotations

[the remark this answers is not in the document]
s-urbaniak: Probably that's the reason why cert-manager went with his own type https://github.com/cert-manager/cert-manager/blob/7d797a45d7040eaad420d04dc829e2ae9457bb35/pkg/apis/meta/v1/types.go#L39-L49 🤓 

I am actually fine to do that as well here.

pkg/controller/customresource/customresource.go
```diff
@@ -139,3 +140,34 @@ func SetAnnotation(resource api.AtlasCustomResource, key, value string) {
...
+func ComputeSecretWithFallback[R api.ResourceWithCredentials](fallback bool, project *akov2.AtlasProject, resource R) (*client.ObjectKey, error) {
+	secret, err := resolveConnectionSecret(project, resource)
+	if fallback {
+		return secret, nil
+	}
+	return secret, err
+}
+
+func resolveConnectionSecret[R api.ResourceWithCredentials](project *akov2.AtlasProject, resource R) (*client.ObjectKey, error) {
+	if reflect.ValueOf(&resource).Elem().IsZero() {
```
s-urbaniak, line 153: let me think how we can simplify this. I don't see reflection is necessary here.

pkg/controller/customresource/customresource.go
```diff
@@ -139,3 +140,34 @@ func SetAnnotation(resource api.AtlasCustomResource, key, value string) {
 	annot[key] = value
 	resource.SetAnnotations(annot)
 }
+
+func ComputeSecretWithFallback[R api.ResourceWithCredentials](fallback bool, project *akov2.AtlasProject, resource R) (*client.ObjectKey, error) {
+	secret, err := resolveConnectionSecret(project, resource)
+	if fallback {
+		return secret, nil
+	}
+	return secret, err
```
s-urbaniak, line 149: Let's not return objects and errors. The idiom is do mutually exclusively return either the result or the error, not both.

pkg/controller/customresource/customresource.go
```diff
@@ -139,3 +140,34 @@ func SetAnnotation(resource api.AtlasCustomResource, key, value string) {
...
+}
+
+func resolveConnectionSecret[R api.ResourceWithCredentials](project *akov2.AtlasProject, resource R) (*client.ObjectKey, error) {
```
s-urbaniak, line 152: same as above, just use:
```
func resolveConnectionSecret(project *akov2.AtlasProject, resource api.ResourceWithCredentials) (*client.ObjectKey, error) {
	if resource == nil {
...
```

pkg/controller/customresource/customresource.go
```diff
@@ -139,3 +140,34 @@ func SetAnnotation(resource api.AtlasCustomResource, key, value string) {
 	annot[key] = value
 	resource.SetAnnotations(annot)
 }
+
+func ComputeSecretWithFallback[R api.ResourceWithCredentials](fallback bool, project *akov2.AtlasProject, resource R) (*client.ObjectKey, error) {
+	secret, err := resolveConnectionSecret(project, resource)
+	if fallback {
+		return secret, nil
```
s-urbaniak, line 147: this logic is very confusing. What are we falling back to? If fallback is `true`, then we never return an error but there might be an error and secret might be `nil`. I don't see the fallback here.

s-urbaniak:
One nit about the global secret fallback handling. The logic is quite confusing to the reader. Why not make it explicit, where:
```
type Provider interface {
...
	GlobalSecret() client.ObjectKey
}
```

and then just invoke:
```
credentialsSecret, err := customresource.ComputeSecret(r.AtlasProvider.GlobalSecret(), project, databaseUser)
...
```

where the fallback logic is explicit in `ComputeSecret`:
```
func ComputeSecret(globalSecret client.ObjectKey, project *akov2.AtlasProject, resource api.ResourceWithCredentials) (*client.ObjectKey, error) {
	if globalSecret.Name != "" {
		return &globalSecret, nil
	}
	return resolveConnectionSecret(project, resource)
}
```

There is less indirection and makes the fallback explicit.

pkg/controller/customresource/customresource_test.go
```diff
@@ -4,13 +4,18 @@ import (
 	"fmt"
 	"testing"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
s-urbaniak, line 7: please fix package ordering, it should be in line 8

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -127,7 +127,17 @@ func (r *AtlasDatabaseUserReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return result.ReconcileResult(), nil
 	}
 
-	dus, err := dbuser.NewAtlasDatabaseUsersService(ctx, r.AtlasProvider, project.ConnectionSecretObjectKey(), log)
+	hasFallback := r.AtlasProvider.HasGlobalFallbackSecret()
+	credentialsSecret, err := customresource.ComputeSecretWithFallback(hasFallback, project, databaseUser)
```
s-urbaniak, line 131: We should be making sure that we call out in the spec for independent CRs that the field for the local connection secret has ramifications on the non-independent case where it is being fallen back to if the global secret is not specified in the provider.

pkg/api/credentials.go
```diff
@@ -0,0 +1,32 @@
...
+// CredentialsProvider gives access to custom local credentials
+type CredentialsProvider interface {
+	Credentials() *corev1.LocalObjectReference
+}
+
+// +k8s:deepcopy-gen=false
+
+// ResourceWithCredentials is to be implemented by all CRDs using custom local credentials
+type ResourceWithCredentials interface {
+	CredentialsProvider
```
roothorp, line 20: Why do we need this spread across 2 interfaces, rather than just having the LocalObjectReference in this one?

config/crd/bases/atlas.mongodb.com_atlasdatabaseusers.yaml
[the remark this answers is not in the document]
josvazg: ok, I will make our reference object.

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
roothorp, line 131: Does this not change the default behaviour going forward? Are we not now attempting to fall back to the global secret at all times?

pkg/controller/customresource/customresource.go
```diff
@@ -139,3 +140,34 @@ func SetAnnotation(resource api.AtlasCustomResource, key, value string) {
 	annot[key] = value
 	resource.SetAnnotations(annot)
 }
+
+func ComputeSecretWithFallback[R api.ResourceWithCredentials](fallback bool, project *akov2.AtlasProject, resource R) (*client.ObjectKey, error) {
```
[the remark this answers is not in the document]
josvazg: The problem is that current `resolveConnectionSecret` cannot actually work just with the interface, as it needs to access the credentials but also the name and namespace of the resource.

Until I simplify that I cannot remove the generics.

I think I have an idea, though. Add all requirements to the interface.

pkg/controller/customresource/customresource.go
josvazg: yes, the fallback is the global secret. The problem is the fallback logic is elsewhere.

We could move such logic here to be more explicit:
1- Use local credential if present.
2- Otherwise use project credential if present.
3- Otherwise use global credential if passed in.
4- Otherwise fail.

pkg/controller/customresource/customresource.go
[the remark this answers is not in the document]
josvazg: Never mind the interface already has it all, will just remove the generics usage.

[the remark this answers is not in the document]
s-urbaniak: @josvazg i just removed the generic and it worked as suggested.

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: will annotate that

pkg/api/credentials.go
josvazg: Because this can be implemented directly by the simple struct object, which is applied to the Spec sub-struct.

The wider interface applies to the actual CRD as a whole, not just the extended Spec.

It is the simplest way I found to make this as reusable as possible. The consumers just need to embed the local ref struct and implement the delegation to the interface it implements from the CRD struct.

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
josvazg: This would allow not having a global, whereas now the global is always expected. But the default behaviour should remain that there is the project creds or the global ones.

pkg/controller/customresource/customresource.go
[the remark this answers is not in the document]
josvazg: Yep, I already have this fixed in the next update.

pkg/controller/customresource/customresource_test.go
josvazg: fix comming

config/crd/bases/atlas.mongodb.com_atlasdatabaseusers.yaml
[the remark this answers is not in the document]
josvazg: Done

pkg/controller/customresource/customresource.go
josvazg: Fixed

pkg/controller/customresource/customresource.go
josvazg: In the end I separated away the global fallback logic, and basically left it alone. We can go deeper on a follow up of we would like to make the global fallback optional.

So the current code only choses the local credential if present, fallbacks to the project's one, if present, or returns nil. The function can still error if the input parameters are passed in nil incorrectly.

internal/translation/dbuser/conversion_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	corev1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
josvazg: I think the issue here is `goimports` notices when you put the import in a group it does not belong to, but rather than fix it on its own, is takes the line out so that you would put it in the right place.

The right place is not as simple as "move it to line 9", the right place is move it to the right block of imports and let `goimports` sort it in there.

We seems to be using 3 blocks of imports, that `goimports` will respect:
1. Go stdlb imports
2. Our code
3. Other libraries (including other libraries by MongoDB, basically any other repo)

Unfortunately, we do not follow the ordering consistently, sometimes it is:
1. Go
2. Ours
3. Theirs

Others is Ours and Theirs are swapped 

I suggest we stick with the above sample order, 3 blocks Go/Ours/Theirs

josvazg:
> One nit about the global secret fallback handling. The logic is quite confusing to the reader. Why not make it explicit, where:
> 
> ```
> type Provider interface {
> ...
> 	GlobalSecret() client.ObjectKey
> }
> ```
> 
> and then just invoke:
> 
> ```
> credentialsSecret, err := customresource.ComputeSecret(r.AtlasProvider.GlobalSecret(), project, databaseUser)
> ...
> ```
> 
> where the fallback logic is explicit in `ComputeSecret`:
> 
> ```
> func ComputeSecret(globalSecret client.ObjectKey, project *akov2.AtlasProject, resource api.ResourceWithCredentials) (*client.ObjectKey, error) {
> 	if globalSecret.Name != "" {
> 		return &globalSecret, nil
> 	}
> 	return resolveConnectionSecret(project, resource)
> }
> ```
> 
> There is less indirection and makes the fallback explicit.

In the end I removed the global fallback from this part of the resolution. That will still work as before.

The only loss is we do not have precise errors when there is no global fallback, but that scenario was problematic to open up so soon anyway. We should decide if the global fallback will be optional separately.

s-urbaniak (approved):

.github/workflows/test-e2e.yml
```diff
@@ -276,6 +276,8 @@ jobs:
           MCLI_PRIVATE_API_KEY: ${{ secrets.ATLAS_PRIVATE_KEY }}
           MCLI_ORG_ID: ${{ secrets.ATLAS_ORG_ID}}
           MCLI_OPS_MANAGER_URL: "https://cloud-qa.mongodb.com/"
+          ATLAS_LOCAL_PUBLIC_KEY: ${{ secrets.ATLAS_LOCAL_PUBLIC_KEY }}
+          ATLAS_LOCAL_PRIVATE_KEY: ${{ secrets.ATLAS_LOCAL_PRIVATE_KEY }}
```
igor-karpukhin, line 280: @josvazg  Why do we need to duplicate API credentials?

josvazg: You are right, we do not need this credentials.

I can simplify, but then the gov test issues will get me, so I won't be able to merge today.

josvazg: removed

s-urbaniak:
rerunning e2e gov, let's evaluate if this is still a blocker.

josvazg:
Our nightly test suite saw recovery today at midnight on the gov tests:
✅ [cloud-tests / test-e2e-gov / E2E Gov tests](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/10658642231/job/29549112159#logs)

josvazg:
@s-urbaniak all green again, I closed CLOUDP-270948

s-urbaniak (approved):

helderjs (approved):

PR #1795, opened 2024-08-27
Fix labeler

josvazg:
### All Submissions:


josvazg:
Not needed.

The issue was that the manual workflow tests passed because they got the credentials from the release environment correctly. The on.push and on.pull events on the context of a `dependabot` change were not getting those because they happened in the `dependabot` context still, which is a separated secrets namespace. The fix was to add the credentials to that separated `depandabot` namespace as well, so the AKO bot fixing the dependbot code would have access.

✅ [#1794 got tested after the license fix update on its own](https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1794)

PR #1809, opened 2024-09-06
CLOUDP-272039: Fail when no tests are run

josvazg:
CLOUDP-272039

### All Submissions:


.github/workflows/test-e2e-gov.yml
```diff
@@ -52,7 +52,11 @@ jobs:
           cd test/e2e
           ginkgo labels
           echo 'Running: AKO_E2E_TEST=1 ginkgo --label-filter="atlas-gov" --timeout 120m --nodes=10  --flake-attempts=1 --randomize-all --race --cover --v --trace --show-nodes-events --output-interceptor-mode=none' && \
-          AKO_E2E_TEST=1 ginkgo --label-filter="atlas-gov" --timeout 120m --nodes=10 --flake-attempts=1 --randomize-all --race --cover --v --trace --show-node-events --output-interceptor-mode=none --coverpkg=github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/...
+          AKO_E2E_TEST=1 ginkgo --label-filter="atlas-gov" --timeout 120m --nodes=10 --flake-attempts=1 --randomize-all --race --cover --v --trace --show-node-events --output-interceptor-mode=none --coverpkg=github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/... | tee .e2e.log
+          if grep -q "0 Passed" < .e2e.log ; then
+            echo "No tests were run!"
+            exit 1
```
s-urbaniak, line 58: I am a bit concerned with the repetition of this line across many scripts below.

josvazg: The CI should be running the same code as the local dev, but it does not. That is why this fix needs so much repetition.

Fixing the duplicity is a bigger issue than the focus of this PR. It has already been flagged before but not prioritised.

helderjs: @josvazg I took a quick look on Ginkgo and we can reduce duplication by adding the following code to the Ginkgo suites we have.

```
var _ = ReportAfterSuite("Ensure test suite was not empty", func(r Report) {
	Expect(r.SuiteSucceeded && r.PreRunStats.SpecsThatWillRun > 0).To(BeTrue(), "Suite must run at least 1 test in order to be valid")
})
```

What do you think?

test/helper/e2e/actions/steps.go
```diff
@@ -130,6 +131,10 @@ func WaitTestApplication(data *model.TestDataProvider, ns, labelKey, labelValue
 	// temp
 	isAppRunning := func() func() bool {
 		return func() bool {
+			status, err := k8s.GetPodStatus(data.Context, data.K8SClient, ns)
```
helderjs, line 134: Are these debugging bits? if not, could you clarify how it supports the second status check (line 138)?

.github/workflows/test-e2e-gov.yml
josvazg: Uhm, yes I think that would be better.

There is a problem, when running locally we are calling the `int` tests and `int/clusterwide` somewhat separately, so we might run a label that only runs on one of them and the other should not complain. There is probably another solution for it, but need to figure that out.

In the meantime, I need to understand the `helm-update` test failure, which only happens in the CI for some reason.

test/helper/e2e/actions/steps.go
josvazg: yes, I am debugging, please ignore these for the time being.

I will mark as draft and iterate in the `helm-update` error first, then will try to move to check the issues within ginkgo instead of from a script, as you suggested above.

josvazg:
Debugging `helm-update` e2e test issue.

.github/workflows/test-e2e-gov.yml
josvazg: Applied this suggestion already.

test/helper/e2e/actions/steps.go
josvazg: Removed the traces, thinking about a different way to debug

josvazg:
After adding traces:
https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/10808820530/job/29983049730?pr=1809
```shell
  failed to get pods: failed to get deployment: deployments.apps "mongodb-atlas-operator" not found
  failed to get stream: container "manager" in pod "mongodb-atlas-operator-6f76cd65d9-8kbkh" is waiting to start: image can't be pulled
```

josvazg:
[98b1cab](https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1809/commits/98b1cabb20c350191f8d93e32f1a3381ab719288)
That worked:
https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/10811663472/job/29992054452?pr=1809

s-urbaniak (approved):

roothorp (approved):
Walked through in zoom together

PR #1812, opened 2024-09-09
[HOLD] CLOUDP-265544: Support local credentials for Users

josvazg:
### All Submissions:


josvazg:
CLOUDP-266544

pkg/api/v1/atlasdatabaseuser_types.go
```diff
@@ -140,6 +143,15 @@ type AtlasDatabaseUserList struct {
 	Items           []AtlasDatabaseUser `json:"items"`
 }
 
+// ReconciliableRequests implements ReconciliableList for Database Users
+func (list *AtlasDatabaseUserList) ReconciliableRequests() []reconcile.Request {
```
s-urbaniak, line 147: I suggest to leave this in the indexer function. This function is part of `pkg/api` which is official API surface. We should keep it clean, just having structs and metadata for kubebuilder.

pkg/indexer/localcredentials.go
```diff
@@ -0,0 +1,83 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api"
+)
+
+const (
+	AtlasDatabaseUserCredentialsIndex = "atlasdatabaseuser.credentials"
+)
+
+type LocalCredential struct {
```
helderjs, line 22: Given the comment from @s-urbaniak about the project ref indexer, I'm wondering a generic indexer from local secret is a good practice.

pkg/indexer/localcredentials.go
```diff
@@ -0,0 +1,83 @@
...
+	obj    client.Object
+	name   string
+	logger *zap.SugaredLogger
+}
+
+func NewLocalCredentialsIndexer(name string, obj client.Object, logger *zap.Logger) *LocalCredential {
+	return &LocalCredential{
+		obj:    obj,
+		name:   name,
+		logger: logger.Named(AtlasStreamInstanceByConnectionIndex).Sugar(),
```
helderjs, line 32: You may want to name logger according to index type

pkg/api/v1/atlasdatabaseuser_types.go
helderjs: We would have a problem with circular dependency, wouldn't we? Also, this is likely to be used by the indexes mapper function which usually lives in the controller.
Last, I'm wondering if a sugar abstraction to avoid iterating over list items is worth it.

helderjs (changes requested):

pkg/api/v1/atlasdatabaseuser_types.go
josvazg: I need to think how to solve this. I understand that it should not be here. My goal was to avoid to have to write this logic for every new CR that would use the local creds.

pkg/indexer/localcredentials.go
josvazg: I really do not want to pi¡y&paste and rename an indexer for each local cred usage.

helderjs: I understand your motivation, but most of our indexes follow a pattern to refer to a project or a secret, we could group and easily generalize the implementation for each case, however, we have an indexer per index. Why should we deviate from this here?

pkg/indexer/localcredentials.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil
+	}
+
+	if localRef := credentialProvider.Credentials(); localRef != nil && localRef.Name != "" {
+		return []string{types.NamespacedName{Namespace: object.GetNamespace(), Name: localRef.Name}.String()}
+	}
+	return []string{}
+}
+
+func CredentialsIndexMapperFunc(indexerName string, list api.ReconciliableList, kubeClient client.Client, logger *zap.SugaredLogger) handler.MapFunc {
```
s-urbaniak, line 62: is this actually used? 🤔

josvazg: Let me double check whether or not I wired in properly.

pkg/api/v1/atlasdatabaseuser_types.go
josvazg: I am using a generic function instead. It is implemented for each supporting CRD on the indexer package, so not exposed any more.

pkg/indexer/localcredentials.go
josvazg: Less code to write, to test and less potential mistakes.

The current fix covers the initial concerns and will support all other usages of the credentials by just adding a single mapping function, similar to `DatabaseUserRequests`.

pkg/indexer/localcredentials.go
josvazg: It was not, but now it is being wired as expected within the database user controller:

```go
func (r *AtlasDatabaseUserReconciler) SetupWithManager(mgr ctrl.Manager, skipNameValidation bool) error {
		Named("AtlasDatabaseUser").
		For(&akov2.AtlasDatabaseUser{}, builder.WithPredicates(r.GlobalPredicates...)).
		Watches(
			&corev1.Secret{},
			handler.EnqueueRequestsFromMapFunc(r.findAtlasDatabaseUserForSecret),
			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
		).
		Watches(
			&corev1.Secret{},
			handler.EnqueueRequestsFromMapFunc(indexer.CredentialsIndexMapperFunc(
				indexer.AtlasDatabaseUserCredentialsIndex,
				&akov2.AtlasDatabaseUserList{},
				indexer.DatabaseUserRequests,
				r.Client,
				r.Log,
			)),
			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
		).
		WithOptions(controller.TypedOptions[reconcile.Request]{SkipNameValidation: pointer.MakePtr(skipNameValidation)}).
		Complete(r)
}
```

pkg/indexer/localcredentials.go
```diff
@@ -0,0 +1,103 @@
...
+// Reconciliable is implemented by CRD objects used by indexes to trigger reconciliations
+type Reconciliable interface {
+	ReconciliableRequests() []reconcile.Request
+}
+
+// ReconciliableList is a Reconciliable that is also a CRD list
+type ReconciliableList interface {
+	client.ObjectList
+	Reconciliable
+}
```
helderjs, line 33: Nobody is implementing or relying on them, why do we need them?

pkg/indexer/localcredentials.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil
+	}
+
+	if localRef := credentialProvider.Credentials(); localRef != nil && localRef.Name != "" {
+		return []string{types.NamespacedName{Namespace: object.GetNamespace(), Name: localRef.Name}.String()}
+	}
+	return []string{}
+}
+
+type requestsFunc[L client.ObjectList] func(L) []reconcile.Request
```
helderjs, line 69: Why is it written as a generic function when it doesn't?
```suggestion
type requestsFunc func(L client.ObjectList) []reconcile.Request
```

helderjs (changes requested):
LGTM overall, just some last clarification/suggestion questions

pkg/indexer/localcredentials.go
josvazg: Checking, I might have overlook to cleanup. Thanks for the heads up!

pkg/indexer/localcredentials.go
josvazg: Let me double check and fix accordingly, thanks!

josvazg: Actually that is required. Because `ObjectList` is not enough. When you pass the list it has to be an ObjectList, but the actual list of items is not defined in the basic Object list, but on something that extends the ObjectList.

See the first implementation:
```go
func DatabaseUserRequests(list *akov2.AtlasDatabaseUserList) []reconcile.Request {
	requests := make([]reconcile.Request, 0, len(list.Items))
	for _, item := range list.Items {
		requests = append(requests, toRequest(&item))
	}
	return requests
}
```
`AtlasDatabaseUserList` is an ObjectList, but it also needs to be able to list items, actual Atlas Database Users in this case which the ObjectList alone cannot do.

helderjs (approved):

igor-karpukhin (approved):

s-urbaniak (approved):

PR #1817, opened 2024-09-13
CLOUDP-228592: Migrate Assigned Teams Api

cveticm:
[CLOUDP-228592](https://jira.mongodb.org/browse/CLOUDP-228592)
Migration of Assigned Teams has been broken into three sections; converting to new SDK, implementing translation layer, implementing state machine logic. This PR covers conversion to new SDK.

### All Submissions:


pkg/api/v1/project_teams.go
```diff
@@ -27,14 +27,15 @@ type Team struct {
 	Roles []TeamRole `json:"roles"`
 }
 
-func (in *Team) ToAtlas(teamID string) *mongodbatlas.ProjectTeam {
-	result := &mongodbatlas.ProjectTeam{
-		TeamID:    teamID,
-		RoleNames: make([]string, 0, len(in.Roles)),
+func (in *Team) ToAtlas(teamID string) admin.TeamRole {
```
s-urbaniak, line 30: It would be great to have this migrated to the translation layer.

pkg/controller/atlasproject/project_test.go
```diff
@@ -147,10 +141,16 @@ func TestHandleProject(t *testing.T) {
 				mockPeeringEndpointAPI.EXPECT().
 					ListPeeringConnectionsExecute(admin.ListPeeringConnectionsApiRequest{ApiService: mockPeeringEndpointAPI}).
 					Return(&admin.PaginatedContainerPeer{}, nil, nil)
+				mockTeamApi := mockadmin.NewTeamsApi(t)
```
josvazg, line 144: If the translation layer is implemented, these tests mocks tend to get easier as the test here only needs to cover up to the translation layer call, not up to the SDK call.

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"sync"
 
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
```
josvazg, line 10: If moving to the translation layer, the references to the SDK disappear from here. Instead the reconcile code knows only the translation layer and such translation layer is the only one concerned with the SDK calls.

pkg/controller/atlasproject/project_test.go
cveticm: Yes agreed, the translation layer will be implemented as a part of a following PR!

pkg/controller/atlasproject/team_reconciler.go
cveticm: As mentioned in my reply to your previous comment, this PR only concerns converting the API to the new SDK and translation layer will be implemented in a following PR

pkg/api/v1/project_teams.go
```diff
@@ -27,14 +27,15 @@ type Team struct {
...
+	roleNames := make([]string, 0, len(in.Roles))
+	result := admin.TeamRole{
+		TeamId:    &teamID,
+		RoleNames: &roleNames,
 	}
 
 	for _, role := range in.Roles {
-		result.RoleNames = append(result.RoleNames, string(role))
+		*result.RoleNames = append(*result.RoleNames, string(role))
```
gssbzn, line 38: use getters and setters to avoid pointer operations https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_5_best_practices.md#using-getters-instead-of-direct-field-access

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -147,7 +150,7 @@ func ensureTeamState(workflowCtx *workflow.Context, team *akov2.AtlasTeam) (stri
 			return "", workflow.Terminate(workflow.TeamNotUpdatedInAtlas, err.Error())
 		}
 
-		return atlasTeam.ID, workflow.OK()
+		return *atlasTeam.Id, workflow.OK()
```
gssbzn, line 153: https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_5_best_practices.md#using-getters-instead-of-direct-field-access

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -109,7 +112,7 @@ func (r *AtlasProjectReconciler) teamReconcile(
 				return workflow.OK().ReconcileResult(), nil
 			} else {
 				log.Infow("-> Starting AtlasTeam deletion", "spec", team.Spec)
-				_, err := teamCtx.Client.Teams.RemoveTeamFromOrganization(teamCtx.Context, orgID, team.Status.ID)
+				_, err := teamCtx.SdkClient.TeamsApi.RemoveProjectTeam(teamCtx.Context, projectID, team.Status.ID).Execute()
```
igor-karpukhin, line 115: I think this changes the logic a bit, no? Teams used to be attached to organizations, not projects. Are we sure here that a team gets deleted from the org, not just detached from a project?

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -186,20 +189,19 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
 	}
 
 	g, taskContext := errgroup.WithContext(workflowCtx.Context)
 
-	for i := range atlasUsers {
-		user := atlasUsers[i]
-		if _, ok := usernamesMap[atlasUsers[i].Username]; !ok {
+	for i := range *atlasUsers.Results {
+		user := (*atlasUsers.Results)[i]
+		if _, ok := usernamesMap[(*atlasUsers.Results)[i].Username]; !ok {
```
igor-karpukhin, line 201:
```suggestion
		if _, ok := usernamesMap[user.Username]; !ok {
```

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -186,20 +189,19 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
+	for _, atlasUser := range *atlasUsers.Results {
 		atlasUsernamesMap[atlasUser.Username] = atlasUser
 	}
 
 	g, taskContext := errgroup.WithContext(workflowCtx.Context)
 
-	for i := range atlasUsers {
-		user := atlasUsers[i]
-		if _, ok := usernamesMap[atlasUsers[i].Username]; !ok {
+	for i := range *atlasUsers.Results {
```
igor-karpukhin, line 199: I'd compare it to `nil` first, before deferencing

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -212,20 +214,20 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
-				user, _, err := workflowCtx.Client.AtlasUsers.GetByName(taskContext, string(username))
+				user, _, err := workflowCtx.SdkClient.MongoDBCloudUsersApi.GetUserByUsername(taskContext, string(username)).Execute()
 
 				if err != nil {
 					return err
 				}
 
 				lock.Lock()
-				toAdd = append(toAdd, user.ID)
+				toAdd = append(toAdd, admin.AddUserToTeam{Id: *user.Id})
```
igor-karpukhin, line 230:
```suggestion
				toAdd = append(toAdd, admin.AddUserToTeam{Id: user.GetId()})
```

pkg/controller/atlasproject/teams.go
```diff
@@ -80,54 +80,62 @@ func (r *AtlasProjectReconciler) syncAssignedTeams(ctx *workflow.Context, projec
...
+	toDelete := make([]*admin.TeamRole, 0)
 
-			continue
-		}
+	if atlasAssignedTeams != nil && atlasAssignedTeams.Results != nil {
+		for _, atlasAssignedTeam := range *(atlasAssignedTeams.Results) {
+			var desiredTeam *akov2.Team
+			var ok bool
+			if atlasAssignedTeam.TeamId != nil {
+				desiredTeam, ok = teamsToAssign[*atlasAssignedTeam.TeamId]
```
igor-karpukhin, line 90:
```suggestion
				desiredTeam, ok = teamsToAssign[*atlasAssignedTeam.GetTeamId()]
```

pkg/controller/atlasproject/teams.go
```diff
@@ -80,54 +80,62 @@ func (r *AtlasProjectReconciler) syncAssignedTeams(ctx *workflow.Context, projec
...
+				ctx.Log.Warnf("failed to remove team %s from project: %s", *atlasAssignedTeam.TeamId, err.Error())
+			}
 		}
 	}
 
 	for _, atlasAssignedTeam := range toDelete {
-		ctx.Log.Debugf("removing team %s from project", atlasAssignedTeam.TeamID)
-		_, err = ctx.Client.Teams.RemoveTeamFromProject(ctx.Context, projectID, atlasAssignedTeam.TeamID)
+		ctx.Log.Debugf("removing team %s from project", *atlasAssignedTeam.TeamId)
+		_, err = ctx.SdkClient.TeamsApi.RemoveProjectTeam(ctx.Context, projectID, *atlasAssignedTeam.TeamId).Execute()
```
igor-karpukhin, line 119:
```suggestion
		_, err = ctx.SdkClient.TeamsApi.RemoveProjectTeam(ctx.Context, projectID, *atlasAssignedTeam.GetTeamId()).Execute()
```

igor-karpukhin (commented):
Looks good in general! 👍 Please consider using getters for pointer values

pkg/controller/atlasproject/team_reconciler.go
cveticm: Yes you're right! The change in func naming convention that 'remove' no longer refers to deletion misled me. Changing it to delete from org and not just detach from project 👍

josvazg: For this PR it makes sense to keep the functionality, as this is just a refactor for the point of view of end users.

Still, as a general idea, I think we settle that the operator should not be removing Teams, as they might be used for other purposes at the Org level. Should we follow up on this topic?
cc @s-urbaniak

cveticm: @josvazg yes, that would be a good point to come to consensus on prior to the state-machine implementation since it would be a natural point to include that logic -that AKO shouldn't delete a resource on the org level such as teams.

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -186,22 +189,23 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
 		usernamesMap[string(username)] = struct{}{}
 	}
 
-	atlasUsernamesMap := map[string]mongodbatlas.AtlasUser{}
-	for _, atlasUser := range atlasUsers {
+	atlasUsernamesMap := map[string]admin.CloudAppUser{}
+	for _, atlasUser := range *atlasUsers.Results {
```
s-urbaniak, line 193:
```suggestion
	for _, atlasUser := range atlasUsers.GetResults() {
```

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -186,22 +189,23 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
-		user := atlasUsers[i]
-		if _, ok := usernamesMap[atlasUsers[i].Username]; !ok {
-			g.Go(func() error {
-				workflowCtx.Log.Debugf("removing user %s from team %s", user.ID, teamID)
-				_, err := workflowCtx.Client.Teams.RemoveUserToTeam(taskContext, workflowCtx.OrgID, teamID, user.ID)
-
-				return err
-			})
+	if atlasUsers.Results != nil {
+		for i := range *atlasUsers.Results {
```
s-urbaniak, line 200:
```suggestion
		for _, user := range atlasUsers.GetResults() {
			if _, ok := usernamesMap[user.Username]; !ok {
```

pkg/controller/atlasproject/teams.go
```diff
@@ -80,54 +80,62 @@ func (r *AtlasProjectReconciler) syncAssignedTeams(ctx *workflow.Context, projec
 
 	defer statushandler.Update(ctx, r.Client, r.EventRecorder, project)
 
-	toDelete := make([]*mongodbatlas.Result, 0, len(atlasAssignedTeams.Results))
-	for _, atlasAssignedTeam := range atlasAssignedTeams.Results {
-		desiredTeam, ok := teamsToAssign[atlasAssignedTeam.TeamID]
-		if !ok {
-			toDelete = append(toDelete, atlasAssignedTeam)
+	toDelete := make([]*admin.TeamRole, 0)
```
s-urbaniak, line 83:
```suggestion
	toDelete := make([]*admin.TeamRole, 0, len(atlasAssignedTeams.GetResults()))
```

pkg/controller/atlasproject/teams.go
```diff
@@ -80,54 +80,62 @@ func (r *AtlasProjectReconciler) syncAssignedTeams(ctx *workflow.Context, projec
...
 
-			continue
-		}
+	if atlasAssignedTeams != nil && atlasAssignedTeams.Results != nil {
+		for _, atlasAssignedTeam := range *(atlasAssignedTeams.Results) {
```
s-urbaniak, line 86:
```suggestion
		for _, atlasAssignedTeam := range atlasAssignedTeams.GetResults() {
```

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -244,59 +248,63 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
-func renameTeam(workflowCtx *workflow.Context, atlasTeam *mongodbatlas.Team, newName string) (*mongodbatlas.Team, error) {
+func renameTeam(workflowCtx *workflow.Context, atlasTeam *admin.Team, newName string) (*admin.Team, error) {
 	if atlasTeam.Name == newName {
 		return atlasTeam, nil
 	}
 
-	workflowCtx.Log.Debugf("updating name of team %s in atlas", atlasTeam.ID)
-	atlasTeam, _, err := workflowCtx.Client.Teams.Rename(workflowCtx.Context, workflowCtx.OrgID, atlasTeam.ID, newName)
+	workflowCtx.Log.Debugf("updating name of team %s in atlas", atlasTeam.Id)
+	teamUpdate := admin.TeamUpdate{Links: atlasTeam.Links, Name: newName}
```
s-urbaniak, line 301:
```suggestion
	teamUpdate := admin.TeamUpdate{Name: newName}
```

Setting the `Links` property in the `TeamUpdate` struct is not desired. As per godoc this is a read-only field:
```
type TeamUpdate struct {
	// List of one or more Uniform Resource Locators (URLs) that point to API sub-resources, related API resources, or both. RFC 5988 outlines these relationships.
	// Read only field.
	Links *[]Link `json:"links,omitempty"`
...
}
```

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -244,59 +248,63 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
+	atlasTeam, _, err := workflowCtx.SdkClient.TeamsApi.CreateTeam(workflowCtx.Context, workflowCtx.OrgID, atlasTeam).Execute()
 	if err != nil {
 		return nil, err
 	}
 
 	return atlasTeam, nil
 }
 
-func renameTeam(workflowCtx *workflow.Context, atlasTeam *mongodbatlas.Team, newName string) (*mongodbatlas.Team, error) {
+func renameTeam(workflowCtx *workflow.Context, atlasTeam *admin.Team, newName string) (*admin.Team, error) {
```
s-urbaniak, line 295:
```suggestion
func renameTeam(workflowCtx *workflow.Context, atlasTeam *admin.Team, newName string) (*admin.TeamResponse, error) {
```

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -244,59 +248,63 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
-	workflowCtx.Log.Debugf("updating name of team %s in atlas", atlasTeam.ID)
-	atlasTeam, _, err := workflowCtx.Client.Teams.Rename(workflowCtx.Context, workflowCtx.OrgID, atlasTeam.ID, newName)
+	workflowCtx.Log.Debugf("updating name of team %s in atlas", atlasTeam.Id)
+	teamUpdate := admin.TeamUpdate{Links: atlasTeam.Links, Name: newName}
+	atlasTeamResponse, _, err := workflowCtx.SdkClient.TeamsApi.RenameTeam(workflowCtx.Context, workflowCtx.OrgID, atlasTeam.GetId(), &teamUpdate).Execute()
 	if err != nil {
 		return nil, err
 	}
 
+	atlasTeam.SetName(atlasTeamResponse.GetName())
```
s-urbaniak, line 307: By returning `*admin.TeamResponse` this conversion can be omitted.

s-urbaniak, line 307:
```suggestion
	return atlasTeamResponse, nil
```

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -321,7 +329,12 @@ func teamsManagedByAtlas(workflowCtx *workflow.Context) customresource.AtlasChec
 			return false, err
 		}
 
-		if team.Spec.Name != atlasTeam.Name || len(atlasTeam.Usernames) == 0 {
+		atlasTeamUsers, _, err := workflowCtx.SdkClient.TeamsApi.ListTeamUsers(workflowCtx.Context, workflowCtx.OrgID, atlasTeam.GetName()).Execute()
+		if err != nil {
+			return false, err
+		}
+
+		if team.Spec.Name != *atlasTeam.Name || len(*atlasTeamUsers.Results) == 0 {
```
s-urbaniak, line 337:
```suggestion
		if team.Spec.Name != atlasTeam.GetName() || len(atlasTeamUsers.GetResults()) == 0 {
```

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -330,6 +343,11 @@ func teamsManagedByAtlas(workflowCtx *workflow.Context) customresource.AtlasChec
 			usernames = append(usernames, string(username))
 		}
 
-		return cmp.Diff(usernames, atlasTeam.Usernames) != "", nil
+		atlasUsernames := make([]string, 0, len(*atlasTeamUsers.Results))
+		for _, user := range *atlasTeamUsers.Results {
```
s-urbaniak, line 347:
```suggestion
		atlasUsernames := make([]string, 0, len(atlasTeamUsers.GetResults()))
```

pkg/controller/atlasproject/team_reconciler_test.go
```diff
@@ -3,24 +3,30 @@ package atlasproject
...
+	"github.com/stretchr/testify/mock"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
+
 	"github.com/stretchr/testify/assert"
 	"go.mongodb.org/atlas/mongodbatlas"
 
-	atlasmock "github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/mocks/atlas"
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+	"go.mongodb.org/atlas-sdk/v20231115008/mockadmin"
```
s-urbaniak, line 17: please fix package ordering by grouping:
1. golang SDK packages
2. external packages
3. internal packages

in separate blocks

pkg/controller/atlasproject/teams.go
```diff
@@ -80,54 +80,62 @@ func (r *AtlasProjectReconciler) syncAssignedTeams(ctx *workflow.Context, projec
...
-				ID:      atlasAssignedTeam.TeamID,
-				TeamRef: desiredTeam.TeamRef,
+					continue
+				}
 			}
-			delete(teamsToAssign, atlasAssignedTeam.TeamID)
 
-			continue
-		}
+			if !hasTeamRolesChanged(atlasAssignedTeam.GetRoleNames(), desiredTeam.Roles) {
```
s-urbaniak, line 99: if `desiredTeam` is nil which can happen if `atlasAssignedTeam` is nil, then this will panic.

pkg/controller/atlasproject/teams.go
```diff
@@ -80,54 +80,62 @@ func (r *AtlasProjectReconciler) syncAssignedTeams(ctx *workflow.Context, projec
...
+				if !ok {
+					result := atlasAssignedTeam
+					toDelete = append(toDelete, &result)
 
-		if !hasTeamRolesChanged(atlasAssignedTeam.RoleNames, desiredTeam.Roles) {
-			currentProjectsStatus[atlasAssignedTeam.TeamID] = status.ProjectTeamStatus{
-				ID:      atlasAssignedTeam.TeamID,
-				TeamRef: desiredTeam.TeamRef,
+					continue
+				}
```
s-urbaniak, line 96: remove this whole block between line 87 and 96 in favor of suggestion available at https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1817/files?diff=split&w=0#r1776459297.

pkg/controller/atlasproject/teams.go
s-urbaniak: note: The suggestion available at https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1817/files?diff=split&w=0#r1776459297 omits this edge case.

s-urbaniak (changes requested):
some nits, else this looks great! 🎉

roothorp (approved):
LGTM - just some notes going forward into translation layer/state machine controller

pkg/api/v1/project_teams.go
```diff
@@ -27,14 +27,17 @@ type Team struct {
...
+		TeamId:    &teamID,
+		RoleNames: &roleNames,
 	}
 
 	for _, role := range in.Roles {
-		result.RoleNames = append(result.RoleNames, string(role))
+		roles := result.GetRoleNames()
+		roles = append(roles, string(role))
+		result.SetRoleNames(roles)
 	}
```
igor-karpukhin, line 41: Wouldn't that be easier?
```
	for _, role := range in.Roles {
		roleNames = append(roleNames, string(role))
	}
	result.SetRoleNames(roleNames)
```
In this way, we won't call Get and Set on `result` on every loop iteration.

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -273,31 +274,35 @@ func fetchTeamByName(workflowCtx *workflow.Context, teamName string) (*mongodbat
...
-	atlasTeam, _, err := workflowCtx.Client.Teams.Create(workflowCtx.Context, workflowCtx.OrgID, atlasTeam)
+	atlasTeam, _, err := workflowCtx.SdkClient.TeamsApi.CreateTeam(workflowCtx.Context, workflowCtx.OrgID, atlasTeam).Execute()
 	if err != nil {
 		return nil, err
 	}
-
-	return atlasTeam, nil
+	response := &admin.TeamResponse{}
+	response.SetId(atlasTeam.GetId())
+	response.SetLinks(atlasTeam.GetLinks())
```
igor-karpukhin, line 288: nit: I think settings links is redundant here

igor-karpukhin (approved):
LGTM from my side if the CI is green. Great job @cveticm! my comments are non-blocking

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -186,22 +187,22 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
-			g.Go(func() error {
-				workflowCtx.Log.Debugf("removing user %s from team %s", user.ID, teamID)
-				_, err := workflowCtx.Client.Teams.RemoveUserToTeam(taskContext, workflowCtx.OrgID, teamID, user.ID)
-
-				return err
-			})
+	if atlasUsers.Results != nil {
+		for _, user := range atlasUsers.GetResults() {
+			if _, ok := usernamesMap[user.Username]; !ok {
+				g.Go(func() error {
```
josvazg, line 200: Not sure why we added concurrency here, we are launching a series of deletions that we are going to wait for. Are they really so slow? Only if this takes longer than 5-15 seconds total it might make sense.

Anyway, this is not a comment about this PR in particular.

josvazg (commented):
LGTM overall, I am holding my approval on checking about that project ID

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -26,6 +26,7 @@ import (
 func (r *AtlasProjectReconciler) teamReconcile(
 	team *akov2.AtlasTeam,
 	connectionSecretKey *client.ObjectKey,
+	projectID string,
 ) reconcile.Func {
```
josvazg, line 30: Remove unused extra public argument.

```suggestion
func (r *AtlasProjectReconciler) teamReconcile(
	team *akov2.AtlasTeam,
	connectionSecretKey *client.ObjectKey,
) reconcile.Func {
```

pkg/controller/atlasproject/teams.go
```diff
@@ -36,7 +36,7 @@ func (r *AtlasProjectReconciler) ensureAssignedTeams(workflowCtx *workflow.Conte
 		}
 
 		team := &akov2.AtlasTeam{}
-		teamReconciler := r.teamReconcile(team, project.ConnectionSecretObjectKey())
+		teamReconciler := r.teamReconcile(team, project.ConnectionSecretObjectKey(), project.ID())
```
josvazg, line 39: Remove unused arg also in invocation.

```suggestion
		teamReconciler := r.teamReconcile(team, project.ConnectionSecretObjectKey())
```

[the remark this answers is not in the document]
josvazg: Definitely not used, lest remove.

josvazg (changes requested):
Requesting removal of unused argument in public method. Else LGTM

pkg/controller/atlasproject/team_reconciler.go
cveticm: I'm not sure I understand what you mean by this, could you expand? I'm not sure how this constitutes as concurrency.

josvazg: `g.Go(func() error {`
https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go
This launches a **goroutine** on an concurrent `errgroup` with the main thread. Then the main thread will check all the returns to see if there are any errors from any of the goroutines.

I wonder how useful or required it is, the only gain is that you wait only for the slowest request, but this does not seem critical here, unless I am missing something.

Again not a critique to this PR. Just noticed we were doing this, I was not aware before and seems odd.

josvazg (approved):

pkg/controller/atlasproject/team_reconciler.go
cveticm: I see. The implementation of the translation layer would be a natural place to reconsider this approach. I'll bring this up in one of the upcoming sync-up optionals to see if the rest of the team has any insights on this detail.

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -273,31 +273,35 @@ func fetchTeamByName(workflowCtx *workflow.Context, teamName string) (*mongodbat
...
+func createTeam(workflowCtx *workflow.Context, atlasTeam *admin.Team) (*admin.TeamResponse, error) {
 	workflowCtx.Log.Debugf("create team named %s in atlas", atlasTeam.Name)
-	atlasTeam, _, err := workflowCtx.Client.Teams.Create(workflowCtx.Context, workflowCtx.OrgID, atlasTeam)
+	atlasTeam, _, err := workflowCtx.SdkClient.TeamsApi.CreateTeam(workflowCtx.Context, workflowCtx.OrgID, atlasTeam).Execute()
 	if err != nil {
 		return nil, err
 	}
-
-	return atlasTeam, nil
+	response := &admin.TeamResponse{}
```
s-urbaniak, line 285: this converts an internal SDK type to another internal SDK type. Let's try to avoid this. Rather, just return `atlasTeam` from the `CreateTeam` invocation directly.

pkg/controller/atlasproject/team_reconciler.go
helderjs: The request is usually fast, but a team can be a single user or an entire org. As there's no bulk deletion on the API this is just to be a little quicker in the worst scenario. I don't get why it is odd.

josvazg: Rule of thumb is avoid concurrency unless absolutely necessary, it adds complexity and possible race conditions hard to debug, etc. I do not see a particular risk here, we are just removing and reading back the results, but still I do not see the gain either, measured against the added complexity. That is all.

s-urbaniak (approved):

PR #1829, opened 2024-09-17
CLOUDP-273419: Don't delete teams when not assigned to a project

helderjs:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasproject/teams.go
+++ b/pkg/controller/atlasproject/teams.go
@@ -188,26 +188,19 @@ func (r *AtlasProjectReconciler) updateTeamState(ctx *workflow.Context, project 
 	conditions := akov2.InitCondition(team, api.FalseCondition(api.ReadyType))
 	teamCtx := workflow.NewContext(log, conditions, ctx.Context)
 
-	atlasClient, orgID, err := r.AtlasProvider.Client(teamCtx.Context, project.ConnectionSecretObjectKey(), log)
-	if err != nil {
-		return err
-	}
-	teamCtx.Client = atlasClient
-
 	if len(assignedProjects) == 0 {
-		if customresource.IsResourcePolicyKeepOrDefault(project, r.ObjectDeletionProtection) {
-			log.Debug("team %s has no project associated, "+
-				"skipping deletion from Atlas due to ObjectDeletionProtection being set", team.Spec.Name)
-		} else {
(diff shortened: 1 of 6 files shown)


josvazg (approved):
LGTM

pkg/controller/atlasproject/teams.go
```diff
@@ -188,26 +188,19 @@ func (r *AtlasProjectReconciler) updateTeamState(ctx *workflow.Context, project
...
-			log.Debugf("team %s has no project associated to it. removing from atlas.", team.Spec.Name)
-			_, err = teamCtx.Client.Teams.RemoveTeamFromOrganization(ctx.Context, orgID, team.Status.ID)
-			if err != nil {
-				return err
-			}
+		if err = customresource.ManageFinalizer(ctx.Context, r.Client, team, customresource.UnsetFinalizer); err != nil {
+			return err
+		}
+
+		teamCtx.SetConditionTrueMsg(api.TeamUnmanaged, "This resource is only reconciled when associated to a project")
```
roothorp, line 196: If we set this to False (or set the 'Ready' to False?) then it should be more clear that the team is in an undesirable state. Otherwise having the resource appear ready when it is blocked from reconciling may be misleading.

roothorp (commented):
Just one question. Love all the testing!

pkg/controller/atlasproject/teams.go
helderjs: My interpretation is that having a `TeamUmanaged` condition doesn't mean the team is in an undesirable state. It just makes clear to users that without being referred by a project a team is not reconciled.
For example, a newly added team resource will not be reconciled until it gets referred by a project, and this doesn't mean the resource is in the wrong state. This behavior might be controversial but it's how it was designed and it works like this since it was implemented.

roothorp: This makes sense, I’m more concerned about how clear it is that the resource doesn’t do anything in this state; changes aren’t reflected in Atlas. I figured having a “bad” condition might make it more clear that this is (probably) undesirable.

helderjs: Kube API conventions recommend some rules for a good condition.
* Condition type names should make sense for humans
* Condition type names should describe the current observed state of the resource
I think `TeamUnamaneged` covers both recommendations.
Also, it says the status of the condition can either have positive or negative polarity, to better reflect the semantics of the condition name.

roothorp (approved):

PR #1831, opened 2024-09-17
[DO NOT MERGE] Validate Manifest Test

helderjs:
### All Submissions:


Makefile
```diff
@@ -238,24 +238,23 @@ generate: ${GO_SOURCES} ## Generate code
...
+else
+	$(info Check: PASS)
 endif
-	@echo "--------------------"
-	@echo "Check: PASS"
 
 .PHONY: validate-manifests
-validate-manifests: generate manifests check-missing-files
+validate-manifests: generate manifests
+	$(MAKE) check-missing-files
```
josvazg, line 257: Is it worth to add a comment here saying that putting `check-missing-files` as a rule dependency was not working?

josvazg (approved):

Makefile
```diff
@@ -238,24 +238,23 @@ generate: ${GO_SOURCES} ## Generate code
...
-	@echo ""
-	@echo "Try running: make generate manifests"
-	@exit 1
+.PHONY: check-missing-files
+check-missing-files: ## Check autogenerated runtime objects and manifest files are missing
+	$(info Checking autogenerated runtime objects and manifests)
+	$(info ============================================)
+ifneq ($(shell git ls-files -o -m --directory --exclude-standard --no-empty-directory),)
+	$(info Detected files that need to be committed:)
+	$(info $(shell git ls-files -o -m --directory --exclude-standard --no-empty-directory | sed -e "s/^/  /"))
```
igor-karpukhin, line 247: nit: I'd be very careful with `sed` as unix and linux versions have significant differences. That's why we tend to use `awk` instead of `sed`. Feel free to skip this comment if this particular command works both on MacOS and Linux.

helderjs: I didn't check on Mac but here specifically is only formatting output, it has no influence on the evaluation of the task.

igor-karpukhin (approved):

PR #1833, opened 2024-09-18
CLOUDP-274200: internal/cmp: normalize empty/nil slices

s-urbaniak:
Currently, nil and empty slices are considered to be different. This fixes it.

Note: This is a prerequisite for normalizing data federation.

### All Submissions:

### DIFF
--- a/internal/cmp/normalize.go
+++ b/internal/cmp/normalize.go
@@ -94,6 +94,11 @@ func traverseValue(value reflect.Value, f func(slice reflect.Value)) {
 		}
 
 	case reflect.Slice:
+		// initialize empty slices to nil for normalization to be comparable
+		if value.Len() == 0 && value.CanSet() {
+			value.Set(reflect.Zero(value.Type()))
+		}
+
 		// omit zero length slices
 		if value.Len() == 0 {
 			return
--- a/internal/cmp/normalize_test.go
+++ b/internal/cmp/normalize_test.go
@@ -27,6 +27,45 @@ func TestEndlessRecursion(t *testing.T) {
 			},
 		},
 		{
+			name: "settable and addressable nil slice",
+			data: &struct {
+				Slice []string
+			}{
+				Slice: []string{},
+			},
+			want: &struct {
+				Slice []string
+			}{
+				Slice: nil,
+			},
+		},
+		{
+			name: "unsettable nil slice",
+			data: &struct {
+				slice []string
+			}{
+				slice: []string{},
+			},
+			want: &struct {
+				slice []string
+			}{
+				slice: []string{},
+			},
+		},
+		{
+			name: "unaddressable nil slice",
+			data: struct {
+				Slice []string
+			}{
+				Slice: []string{},
+			},
+			want: struct {
+				Slice []string
+			}{
+				Slice: []string{},
+			},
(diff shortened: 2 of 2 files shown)


internal/cmp/normalize.go
```diff
@@ -94,6 +94,11 @@ func traverseValue(value reflect.Value, f func(slice reflect.Value)) {
 		}
 
 	case reflect.Slice:
+		// initialize empty slices to nil for normalization to be comparable
+		if value.Len() == 0 && value.CanSet() {
+			value.Set(reflect.Zero(value.Type()))
+		}
```
josvazg, line 100: Why not the other way around, I mean, nil is not iterable, while empty it. Empty arrays are usually less problematic.

nit pick question, BTW it is fine by me either way.

internal/cmp/normalize_test.go
```diff
@@ -26,6 +26,45 @@ func TestEndlessRecursion(t *testing.T) {
...
+				Slice []string
+			}{
+				Slice: []string{},
+			},
+			want: struct {
+				Slice []string
+			}{
+				Slice: []string{},
+			},
+		},
```
josvazg, line 67: How is this case different from `"settable and addressable nil slice"` ?

josvazg: ah wait, this is an inline struct, not a pointer.

josvazg (approved):

internal/cmp/normalize_test.go
s-urbaniak: `struct` vs `&struct`

internal/cmp/normalize.go
s-urbaniak: > nil is not iterable, while empty it

`nil` indeed is iterable

s-urbaniak: https://go.dev/play/p/murKIez6rpW

josvazg: TIL 
Thanks!

roothorp:
Is there a scenario in which we care to distinguish between empty and nil? I understand Atlas uses it to differentiate between something being set (and just empty) or unset, but I dont know if that applies to us.

roothorp (approved):

josvazg:
> Is there a scenario in which we care to distinguish between empty and nil? I understand Atlas uses it to differentiate between something being set (and just empty) or unset, but I dont know if that applies to us.

Atlas may care about nil vs empty, but Kubernetes does not. So the idea is that such issue needs to be resolved in the translation layer when doing `toAtlas` / and `fromAtlas` conversion. I understand it is outside of the scope of this generic package. Did I get it right @s-urbaniak ?

s-urbaniak:
@roothorp you bring up a good point actually. I am not aware of any instance us using the functionality of "nil" vs "empty" in Atlas and likewise not sure this applies to us. So far I have not seen an instance in our translation layer where this actually matters. Do you have concrete examples?

@josvazg well, Kubernetes exposes whatever you set in custom resources. The real question is whether it really matters semantically to us and Atlas

s-urbaniak:
@roothorp @josvazg I think you have a point, closing for now. This is dangerous.

PR #1850, opened 2024-10-07
CLOUDP-276151: Migrate Assigned Teams Translation Layer

cveticm:
[CLOUDP-276151](https://jira.mongodb.org/browse/CLOUDP-276151): Implementation of translation layer for assigned teams.


pkg/api/v1/project_teams.go
```diff
@@ -28,7 +28,7 @@ type Team struct {
 }
 
 func (in *Team) ToAtlas(teamID string) admin.TeamRole {
-	roleNames := make([]string, 0, len(in.Roles))
+	roleNames := make([]string, 0)
```
s-urbaniak, line 31:
```suggestion
	roleNames := make([]string, 0, len(in.Roles))
```

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,131 @@
...
+	Username string
+	UserID   string
+}
+
+func NewTeam(projTeamSpec *akov2.Team, teamID string) *Team {
+	if projTeamSpec == nil {
+		return nil
+	}
+
+	roles := make([]string, 0)
```
s-urbaniak, line 33:
```suggestion
	roles := make([]string, 0, len(projTeamSpec.Roles))
```

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/teams"
```
s-urbaniak, line 23: please move this import next to the other mongodb-atlas-kubernetes imports.

pkg/controller/atlasproject/atlasproject_controller.go
```diff
@@ -137,7 +140,7 @@ func (r *AtlasProjectReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	workflowCtx.SdkClient = atlasSdkClient
 	r.projectService = project.NewProjectAPIService(atlasSdkClient.ProjectsApi)
-
+	r.teamsService = teams.NewTeamsAPIService(atlasSdkClient.TeamsApi, atlasSdkClient.MongoDBCloudUsersApi)
 	atlasClient, _, err := r.AtlasProvider.Client(workflowCtx.Context, atlasProject.ConnectionSecretObjectKey(), log)
```
s-urbaniak, line 144: is the old client still needed? If not, we can remove it.

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -4,9 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/http"
 	"sync"
 
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/teams"
```
s-urbaniak, line 9: see above, move the package import below

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -132,53 +133,49 @@ func (r *AtlasProjectReconciler) teamReconcile(
...
-	if err != nil {
-		return "", workflow.Terminate(workflow.TeamNotCreatedInAtlas, err.Error())
 	}
 
-	if atlasTeamResponse == nil {
-		atlasTeam, err := team.ToAtlas()
-		if err != nil {
+	if atlasTeam == nil {
+		desiredAtlasTeam := teams.NewAssignedTeam(&team.Spec, team.Status.ID)
+		if desiredAtlasTeam == nil {
```
s-urbaniak, line 159: this won't work. You don't have an `err` in the first place for the `workflow.Terminate`.

pkg/controller/atlasproject/teams.go
```diff
@@ -1,10 +1,13 @@
 package atlasproject
 
 import (
-	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/types"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/teams"
```
s-urbaniak, line 9: move package import below

pkg/controller/atlasproject/teams.go
```diff
@@ -67,7 +70,11 @@ func (r *AtlasProjectReconciler) ensureAssignedTeams(workflowCtx *workflow.Conte
 
 func (r *AtlasProjectReconciler) syncAssignedTeams(ctx *workflow.Context, projectID string, project *akov2.AtlasProject, teamsToAssign map[string]*akov2.Team) error {
 	ctx.Log.Debug("fetching assigned teams from atlas")
-	atlasAssignedTeams, _, err := ctx.SdkClient.TeamsApi.ListProjectTeams(ctx.Context, projectID).Execute()
+	if r.teamsService == nil {
```
s-urbaniak, line 73: the question is whether we really want to assert this and return an error. If we return an error, the reconciler will set a status on the resource and reenqueue.

if we don't catch this case, the code will actually panic and will cause the operator to shut down.

given the case of `r.teamsService` being `nil` is actually a programming error, I suggest to let it indeed panic.

pkg/controller/atlasproject/teams.go
```diff
@@ -80,76 +87,73 @@ func (r *AtlasProjectReconciler) syncAssignedTeams(ctx *workflow.Context, projec
 
 	defer statushandler.Update(ctx, r.Client, r.EventRecorder, project)
 
-	toDelete := make([]*admin.TeamRole, 0, len(atlasAssignedTeams.GetResults()))
+	toDelete := make([]*teams.Team, 0)
```
s-urbaniak, line 90:
```suggestion
	toDelete := make([]*teams.Team, 0, len(atlasAssignedTeams))
```

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, role := range projTeamSpec.Roles {
+		roles = append(roles, string(role))
+	}
+
+	team := &Team{
+		Roles:  roles,
+		TeamID: teamID,
```
josvazg, line 40: Why is `TeamName` not set here?

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,131 @@
...
+	return team
+}
+
+func TeamFromAtlas(team *admin.TeamResponse) *Team {
+	if team == nil {
+		return nil
+	}
+	tm := &Team{
+		TeamID:   team.GetId(),
+		TeamName: team.GetName(),
```
josvazg, line 70: Are roles missing to be set here?

s-urbaniak (changes requested):
there is one bug lurking in the comment at https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1850/files#r1790360836.

Else, this looks good! I would leave the final call to other reviewers as well.

pkg/api/v1/project_teams.go
cveticm: @s-urbaniak I've been confused about this. If we pre-allocate a slice of some length and then we append onto this slice, we effectively have a slice which is twice the length we actually need, with the first half full of nil/"" (created at make([]string, 0, len(in,Roles) ) and the other half with the actual information. Is this not a correct assumption of the behaviour of make and append?

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,131 @@
...
+
+func TeamToAtlas(team *akov2.Team, teamID string) *Team {
+	roleNames := make([]string, 0)
+	for _, role := range team.Roles {
+		roleNames = append(roleNames, string(role))
+	}
+	return &Team{
+		TeamID: teamID,
+		Roles:  roleNames,
+	}
```
josvazg, line 83: The role names conversion could be delegated so that conversions look more regular.

Also does it make sense to return nil if the input team is nil here? I am not sure myself, just asking

```suggestion
	if team == nil {
	  return nil
	}
	return &Team{
		TeamID: teamID,
		Roles:  RolesToAtlas(team.Roles),
	}
```

internal/translation/teams/conversion.go
cveticm: Type admin.TeamResponse does not hold team roles. Would setting Roles to an empty slice be safer behaviour in this case @josvazg ?

internal/translation/teams/conversion.go
cveticm: Noted, will add! And I think a nil check is a good idea, I'll add this to the other ToAtlas functions as well

internal/translation/teams/teams.go
```diff
@@ -0,0 +1,130 @@
...
+	Assign(ctx context.Context, at *[]Team, projectID string) error
+	Unassign(ctx context.Context, projectID, teamID string) error
+	Create(ctx context.Context, at *AssignedTeam, orgID string) (*Team, error)
+
+	GetTeamUsers(ctx context.Context, orgID, teamID string) ([]TeamUser, error)
+	UpdateRoles(ctx context.Context, at *Team, projectID string, newRoles []akov2.TeamRole) error
+	AddUsers(ctx context.Context, usersToAdd *[]admin.AddUserToTeam, orgID, teamID string) error
+	RemoveUser(ctx context.Context, orgID, teamID, userID string) error
+	RenameTeam(ctx context.Context, at *Team, orgID, newName string) (*Team, error)
+}
```
josvazg, line 26: I would reorder a bit for clarity on type managed:
```suggestion
type AtlasTeamsService interface {
	// Team  
	ListProjectTeams(ctx context.Context, projectID string) ([]Team, error)
	GetTeamByName(ctx context.Context, orgID, teamName string) (*Team, error)
	GetTeamByID(ctx context.Context, orgID, teamID string) (*Team, error)
	Assign(ctx context.Context, at *[]Team, projectID string) error
	Unassign(ctx context.Context, projectID, teamID string) error
	Create(ctx context.Context, at *AssignedTeam, orgID string) (*Team, error)
	UpdateRoles(ctx context.Context, at *Team, projectID string, newRoles []akov2.TeamRole) error
	RenameTeam(ctx context.Context, at *Team, orgID, newName string) (*Team, error)

	// TeamUser
	GetTeamUsers(ctx context.Context, orgID, teamID string) ([]TeamUser, error)
	AddUsers(ctx context.Context, usersToAdd *[]admin.AddUserToTeam, orgID, teamID string) error
	RemoveUser(ctx context.Context, orgID, teamID, userID string) error
}
```

Or maybe even separate interfaces joined by embedding:
```go
type AtlasTeamsService interface {
    TeamService
    TeamUserService
}
```

pkg/controller/atlasproject/teams.go
cveticm: That makes sense! To clarify, you suggest to remove this nil check entirely as the operator will panic when calling r.teamsService.ListProjectTeams in the case that teamsService==nil?

internal/translation/teams/conversion.go
cveticm: The only context in which this function is used is for calling to API request AddAllProjectTeams which only requires teamID and Roles.

pkg/controller/atlasproject/teams.go
josvazg: To give more context, we should standardise around just 3 groups of imports max, in order:

1. stdlib
2. other deps
3. ako internal deps 

`goimports` will respect our groupings and reorder each group alphabetically.

Sometimes when you change something, a new line is added and we end up with more groups than needed, and the go formatter does not fix it for us.

internal/translation/teams/conversion.go
josvazg: Empty slice and nil should be the same for a regular loop. On top of that, probably not being used.

Seems them this is only handling a Team name-id pair. I guess it does not make sense for us to have a separate type for just the name and id, without roles, and embed it in the team with assigned roles. In other words, if that is not used, I guess it is fine like this.

josvazg (commented):
LGTM overall, good job!

waiting on that bug fix and the CI to go green for approval

pkg/api/v1/project_teams.go
cveticm: I've investigated this and I understand now. This confusion came about when I had initialised a var as follows `mySlice := make([]string, len(otherSlice))`, resulting in a slice full of "" strings and all subsequent additions were appended to the end of this. In actuality, with the implementation you've provided, len(in.Roles) strings of memory is allocated to the var but the slice is of size 0. All appends to the slice (within this length) will be stored within this allocated memory. 👍

internal/translation/teams/teams.go
```diff
@@ -0,0 +1,130 @@
...
+func NewTeamsAPIService(teamAPI admin.TeamsApi, userAPI admin.MongoDBCloudUsersApi) *TeamsAPI {
+	return &TeamsAPI{
+		teamsAPI:     teamAPI,
+		teamUsersAPI: userAPI,
+	}
+}
+
+func (tm *TeamsAPI) ListProjectTeams(ctx context.Context, projectID string) ([]Team, error) {
+	atlasAssignedTeams, _, err := tm.teamsAPI.ListProjectTeams(ctx, projectID).Execute()
+	if err != nil {
```
helderjs, line 42: Why doesn't return AssignedTeam? this relates to the other comment regarding sematics

helderjs (changes requested):

internal/translation/teams/teams.go
```diff
@@ -0,0 +1,130 @@
...
+	"context"
+	"fmt"
+	"net/http"
+
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+)
+
+type AtlasTeamsService interface {
```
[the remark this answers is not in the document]
cveticm: Fixed 👍

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,143 @@
...
+	TeamName string
+	TeamID   string
+}
+
+type AssignedTeam struct {
+	Usernames []string
+	TeamName  string
+	TeamID    string
+}
+
```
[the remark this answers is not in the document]
cveticm: Noted, I've adjusted these structs

internal/translation/teams/teams.go
cveticm: Fixed 👍

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,143 @@
...
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+)
+
+type Team struct {
+	Usernames []string
+	TeamID    string
+	TeamName  string
+}
+
+type AssignedTeam struct {
```
josvazg, line 17: I find it confusing with the current naming.

The Team type above is named the same way as in Atlas: `Team` <-> `admin.Team` which is expected and straigh forward.

But here `AssignedTeam` <-> `TeamRole` feels confusing.

I understand the SDK naming is suboptimal, it is named `admin.TeamRole` when it actually means "TeamRoles".

My suggestion would be to rename as such: `TeamRoles` <-> `admin.TeamRole`

```suggestion
type TeamRoles struct {
```

That would make the conversion function names clearer as well:

`TeamRolesToAtlas` & `TeamRolesFromAtlas` always using our name, not the external name.

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,143 @@
...
+
+func TeamToAtlas(team *Team) *admin.Team {
+	return &admin.Team{
+		Id:        pointer.MakePtrOrNil(team.TeamID),
+		Name:      team.TeamName,
+		Usernames: &team.Usernames,
+	}
+}
+
+func AssignedTeamFromAtlas(team *admin.TeamResponse) *AssignedTeam {
```
josvazg, line 79:
```suggestion
func TeamRolesFromAtlas(team *admin.TeamResponse) *TeamRoles {
```

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,143 @@
...
+func AssignedTeamToAtlas(team *akov2.Team, teamID string) *AssignedTeam {
+	if team == nil {
+		return nil
+	}
+
+	return &AssignedTeam{
+		TeamID: teamID,
+		Roles:  RolesToAtlas(team.Roles),
+	}
+}
```
josvazg, line 100: This is unused right? also it does not produce an Atlas result, as the name implies. 

I would just remove:
```suggestion
```

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,143 @@
...
+
+func TeamRoleFromAtlas(atlasTeams []admin.TeamRole) []AssignedTeam {
+	teams := make([]AssignedTeam, 0, len(atlasTeams))
+	for _, team := range atlasTeams {
+		teams = append(teams, AssignedTeam{Roles: team.GetRoleNames(), TeamID: team.GetTeamId()})
+	}
+	return teams
+}
+
+func TeamRoleToAtlas(atlasTeams []AssignedTeam) []admin.TeamRole {
```
josvazg, line 110:
```suggestion
func TeamRolesToAtlas(teamRolesList []TeamRoles) []admin.TeamRole {
```

internal/translation/teams/teams.go
```diff
@@ -0,0 +1,130 @@
...
+	Assign(ctx context.Context, at *[]AssignedTeam, projectID string) error
+	Unassign(ctx context.Context, projectID, teamID string) error
+	Create(ctx context.Context, at *Team, orgID string) (*Team, error)
+	RenameTeam(ctx context.Context, at *AssignedTeam, orgID, newName string) (*AssignedTeam, error)
+	UpdateRoles(ctx context.Context, at *AssignedTeam, projectID string, newRoles []akov2.TeamRole) error
+
+	GetTeamUsers(ctx context.Context, orgID, teamID string) ([]TeamUser, error)
+	AddUsers(ctx context.Context, usersToAdd *[]admin.AddUserToTeam, orgID, teamID string) error
+	RemoveUser(ctx context.Context, orgID, teamID, userID string) error
+}
```
josvazg, line 26: Trying to order my head around these, it would seem we are actually wrapping 3 sub services.

type TeamsService interface {
  TeamProjectsService
  TeamRolesService
  TeamMembersService
}

type TeamProjectsService interface { // manages Team's associations to Projects
  	ListProjectTeams(ctx context.Context, projectID string) ([]TeamRoles, error)
	Assign(ctx context.Context, at *[]TeamRoles, projectID string) error
	Unassign(ctx context.Context, projectID, teamID string) error
}

type TeamRolesService interface { // manages Team's Roles
	GetTeamByName(ctx context.Context, orgID, teamName string) (*TeamRoles, error)
	GetTeamByID(ctx context.Context, orgID, teamID string) (*TeamRoles, error)
	RenameTeam(ctx context.Context, at *TeamRoles, orgID, newName string) (*TeamRoles, error)
	UpdateRoles(ctx context.Context, at *TeamRoles, projectID string, newRoles []akov2.TeamRole) error
}

Type TeamMembersService interface { // manages Team's Members (Users)
	GetTeamUsers(ctx context.Context, orgID, teamID string) ([]TeamUser, error)
	AddUsers(ctx context.Context, usersToAdd *[]admin.AddUserToTeam, orgID, teamID string) error
	RemoveUser(ctx context.Context, orgID, teamID, userID string) error
}

internal/translation/teams/teams.go
```diff
@@ -0,0 +1,130 @@
...
+	GetTeamByName(ctx context.Context, orgID, teamName string) (*AssignedTeam, error)
+	GetTeamByID(ctx context.Context, orgID, teamID string) (*AssignedTeam, error)
+	Assign(ctx context.Context, at *[]AssignedTeam, projectID string) error
+	Unassign(ctx context.Context, projectID, teamID string) error
+	Create(ctx context.Context, at *Team, orgID string) (*Team, error)
+	RenameTeam(ctx context.Context, at *AssignedTeam, orgID, newName string) (*AssignedTeam, error)
+	UpdateRoles(ctx context.Context, at *AssignedTeam, projectID string, newRoles []akov2.TeamRole) error
+
+	GetTeamUsers(ctx context.Context, orgID, teamID string) ([]TeamUser, error)
+	AddUsers(ctx context.Context, usersToAdd *[]admin.AddUserToTeam, orgID, teamID string) error
```
josvazg, line 24: The `admin.` types should not leak out of the translation layer.

```suggestion
	AddUsers(ctx context.Context, users []TeamUser, orgID, teamID string) error
```

internal/translation/teams/conversion.go
helderjs: The sdk makes it confusing, a team is assigned to a project and it has a set of roles which apply to the project the team is being assigned.

In my opinion the rename would wrongly represent the involved objects.

josvazg: `TeamProjectRoles` then? Assigned Team does not tell you what is assigned to.

cveticm: Although I do understand the confusion with `AssignedTeam` <-> `admin.teamRole`, but I would agree with Helder that TeamRole doesn't really capture the object well. As teams can only be assigned to projects, I think it is fair to assume `AssignedTeam` would imply the team is assigned to the project we are reconciling. AssignedTeamRoles may make the relationship between the object and `admin.teamRole` more clear. Thoughts @josvazg @helderjs ?

internal/translation/teams/teams.go
cveticm: Agreed, this is more clear. Will implement 👍

internal/translation/teams/teams.go
cveticm: Noted! Fixing now

internal/translation/teams/conversion.go
josvazg: `ProjectTeamRoles` seems the most accurate to me.
Still I will not block this in naming, I leave Helder the unblocking vote.

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func TeamRolesToAtlas(atlasTeams []AssignedTeam) []admin.TeamRole {
+	if atlasTeams == nil {
+		return nil
+	}
+	teams := make([]admin.TeamRole, 0, len(atlasTeams))
+
+	for _, team := range atlasTeams {
+		result := admin.TeamRole{
```
roothorp, line 106: This is a nit but we don't enter this loop in any tests - we should add/modify a test that actually converts roles to Atlas (I think this is only called in tests with an empty slice).

internal/translation/teams/conversion.go
```diff
@@ -0,0 +1,135 @@
...
+		})
+	}
+	return teamUsers
+}
+
+func UsersToAtlas(teamUsers *[]TeamUser) *[]admin.AddUserToTeam {
+	users := *teamUsers
+	desiredUsers := make([]admin.AddUserToTeam, 0, len(users))
+	for _, user := range users {
+		desiredUsers = append(desiredUsers, admin.AddUserToTeam{
```
roothorp, line 130: Same as above - I think this is only passed empty slices in tests

roothorp (commented):
One testing nit - otherwise looks great 👍

internal/translation/teams/conversion.go
cveticm: Good catch! Adding this coverage now 👍

internal/translation/teams/conversion.go
helderjs: IMHO, AssignedTeam reflect better the entity. The role is a property of the assigned team. I'm good with Melanija suggestion AssignedTeamRole if you both think role must be in object name

cveticm: I'll leave the naming as is then. As you previously mentioned, Helder, the SDK adds confusion by calling a team assigned to a project `TeamRole` and, hopefully `AssignedTeam` will mitigate this confusion when reading the reconcilers. Given Helder's vote, I'll mark this as resolved, feel free to unresolve if anyone wishes to discuss this further

josvazg (approved):
LGTM! Good job!

helderjs (approved):

roothorp (approved):

s-urbaniak (approved):

igor-karpukhin (approved):
Great job @cveticm! 🔥 LGTM 👍

pkg/controller/atlasproject/team_reconciler.go
```diff
@@ -188,22 +176,20 @@ func ensureTeamUsersAreInSync(workflowCtx *workflow.Context, teamID string, team
...
-					_, err := workflowCtx.SdkClient.TeamsApi.RemoveTeamUser(taskContext, workflowCtx.OrgID, teamID, user.GetId()).Execute()
-					return err
-				})
-			}
+	for _, user := range atlasUsers {
+		if _, ok := usernamesMap[user.Username]; !ok {
+			g.Go(func() error {
+				workflowCtx.Log.Debugf("removing user %s from team %s", user.UserID, teamID)
+				err := r.teamsService.RemoveUser(workflowCtx.Context, workflowCtx.OrgID, teamID, user.UserID)
+				return err
```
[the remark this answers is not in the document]
cveticm: For the sake of pushing this through merge and avoiding sitting through another set of flakey test runs, I'll dismiss this nit.

PR #1854, opened 2024-10-08
CLOUDP-277700: Reduced reconcile period for externally linked resources

helderjs:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
+++ b/pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
@@ -202,6 +202,10 @@ func (r *AtlasDatabaseUserReconciler) ready(ctx *workflow.Context, atlasDatabase
 		EnsureStatusOption(status.AtlasDatabaseUserNameOption(atlasDatabaseUser.Spec.Username)).
 		EnsureStatusOption(status.AtlasDatabaseUserPasswordVersion(passwordVersion))
 
+	if atlasDatabaseUser.Spec.ExternalProjectRef != nil {
+		return workflow.Requeue(workflow.StandaloneResourceRequeuePeriod).ReconcileResult()
+	}
+
 	return workflow.OK().ReconcileResult()
 }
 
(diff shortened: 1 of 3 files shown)


s-urbaniak (approved):

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
```diff
@@ -202,6 +202,10 @@ func (r *AtlasDatabaseUserReconciler) ready(ctx *workflow.Context, atlasDatabase
 		EnsureStatusOption(status.AtlasDatabaseUserNameOption(atlasDatabaseUser.Spec.Username)).
 		EnsureStatusOption(status.AtlasDatabaseUserPasswordVersion(passwordVersion))
 
+	if atlasDatabaseUser.Spec.ExternalProjectRef != nil {
+		return workflow.Requeue(workflow.StandaloneResourceRequeuePeriod).ReconcileResult()
+	}
+
 	return workflow.OK().ReconcileResult()
```
josvazg, line 209: This is the only place we exit OK right?

pkg/controller/workflow/result.go
```diff
@@ -7,8 +7,9 @@ import (
 )
 
 const (
-	DefaultRetry   = time.Second * 10
-	DefaultTimeout = time.Minute * 20
+	DefaultRetry                    = time.Second * 10
+	StandaloneResourceRequeuePeriod = time.Minute * 15
```
josvazg, line 11: We said we would make this configurable down to 5 mins minimum, but not advertised. We can do it on a follow up.

josvazg (approved):
LGTM
(I expect the CI would eventually pass)

pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go
helderjs: Yes

pkg/controller/workflow/result.go
josvazg: Thanks for the heads up, I have fixed the TD to reiterate what was said in the scope, this should be configurable before release and 5 minutes is the minimum value allowed.

PR #1856, opened 2024-10-10
CLOUDP-277876: Reap orphan connection secrets

josvazg:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasdatabaseuser/databaseuser.go
+++ b/pkg/controller/atlasdatabaseuser/databaseuser.go
@@ -194,9 +194,9 @@ func (r *AtlasDatabaseUserReconciler) readiness(ctx *workflow.Context, atlasProj
 		return r.terminate(ctx, atlasDatabaseUser, api.DatabaseUserReadyType, workflow.Internal, true, err)
 	}
 	if len(removedOrphanSecrets) > 0 {
-		r.Log.Info("Removed %d orphan secrets on project %s bound to an non existent deployment:")
+		r.Log.Debug("Removed %d orphan secrets on project %s bound to an non existent deployment:")
 		for _, orphan := range removedOrphanSecrets {
-			r.Log.Info("Removed orphan secret %q", orphan)
+			r.Log.Debug("Removed orphan secret %q", orphan)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/atlasdatabaseuser/databaseuser.go
```diff
@@ -188,6 +188,18 @@ func (r *AtlasDatabaseUserReconciler) readiness(ctx *workflow.Context, atlasProj
 		return r.terminate(ctx, atlasDatabaseUser, api.DatabaseUserReadyType, workflow.Internal, true, err)
 	}
 
+	removedOrphanSecrets, err := connectionsecret.ReapOrphanConnectionSecrets(
+		ctx.Context, r.Client, atlasProject.ID, atlasDatabaseUser.Namespace, allDeploymentNames)
+	if err != nil {
+		return r.terminate(ctx, atlasDatabaseUser, api.DatabaseUserReadyType, workflow.Internal, true, err)
+	}
+	if len(removedOrphanSecrets) > 0 {
+		r.Log.Info("Removed %d orphan secrets on project %s bound to an non existent deployment:")
```
helderjs, line 197: I'd log this as debug to avoid pollute logs with less relevant info.

helderjs (changes requested):

helderjs (approved):

pkg/controller/connectionsecret/connectionsecrets.go
```diff
@@ -20,6 +22,35 @@ import (
...
+	removedOrphanSecrets := []string{}
+	for _, secret := range secretList.Items {
+		clusterName, ok := secret.Labels[ClusterLabelKey]
+		if !ok {
+			continue
+		}
+		if clusterExists := stringutil.Contains(projectDeploymentNames, clusterName); clusterExists {
+			continue
+		}
+		if err := k8sClient.Delete(ctx, &secret); err != nil {
```
s-urbaniak, line 45: the method `ReapOrphanConnectionSecrets` is doing more than the reader is assuming. it is also causing the side-effect of deleting secrets. Either we:
1. factor deletion in its own function, not called in a `Read...` function
2. Rename this function to something that informs the reader that a deletion happens here as well

s-urbaniak (approved):

PR #1857, opened 2024-10-10
CLOUDP-277319: Added helm automation job

igor-karpukhin:
### All Submissions:


.github/workflows/update-helm.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+          yq '.spec.install.spec.clusterPermissions[0].rules' ./bundle/manifests/mongodb-atlas-kubernetes.clusterserviceversion.yaml > rbac.yaml
+
+          echo "Comparing RBAC for CSV to RBAC in AKO helm chart"
+          if ! diff rbac.yaml "$HELM_RBAC_FILE"; then
+            echo "Copying RBAC"
+            cp rbac.yaml "$HELM_RBAC_FILE"
+          else
+            echo "No changes detected"
+          fi
```
josvazg, line 72: Why not make this into an script (eg. `scripts/update-helm.sh`) and call it though `devbox` and `make`. Eg:
```
  run: devbox run --  'make update-helm'
```

That way all scripts dependencies will be using the environment set by our locked devbox dependencies. Also we are able to run and test this locally.

.github/workflows/update-helm.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+jobs:
+  verify-helm-changes:
+    name: Verify if AKO helm charts needs updates
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout AKO repo
+        uses: actions/checkout@v4
+
+      - name: Install YQ
```
josvazg, line 16: devbox takes care of this

igor-karpukhin: Is it comes preinstalled?

josvazg: with `devbox` it is already included.

.github/workflows/update-helm.yaml
josvazg: Precisely because it is a lot of bash I'd rather have it in a regular bash a script file, not even in make. It is easier to test and we can invoke outside of the CI as needed.

Whether it is one or 2 scripts, I don't have a strong opinion about that. The make file invocation can do both in one go: `run: devbox run -- 'make update-crds update-rbac'`

igor-karpukhin: I reverted it back to two bash scripts included into Makefile

.github/workflows/update-helm.yaml
```diff
@@ -0,0 +1,57 @@
...
+          HELM_CRDS_PATH: "./helm-charts-cloned/charts/atlas-operator-crds/templates"
+        run: |
+          devbox run -- "make helm-upd-crds"
+
+      - name: Verify if RBAC were changed
+        id: rbac-check
+        env:
+          HELM_RBAC_FILE: "./helm-charts-cloned/charts/atlas-operator/rbac.yaml"
+        run: |
+          devbox run -- "make helm-upd-rbac"
```
josvazg, line 45: where are these rules in the makefile? 
(same for `make helm-upd-crds`) I do not see the m¡Makefile changes.

igor-karpukhin: omg, forgot to add it. Fixed!

Makefile
```diff
@@ -431,24 +431,31 @@ test-metrics:
 .PHONY: test-tools ## Test all tools
 test-tools: test-clean test-makejwt test-metrics
 
-.PHONY: sign 
+.PHONY: sign
```
josvazg, line 434: Nice!

josvazg (approved):

.github/workflows/update-helm.yaml
```diff
@@ -0,0 +1,63 @@
...
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  verify-helm-changes:
+    name: Verify if AKO helm charts needs updates
+    runs-on: ubuntu-latest
+    env:
+      JWT_APP_ID: ${{ secrets.AKO_RELEASER_APP_ID }}
+      JWT_RSA_PEM_KEY_BASE64: ${{ secrets.AKO_RELEASER_RSA_KEY_BASE64 }}
```
josvazg, line 14: In the end the AKO releaser github app was needed then?

.github/workflows/update-helm.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Verify if RBAC were changed
+        id: rbac-check
+        env:
+          HELM_RBAC_FILE: "./helm-charts-cloned/charts/atlas-operator/rbac.yaml"
+        run: |
+          devbox run -- "make helm-upd-rbac"
+
+      - name: Create PR for helm-charts repo
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
josvazg, line 50: you are using the regular github secret, which i guess worked in your tests. We do not need to use the gh app then.

.github/workflows/update-helm.yaml
josvazg: Seems we do not need it.
```suggestion
```

josvazg (commented):
LGTM but seems we do not need to use the github app for this, so we can remove the JWT secrets.

PR #1860, opened 2024-10-14
CLOUDP-278387: Add index for dbUser by project and improve connection secret generation

helderjs:
### All Submissions:

### DIFF
--- a/pkg/indexer/atlasdatabaseuserprojects.go
+++ b/pkg/indexer/atlasdatabaseuserprojects.go
@@ -55,7 +55,9 @@ func (a *AtlasDatabaseUserByProjectIndexer) Keys(object client.Object) []string 
 			return nil
 		}
 
-		return []string{project.ID()}
+		if project.ID() != "" {
+			return []string{project.ID()}
+		}
 	}
 
 	return nil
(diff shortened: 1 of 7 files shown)


s-urbaniak (approved):
looks fine and dandy 👏

pkg/indexer/atlasdatabaseuserprojects.go
```diff
@@ -0,0 +1,62 @@
...
+	if user.Spec.Project != nil && user.Spec.Project.Name != "" {
+		project := &akov2.AtlasProject{}
+		err := a.client.Get(a.ctx, *user.Spec.Project.GetObject(user.Namespace), project)
+		if err != nil {
+			a.logger.Errorf("unable to find project to index: %s", err)
+
+			return nil
+		}
+
+		return []string{project.ID()}
```
s-urbaniak, line 58: nit: this can be an empty string silently.

s-urbaniak: if it is empty string, we shouldn't index, just return nil.

josvazg (commented):
Did I miss the reconciler setup in which the indexer is wired for production use?

helderjs:
> Did I miss the reconciler setup in which the indexer is wired for production use?

I think you are mixing up goals. We usually create index for watching purpose, in that case we setup the watcher to use the index to look up the watched object, however it is not the case here.
We are just indexing it to be able to quickly find db users that need connection secret to be generated

PR #1867, opened 2024-10-15
CLOUDP-278494: Fix conversion/comparison of shared clusters

helderjs:
### All Submissions:

**Note**: This change doesn't allow updating shared cluster tier (which is not allowed via public API) but process the change and  reflects the error on status.
image

### DIFF
--- a/internal/translation/deployment/conversion.go
+++ b/internal/translation/deployment/conversion.go
@@ -474,8 +474,8 @@ func labelsFromAtlas(cLabels []admin.ComponentLabel) []common.LabelSpec {
 }
 
 func replicationSpecFromAtlas(replicationSpecs []admin.ReplicationSpec) []*akov2.AdvancedReplicationSpec {
-	hSpecOrDefault := func(spec admin.HardwareSpec) *akov2.Specs {
-		if spec.GetNodeCount() == 0 {
+	hSpecOrDefault := func(spec admin.HardwareSpec, providerName string) *akov2.Specs {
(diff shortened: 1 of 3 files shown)


internal/translation/deployment/conversion.go
```diff
@@ -474,8 +474,8 @@ func labelsFromAtlas(cLabels []admin.ComponentLabel) []common.LabelSpec {
 }
 
 func replicationSpecFromAtlas(replicationSpecs []admin.ReplicationSpec) []*akov2.AdvancedReplicationSpec {
-	hSpecOrDefault := func(spec admin.HardwareSpec) *akov2.Specs {
-		if spec.GetNodeCount() == 0 {
+	hSpecOrDefault := func(spec admin.HardwareSpec, providerName string) *akov2.Specs {
+		if spec.GetNodeCount() == 0 && providerName != string(provider.ProviderTenant) {
```
josvazg, line 478: Why was this also necessary?

helderjs: during comparison, we were turning node spec with nodeCount 0 to nil. As the shared cluster has only instance size defined (nodeCount will be always 3 and the API omits the value :man_shrugging: ), we need to consider this when preparing it to compare.

josvazg: sorry, but what does the provider name have to do with the node count or instance size? I am missing that detail

helderjs: we identify shared cluster by provider name (TENANT)

josvazg (approved):

s-urbaniak (approved):

PR #1869, opened 2024-10-15
CLOUDP-278492: Avoid panic when changing type of existing cluster

helderjs:
### All Submissions:

### DIFF
--- a/pkg/controller/atlasdeployment/atlasdeployment_controller.go
+++ b/pkg/controller/atlasdeployment/atlasdeployment_controller.go
@@ -162,7 +162,7 @@ func (r *AtlasDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	wasDeleted := !atlasDeployment.GetDeletionTimestamp().IsZero()
 	existsInAtlas := deploymentInAtlas != nil
 
-	if existsInAtlas && fmt.Sprintf("%T", deploymentInAKO) != fmt.Sprintf("%T", deploymentInAtlas) {
+	if existsInAtlas && deploymentInAKO.IsServerless() != deploymentInAtlas.IsServerless() {
 		return r.terminate(workflowCtx, workflow.Internal, errors.New("regular deployment cannot be converted into a serverless deployment and vice-versa"))
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controller/atlasdeployment/atlasdeployment_controller.go
```diff
@@ -162,6 +162,10 @@ func (r *AtlasDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	wasDeleted := !atlasDeployment.GetDeletionTimestamp().IsZero()
 	existsInAtlas := deploymentInAtlas != nil
 
+	if existsInAtlas && fmt.Sprintf("%T", deploymentInAKO) != fmt.Sprintf("%T", deploymentInAtlas) {
```
s-urbaniak, line 165: No, please don't switch on string representations of types, this is not right. A low level `%T` language type primitive must not be used to assert high level business logic. You can accomplish this using different techniques.

I.e. you can extend the `translation.Deployment` interface with a `IsServerless()` method. In the case of `NewDeployment` it returns it based on the spec and in the case of the `GetDeployment` service it returns it based whether an underlying serverless or regular deployment was found.

s-urbaniak: the assertion would then change to something like:
```
	if existsInAtlas && deploymentInAKO.IsServerless() != deploymentInAtlas.IsServerless() {
```

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
```diff
@@ -1422,3 +1424,175 @@ func TestGetProjectFromKube(t *testing.T) {
 		})
 	}
 }
+
+func TestChangeDeploymentType(t *testing.T) {
+	tests := map[string]struct {
+		deployment *akov2.AtlasDeployment
+	}{
+		"should fail when existing cluster is regular but manifest defines a serverless instance": {
```
josvazg, line 1432: I was expecting two inputs; existing and new spec. How can this work with one input only?

helderjs: The input is an already-created deployment that will conflict with its mocked counterpart (lines 1534-1560) coming from Atlas.

s-urbaniak (approved):

pkg/controller/atlasdeployment/atlasdeployment_controller_test.go
josvazg: Fine. I'd rather make the test logic dumb simple and ensure the inputs are more explicit, but will not block the PR on this.

josvazg (approved):
LGTM, providing CI agrees

PR #1871, opened 2024-10-18
CLOUDP-279474: Use generator to avoid sharing list value

josvazg:
- 1st commit adds a test to detect the race condition.
- 2nd commit fixes the race condition by using a function that generates the empty list for each call, instead of reusing the same value for all calls.
 
### All Submissions:


s-urbaniak (commented):
Generally LGTM. but please add the unit tests as discussed OOB that proof the racy behavior.

s-urbaniak (approved):

PR #1873, opened 2024-10-21
CLOUDP-228586: Migrate Integrations

roothorp:
### All Submissions:


internal/translation/integrations/conversion.go
```diff
@@ -0,0 +1,106 @@
...
+		Type:                     pointer.MakePtr(in.Type),
+		AccountId:                pointer.MakePtr(in.AccountID),
+		Region:                   pointer.MakePtr(in.Region),
+		TeamName:                 pointer.MakePtr(in.TeamName),
+		ChannelName:              pointer.MakePtr(in.ChannelName),
+		Url:                      pointer.MakePtr(in.URL),
+		MicrosoftTeamsWebhookUrl: pointer.MakePtr(in.MicrosoftTeamsWebhookURL),
+		Username:                 pointer.MakePtr(in.UserName),
+		ServiceDiscovery:         pointer.MakePtr(in.ServiceDiscovery),
+		LicenseKey:               pointer.MakePtr(""),
```
josvazg, line 74: Setting nil or empty on the Atlas side sometimes is not the same. Contract testing would help with discerning those cases. Not saying it is required here, but we should just keep that in mind if we see some misbehaviour on e2e tests.

internal/translation/integrations/conversion.go
```diff
@@ -0,0 +1,106 @@
...
+			URL:                      in.GetUrl(),
+			MicrosoftTeamsWebhookURL: in.GetMicrosoftTeamsWebhookUrl(),
+			UserName:                 in.GetUsername(),
+			ServiceDiscovery:         in.GetServiceDiscovery(),
+			Enabled:                  in.GetEnabled(),
+		},
+	)
+}
+
+func toAtlas(in Integration, ctx context.Context, c client.Client, defaultNS string) (result *admin.ThirdPartyIntegration, err error) {
```
josvazg, line 63: I'd rather avoid the translation layer making calls to clients here. We are coupling the atlas side with the Kubernetes side like this.

As an alternative I would make use of the local type to also hold the passwords:

```go
type Integration struct {
	project.Integration
        LicenseKey string
        WriteToken string
        ...
}

Such type is populated properly before the translation.
```

internal/translation/integrations/conversion.go
```diff
@@ -0,0 +1,106 @@
...
+
+type Integration struct {
+	project.Integration
+}
+
+func (i Integration) Identifier() interface{} {
+	return i.Type
+}
+
+func NewIntegration(spec *project.Integration) (*Integration, error) {
```
josvazg, line 24: I would pass the passwords in here, do not make this conversion access k8s client.

internal/translation/integrations/conversion.go
```diff
@@ -0,0 +1,106 @@
...
+	if spec == nil {
+		return nil, nil
+	}
+	if err := cmp.Normalize(spec); err != nil {
+		return nil, fmt.Errorf("failed to normalize integration: %w", err)
+	}
+	return &Integration{Integration: *spec}, nil
+}
+
+func NewIntegrations(specs []project.Integration) ([]Integration, error) {
```
josvazg, line 34: I would remove this from here. Once we add the passwords as parameters in, this loop is probably better run with the reconcile code, getting all passwords and then passing them in.

internal/translation/integrations/conversion.go
```diff
@@ -0,0 +1,106 @@
...
+		integration, err := NewIntegration(&spec)
+		if err != nil {
+			return nil, err
+		}
+		list = append(list, *integration)
+	}
+	return list, nil
+}
+
+func fromAtlas(in *admin.ThirdPartyIntegration) (*Integration, error) {
```
josvazg, line 46: I guess is fine not to set the passwords coming from Atlas.

internal/translation/integrations/conversion.go
josvazg: Wait, I think this is only used for comparisons, so it might be fine without filling passwords.

Maybe we could use 2 separate types with and without passwords instead.

internal/translation/integrations/conversion.go
josvazg: We could be more explicit with 2 types:

```go
type Integration struct {
	project.Integration
}

type IntegrationWithCredentials struct {
	Integration
        LicenseKey string
        WriteToken string
        ...
}
```

Then the API calls also need to be explicit:

```go
Get(ctx context.Context, projectID string, integrationType string) (*Integration, error)
List(ctx context.Context, projectID string) ([]Integration, error)
Create(ctx context.Context, projectID string, integrationType string, integration IntegrationWithCredentials, client client.Client, ns string) error
Update(ctx context.Context, projectID string, integrationType string, integration IntegrationWithCredentials, client client.Client, ns string) error
```

josvazg (commented):
I really do not like mixing k8s code within the translation layer. I feel we should do password handling before we git it.
Let me know what you think of my suggestions.

josvazg (changes requested):
Holding until we clarify the password handling.

helderjs (changes requested):
I'd join Jose here. I think we should not have k8s client in the translation layer.
If passing the creds as a parameter od the conversion functions is an effort beyond the scope of the task, I recommend using a func type instead, something like:
`type SecretResolver func(ref *common.ResourceRefNamespaced) (map[string][]byte, error)`

The conversion will rely on it and not be bound to k8s client.

josvazg:
keep it alive!

PR #1880, opened 2024-10-22
CLOUDP-228587: Migrate Maintenance Window

roothorp:
### All Submissions:


internal/translation/maintenancewindow/conversion.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/pointer"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1/project"
+)
+
+type MaintenanceWindow struct {
+	*project.MaintenanceWindow
+}
```
josvazg, line 14: Why do we need the wrapping if we are not adding more fields?

internal/translation/maintenancewindow/conversion.go
```diff
@@ -0,0 +1,58 @@
...
+
+func NewMaintenanceWindow(spec *project.MaintenanceWindow) *MaintenanceWindow {
+	if spec == nil {
+		return nil
+	}
+	// No slices so no normalization needed
+	return &MaintenanceWindow{MaintenanceWindow: spec}
+}
```
josvazg, line 22: Without the wrapping we would also not need this.

internal/translation/maintenancewindow/conversion.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (mw *MaintenanceWindow) WithStartASAP(asap bool) *MaintenanceWindow {
+	mw.StartASAP = asap
+
+	return mw
+}
```
josvazg, line 28: This might also go away without wrapping.

internal/translation/maintenancewindow/conversion_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"regexp"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	fuzz "github.com/google/gofuzz"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRoundtrip_MaintenanceWindow(t *testing.T) {
```
josvazg, line 12: FYI gofuzz [supports integration with dvyukovgo-fuzz](https://github.com/google/gofuzz?tab=readme-ov-file#dvyukovgo-fuzz-integration) which in turns means we might be able to make our fuzz tests official go fuzz tests, with the language support that this brings.
No need to change for this one, but maybe we could explore it going forward.

josvazg (approved):
LGTM

About the wrapping, it is not a blocker, but also not sure of the benefit of wrapping such type in this particular case.

helderjs (approved):

internal/translation/maintenancewindow/conversion_test.go
s-urbaniak: that's indeed a nice integration. (non blocking for this PR)

internal/translation/maintenancewindow/conversion.go
```diff
@@ -0,0 +1,58 @@
...
+
+type MaintenanceWindow struct {
+	*project.MaintenanceWindow
+}
+
+func NewMaintenanceWindow(spec *project.MaintenanceWindow) *MaintenanceWindow {
+	if spec == nil {
+		return nil
+	}
+	// No slices so no normalization needed
```
s-urbaniak, line 20: very nice comment! 👏

PR #1883, opened 2024-10-23
CLOUDP-280713: Remove data federation PE explicitly at e2e test

igor-karpukhin:
### All Submissions:

### DIFF
--- a/test/e2e/datafederation_pe_test.go
+++ b/test/e2e/datafederation_pe_test.go
@@ -120,14 +120,6 @@ var _ = Describe("UserLogin", Label("datafederation"), func() {
 			}).WithTimeout(2 * time.Minute).WithPolling(20 * time.Second).Should(BeTrue())
 		})
 
-		By("Deleting DataFederation Private endpoint", func() {
-			_, resp, err := atlasClient.Client.DataFederationApi.DeleteDataFederationPrivateEndpoint(testData.Context,
-				testData.Project.ID(), pe.ID).Execute()
-			Expect(err).To(BeNil())
-			Expect(resp).NotTo(BeNil())
-			Expect(resp.StatusCode).To(BeEquivalentTo(http.StatusNoContent))
-		})
-
 		By("Delete DataFederation", func() {
 			df := &akov2.AtlasDataFederation{}
 			Expect(testData.K8SClient.Get(context.Background(), types.NamespacedName{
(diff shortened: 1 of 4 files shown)


s-urbaniak (changes requested):
@igor-karpukhin great catch! 🎉 

But please prefix the PR with a ticket.

josvazg:
> @igor-karpukhin great catch! 🎉
> 
> But please prefix the PR with a ticket.

I can take this over and create the ticket. I still see failures, so we might need a manual cleanup as well.

josvazg:
CLOUDP-280713

test/e2e/datafederation_pe_test.go
```diff
@@ -127,5 +128,26 @@ var _ = Describe("UserLogin", Label("datafederation"), func() {
...
+			Expect(vpe.GetResults()).To(BeEmpty())
+		})
+		By("Deleting DataFederation Private Endpoint again, as it does not do it the first time around", func() {
+			time.Sleep(5 * time.Second)
+			_, resp, err := atlasClient.Client.DataFederationApi.DeleteDataFederationPrivateEndpoint(testData.Context,
+				testData.Project.ID(), pe.ID).Execute()
+			Expect(err).To(BeNil())
+			Expect(resp).NotTo(BeNil())
+			Expect(resp.StatusCode).To(BeEquivalentTo(http.StatusNoContent))
+		})
```
josvazg, line 151: cc @s-urbaniak this does not make sense, but it works. Not sure why.

Still this is not an issue on our side, but behaviour on Atlas or the cloud provider AFAIU.

test/e2e/datafederation_pe_test.go
```diff
@@ -127,5 +130,24 @@ var _ = Describe("UserLogin", Label("datafederation"), func() {
...
+			// for some reason, doing it just once doesn't work
+			for i := 0; i < 5; i++ { // 3 retries or less seems to fail
+				_, resp, err := atlasClient.Client.DataFederationApi.DeleteDataFederationPrivateEndpoint(testData.Context,
+					testData.Project.ID(), pe.ID).Execute()
+				if admin.IsErrorCode(err, "RESOURCE_NOT_FOUND") {
+					msg := fmt.Sprintf("It says it Removed DataFederation Private Endpoint %q for project %s after retry %d\n",
+						pe.ID, testData.Project.ID(), i+1)
+					GinkgoWriter.Write([]byte(msg))
+					continue
+				}
```
igor-karpukhin, line 146: non-blocking: @josvazg you can replace it with `Eventually()`

josvazg: I would rather not change that. `Eventually` is used when you have a eventually consistent state to check for, which is not the case here. I would instead add a comment linking to the Atlas issue so we can revisit and fix this code later on, removing the for completely and, most likely not needing even to use `Eventually` either.

test/e2e/datafederation_pe_test.go
```diff
@@ -127,5 +130,25 @@ var _ = Describe("UserLogin", Label("datafederation"), func() {
...
+				Expect(err).To(BeNil())
+				Expect(resp).NotTo(BeNil())
+				Expect(resp.StatusCode).To(BeEquivalentTo(http.StatusNoContent))
+			}
```
helderjs, line 151: I'd do Ginkgo style

```suggestion
			Eventually(func(g Gomega) {
				_, resp, err := atlasClient.Client.DataFederationApi.
					DeleteDataFederationPrivateEndpoint(testData.Context, testData.Project.ID(), pe.ID).
					Execute()

				g.Expect(err).To(BeNil(), fmt.Sprintf("deletion of private endpoint failed with error %s", err))
				g.Expect(resp).NotTo(BeNil())
				g.Expect(resp.StatusCode).To(BeEquivalentTo(http.StatusNoContent))
			}).WithTimeout(5 * time.Minute).WithPolling(15 * time.Second).Should(Succeed())
```

helderjs (approved):
lgtm but I have a suggestion

test/e2e/datafederation_pe_test.go
josvazg: This would not work.

I can actually use an `Eventually`, as you and @igor-karpukhin suggested, but I need to expect at the return to be successful twice. Once is not enough due to the server side issue.

I will rewrite, but it will be a bit different than what is suggested here.

helderjs (approved):

PR #1884, opened 2024-10-23
CLOUDP-280230: Network peering translation layer

josvazg:
This comes in 3 separate commits for review that can be checked separately:

1. The translation layer code itself.
2. Improvements in the contract testing helper required by this network peering translation layer.
3. The network peering translation layer contract tests.

There is also a previous commit added to remove left over extra comment in the `go.mod` on a prior change.

### All Submissions:


internal/translation/networkpeering/conversion.go
```diff
@@ -0,0 +1,125 @@
...
+		return &admin.BaseNetworkPeeringConnectionSettings{
+			ContainerId:         peer.ContainerID,
+			ProviderName:        pointer.SetOrNil(string(peer.ProviderName), ""),
+			AzureDirectoryId:    pointer.SetOrNil(peer.AzureDirectoryID, ""),
+			AzureSubscriptionId: pointer.SetOrNil(peer.AzureSubscriptionID, ""),
+			ResourceGroupName:   pointer.SetOrNil(peer.ResourceGroupName, ""),
+			VnetName:            pointer.SetOrNil(peer.VNetName, ""),
+		}
+	default:
+		panic(fmt.Errorf("unsupported provider %q", peer.ProviderName))
```
s-urbaniak, line 49: i am sceptical this should panic 🤔 are we really considering this being a programmatic error?

josvazg: I am on the fence as well. There are 2 ways to hit this:

1. Our validation fails to filter the provider name in the CRD.
2. Atlas sends an incorrect value.

For the first one we could panic, is our fault. But for the second, I guess we should error out instead.

s-urbaniak: @josvazg given this is driven by user input I am in strong favor of returning an error.

A panic is valid if for we wire things incorrectly internally, i.e. use the "wrong indexer name" or something like that. But anything that arrives "via wire" from the user to AKO shouldn't panic unless there are very very strong reasons.

s-urbaniak: > Atlas sends an incorrect value.

same rule as above. This is Atlas output, not an internal wiring error, it shouldn't panic.

josvazg: I will introduce an error return value and change all the code accordingly.

josvazg:
CLOUDP-280230

internal/translation/networkpeering/conversion.go
josvazg: Done, we use errors both ways now.

josvazg:
The contract tests have issues in the CI that do not arise locally that often. Need to debug.
Also need to add GCP and Azure coverage.

test/contract/audit/audit_test.go
```diff
@@ -86,12 +87,13 @@ func TestSyncs(t *testing.T) {
...
 	}
-	contract.RunContractTest(t, "test syncs", func(ct *contract.ContractTest) {
+	contract.RunGoContractTest(ctx, t, "test syncs", func(ch contract.ContractHelper) {
 		projectName := "audit-syncs-project"
-		ct.AddResources(time.Minute, contract.DefaultAtlasProject(projectName))
-		testProjectID := mustReadProjectID(t, ct.Ctx, ct.K8sClient, ct.Namespace(), projectName)
+		require.NoError(t, ch.AddResources(ctx, time.Minute, contract.DefaultAtlasProject(projectName)))
+		testProjectID, err := ch.ProjectID(ctx, projectName)
+		require.NoError(t, err)
 		ctx := context.Background()
```
helderjs, line 95: Not needed once there's already one on line 33

test/contract/networkpeering/aws.go
helderjs, line 1: It would be good to merge/move it to some common package and reuse it here and for e2e (same applies for azure and GCP helpers)

test/contract/networkpeering/networkpeering_test.go
```diff
@@ -0,0 +1,253 @@
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/test/helper/contract"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/test/helper/e2e/utils"
+)
+
+const (
+	testVPCName = "ako-test-network-peering-vpc"
+)
+
+func TestPeerContainerServiceCRUD(t *testing.T) {
+	ctx := context.Background()
```
helderjs, line 26: Suggestion for improvement: The "framework" should be generating and making context available for test function, it might grow complex and need more control of it.

helderjs (changes requested):
Looks good overall. Just some comments (mostly suggestions)

pkg/api/v1/atlasnetworkpeering_types.go
```diff
@@ -0,0 +1,75 @@
...
+limitations under the License.
+*/
+
+package v1
+
+import "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1/provider"
+
+// +k8s:deepcopy-gen=false
+
+type AtlasNetworkPeeringConfig struct {
```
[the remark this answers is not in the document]
josvazg: Can you elaborate?

I mean, I assume that doing the translation layer before the reconcile logic I might need to fix this a bit in the next PR. But I would not like to miss anything big here, or write code that will not be needed.

This here is the type that will later be included inline within the spec of the standalone network peering CRD. I believe I copied all relevant fields as is from the Atlas Project type.

Did I miss something here?

josvazg (commented):
replied, will perform the agreed changes next

internal/translation/networkpeering/conversion.go
```diff
@@ -0,0 +1,158 @@
...
+	akov2.AtlasNetworkPeeringConfig
+	ID string
+}
+
+type ProviderContainer struct {
+	ID             string
+	ProviderName   provider.ProviderName
+	AtlasCIDRBlock string
+	RegionName     string // AWS
+	Region         string // Azure
```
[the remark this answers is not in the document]
josvazg: Done, coming in next update

internal/translation/networkpeering/conversion_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if peer.ProviderName != provider.ProviderGCP {
+		peer.GCPConfiguration = nil
+	}
+	if peer.ProviderName != provider.ProviderAzure {
+		peer.AzureConfiguration = nil
+	}
+}
+
+func makeUInt(i int) uint {
```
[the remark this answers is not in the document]
josvazg: I managed to remove it. It did not fit well to create a package for this and there was not good place anywhere.
Coming on next update.

test/contract/audit/audit_test.go
josvazg: Done

test/contract/networkpeering/aws.go
josvazg: coming in next update

internal/translation/networkpeering/conversion.go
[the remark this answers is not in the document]
josvazg: GCP regions is pending, not sure we support it ourselves, do we?

pkg/api/v1/atlasnetworkpeering_types.go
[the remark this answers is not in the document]
josvazg: So I am missing ContainerRegion and AtlasCIDRBlock. Will fix.

internal/translation/networkpeering/conversion.go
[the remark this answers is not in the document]
josvazg: About GCP regions I will create a Jira and discuss as it would be a feature extension.

pkg/api/v1/atlasnetworkpeering_types.go
```diff
@@ -0,0 +1,86 @@
...
+	//ProviderName is the name of the provider. If not set, it will be set to "AWS".
+	// +optional
+	ProviderName provider.ProviderName `json:"providerName,omitempty"`
```
helderjs, line 26: I'm wondering about the value of using this type. I decided to not use it in private endpoint CRD

helderjs (approved):
LGTM. Just raised a question.

pkg/api/v1/atlasnetworkpeering_types.go
josvazg: Can you elaborate?
I mean, this and the other config will go embedded in the spec and are the part of the API that directly map to Atlas concepts and NOT Kubernetes related fields.
What did you drop on private endpoints exactly?

helderjs: The `provider.ProviderName` as the type for provider value(AWS/AZURE/GCP)

josvazg: ok, so you just mean make the provided name here a plain string?

internal/translation/networkpeering/networkpeering.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/pointer"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1/provider"
+)
+
+type PeerConnectionsService interface {
+	CreatePeer(ctx context.Context, projectID string, conn *NetworkPeer) (*NetworkPeer, error)
+	ListPeers(ctx context.Context, projectID string) ([]NetworkPeer, error)
```
s-urbaniak, line 15: curiosity question: why not pointers to save memory?

internal/translation/networkpeering/networkpeering.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil, fmt.Errorf("failed to list network peers for %s: %w", string(providerName), err)
+		}
+		peersList = append(peersList, peers...)
+	}
+	return peersList, nil
+}
+
+func (np *networkPeeringService) listPeersForProvider(ctx context.Context, projectID string, providerName provider.ProviderName) ([]NetworkPeer, error) {
+	results := []NetworkPeer{}
+	pageNum := 1
```
s-urbaniak, line 70: it's the first time I see paging being used. Why do we need to do this here and not elsewhere?

pkg/api/v1/atlasnetworkpeering_types.go
```diff
@@ -0,0 +1,86 @@
+/*
+Copyright 2020 MongoDB.
```
s-urbaniak, line 2: nit: bump to 2024 and elsewhere in the contribution.

pkg/api/v1/atlasnetworkpeering_types.go
s-urbaniak: Most importantly. Is `ProviderName` already used in the existing subobject structs?

pkg/api/v1/atlasnetworkpeering_types.go
```diff
@@ -0,0 +1,86 @@
...
+	ContainerRegion string `json:"containerRegion"`
+	//Atlas CIDR. It needs to be set if ContainerID is not set.
+	// +optional
+	AtlasCIDRBlock string `json:"atlasCidrBlock"`
+}
+
+// +k8s:deepcopy-gen=false
+
+type AWSNetworkPeeringConfiguration struct {
+	//AccepterRegionName is the provider region name of user's vpc.
```
s-urbaniak, line 56: Please add `+optional` if they are truly optional everywhere. See https://ahmet.im/blog/crd-generation-pitfalls/#explicit-required-or-optional-on-every-field

pkg/api/v1/atlasnetworkpeering_types.go
```diff
@@ -0,0 +1,86 @@
...
+type AtlasProviderContainerConfig struct {
+	// ContainerRegion is the provider region name of Atlas network peer container. If not set, AccepterRegionName is used.
+	// +optional
+	ContainerRegion string `json:"containerRegion"`
+	//Atlas CIDR. It needs to be set if ContainerID is not set.
+	// +optional
+	AtlasCIDRBlock string `json:"atlasCidrBlock"`
+}
+
+// +k8s:deepcopy-gen=false
```
s-urbaniak, line 53: why is this omitting deepgen?

s-urbaniak (changes requested):
some concerns and questions

internal/translation/networkpeering/conversion_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+var providerNames = []provider.ProviderName{
+	provider.ProviderAWS,
+	provider.ProviderAzure,
+	provider.ProviderGCP,
+}
+
+func FuzzConvertConnection(f *testing.F) {
+	for i := uint(0); i < fuzzIterations; i++ {
+		f.Add(([]byte)(fmt.Sprintf("seed sample %x", i)), i)
```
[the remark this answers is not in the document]
josvazg: Note this is NOT the CRD yet, this is only the structs needed by the translation layer that will be part of the CRD's spec in the next PR. This is ALL private code still, nothing public facing.

internal/translation/networkpeering/networkpeering.go
josvazg: `[]*something` makes sense when items might be nil, which I do not think is the case here.

A list coming from Atlas will have a definite length and no empty items in the list. In fact this is more memory compact, you have a slice which points to an array of contiguous structs. Otherwise you have a slice pointing to an array of contiguous pointers that will point to individually allocated structs. Right?

internal/translation/networkpeering/networkpeering.go
```diff
@@ -0,0 +1,147 @@
...
+type PeerConnectionsService interface {
+	CreatePeer(ctx context.Context, projectID string, conn *NetworkPeer) (*NetworkPeer, error)
+	ListPeers(ctx context.Context, projectID string) ([]NetworkPeer, error)
+	DeletePeer(ctx context.Context, projectID, containerID string) error
+}
+
+type PeeringContainerService interface {
+	CreateContainer(ctx context.Context, projectID string, container *ProviderContainer) (*ProviderContainer, error)
+	GetContainer(ctx context.Context, projectID, containerID string) (*ProviderContainer, error)
+	ListContainers(ctx context.Context, projectID, providerName string) ([]ProviderContainer, error)
```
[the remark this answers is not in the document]
josvazg: Apart from the previous explanation, not having a pointer avoids extra checks for nils whiuch are not required here.

internal/translation/networkpeering/networkpeering.go
josvazg: I think elsewhere it was a bug, because usually it will not be needed, but we have no guarantees some config might not pass one page.

Still this might not be needed now that peers are individually managed instead of being part of a list. I will need to revisit it.

pkg/api/v1/atlasnetworkpeering_types.go
```diff
@@ -0,0 +1,86 @@
...
+*/
+
+package v1
+
+import "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1/provider"
+
+// +k8s:deepcopy-gen=false
+
+type AtlasNetworkPeeringConfig struct {
+	//ProviderName is the name of the provider. If not set, it will be set to "AWS".
```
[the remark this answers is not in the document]
josvazg: Seems we need a default there, thanks!

pkg/api/v1/atlasnetworkpeering_types.go
josvazg: I can revisit usages, but if we must be consistent we should be across also private endpoints right?

pkg/api/v1/atlasnetworkpeering_types.go
josvazg: not a CRD yet, explained elswhere

s-urbaniak: deepcopy is important (maybe not here but elsewhere) in the translation. I _think_ it is possible to generate deepcopy without having it CRD generated. Try not using the Scheme registration, maybe that is enough.

s-urbaniak: and in fact, deepcopy is valuable for translation layer types as well.

internal/translation/networkpeering/networkpeering.go
s-urbaniak: Ohhh! In that case please open a bug ticket to investigate List operations everywhere! We use lists in a couple of translation layer services already without paging.

s-urbaniak: Please investigate if a List without paging yields **all** results. If so, paging is not necessary here.

internal/translation/networkpeering/conversion_test.go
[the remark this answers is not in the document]
s-urbaniak: Is this comment for a different question? :-D

pkg/api/v1/atlasnetworkpeering_types.go
josvazg: will try to see if we can use deep copy while not making a CRD yet.

internal/translation/networkpeering/conversion_test.go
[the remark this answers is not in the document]
josvazg: https://go.dev/doc/security/fuzz/
the idea is that when you use go fuzz you have 2 modes of operation:
- Regular unit tests, in which each add call is like setting a regular static test case.
- Fuzzing round with random data, which need to be invoked differently and per fuzzing test.

Unfortunately the second mode of operation means you need to make changes to the CI pipeline to use.

Which begs the question, do we go for the go idiomatic mode and change our CI?
Or should we compromise?

The good thing of the go proposed way is that unit tests are still deterministic and fast. The bad is that fuzzing is explicit and separate, which mean you need to change your CI.

My proposal:

1. Have some deterministic entries.
2. Use some random rounds as well, at least until we do not make fuzzing explicit in the CI.
3. When pathological inputs are found, make them part of the deterministic examples.

WDYT?

internal/translation/networkpeering/networkpeering.go
josvazg: I got this AI response from AI, these is no definite idiomacy here:

> In Go, when dealing with large structs, it is indeed more idiomatic and efficient to use pointers (e.g., *BigStruct) when passing them to functions or methods. This is because passing a pointer avoids copying the entire struct, which can be costly in terms of performance and memory usage.
> 
> When it comes to slices of large structs, the choice between []*BigStruct and []BigStruct depends on the specific use case and trade-offs:
> 
> []*BigStruct:
> 
> Pros: Passing around slices of pointers can be more efficient if the structs are large, as only the pointers are copied, not the entire structs.
> Cons: This approach can lead to more complex memory management, as you need to be careful about the lifetime and ownership of the pointed-to structs. It can also lead to more cache misses, as the data is not contiguous in memory.
> []BigStruct:
> 
> Pros: The data is stored contiguously in memory, which can lead to better cache performance. This can be beneficial for read-heavy operations.
> Cons: Copying slices of large structs can be expensive in terms of memory and performance.
> In general, if you need to modify the structs in the slice or if the structs are very large, using []*BigStruct might be more appropriate. If the structs are relatively small or if you are primarily reading from them, []BigStruct might be more efficient.
> 
> As for sources, the Go community often discusses these patterns in forums, blog posts, and the official Go documentation. While there might not be a single authoritative source that explicitly states this preference, you can find discussions on this topic in places like the Go Blog, Go's official documentation, and community forums like the Go subreddit or Stack Overflow. The Go Proverbs, a collection of idiomatic guidelines, also provide insights into Go's design philosophy, although they don't specifically address this question.

In our particular case the lists are read only and relatively small, so []Type makes more sense I think. Whereas []*Type is best for when the list itself grows large, mutates often and/or can have nils.

pkg/api/v1/atlasnetworkpeering_types.go
josvazg: removed and all is fine, I think I rejected the changes to deep copy autogen, but you are right they are already useful even before the CRD

s-urbaniak (approved):

pkg/api/v1/atlasnetworkpeering_types.go
s-urbaniak: did you push all commits? I still see it here.

pkg/api/v1/atlasnetworkpeering_types.go
josvazg: I think is not optional

pkg/api/v1/atlasnetworkpeering_types.go
josvazg: not yet, but I already tested locally

internal/translation/networkpeering/networkpeering.go
josvazg: CLOUDP-285975

PR #1902, opened 2024-10-31
CLOUDP-280225: Refactor integration secret checks

josvazg:
Atlas redacts integrations secrets for security reasons, which means we cannot:
- Compare them to check whether they need to be re-applied or not.
- Detect whether or not applied secrets got properly set.

This change preserves that behaviour without assuming any extension to this behaviour coming any time soon. A proper fix will have to come from Atlas, for instance, sending a hash of the secret that allows comparison by applying the same on the client side.

### All Submissions:


helderjs (approved):

josvazg:
Will hold until @s-urbaniak can review this one

josvazg:
Alternatively we could just remove the secret values from comparisons, so we only take into account Kubernetes changes, and ignore changes from Atlas.

s-urbaniak (approved):

PR #1903, opened 2024-10-31
CLOUDP-281207: Tag main on release merge

josvazg:
With `Automatically delete head branches` ON we no longer accumulate merged branches, but the tag workflow broke, because the merged branch was deleted before the workflow could tag it.

This fix:

1. Tags the just merged `main` branch instead of the removed merged branch.
2. Allows to call tag manually, in which case the branch to tag for release is an input parameter defaulting to `main`.
3. The SBOM PR in triggered by the release workflow completion. The version is taken as an output from the release.

### All Submissions:


helderjs (approved):
LGTM, but can we test it before the next release?

josvazg:
> LGTM, but can we test it before the next release?

Yep, that is the tricky part. I will try to see how to test without actually releasing. Maybe a pre-release test.

s-urbaniak (approved):

PR #1912, opened 2024-11-06
CLOUDP-280230: Replace launcher with contract helper

josvazg:
Simplify contract testing:

- No `TestMain` preparations required.
- No need for external YAML files for infrastructure.
- No need for a AKO helm version, latest will be used by default.
- No explicit skipping and cleaning, it happens within the contract test run wrapper.
- No hard dependency of the helper in `kind`, assumes a working k8s config is available, which CI and make provide easily.

Note: Using `make` locally, the contract tests launch `kind` with them. On the `CI`, this is not needed because the workflow launched `kind`already before running the tests, so we use an existing env var in the launch script to avoid launching `kind` again.

### All Submissions:


test/helper/contract/k8s.go
```diff
@@ -0,0 +1,87 @@
...
+
+func k8sRecreate(ctx context.Context, k8sClient client.Client, obj client.Object) error {
+	err := k8sClient.Create(ctx, obj)
+	if err == nil {
+		return nil
+	}
+	if !k8sErrors.IsAlreadyExists(err) {
+		return fmt.Errorf("failed to create object: %w", err)
+	}
+	if err := k8sClient.Update(ctx, obj); err != nil {
```
s-urbaniak, line 83: this is missing conflict resolution.

s-urbaniak (changes requested):
one comment, else lgtm 👍

josvazg:
CLOUDP-280230

test/helper/contract/k8s.go
josvazg: Have added a retry on update.

s-urbaniak (approved):

PR #1913, opened 2024-11-06
CLOUDP-268810: Independent custom roles

igor-karpukhin:
### All Submissions:

This PR introduces a new CRD: `AtlasCustomRole` together with a dedicated controller that works exactly as `custom roles` for `AtlasProject` CRD


bundle.Dockerfile
```diff
@@ -1,17 +1,13 @@
...
 LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=mongodb-atlas-kubernetes
-LABEL operators.operatorframework.io.bundle.channels.v1=stable
-LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.34.1
+LABEL operators.operatorframework.io.bundle.channels.v1=beta
+LABEL operators.operatorframework.io.bundle.channel.default.v1=beta
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.35.0
```
s-urbaniak, line 10: this looks strange. What is it beta and 1.35?

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,105 @@
...
+		projectID: akoCustomRole.Spec.ProjectIDRef.ID,
+		deleted:   !akoCustomRole.DeletionTimestamp.IsZero(),
+		role:      akoCustomRole,
+	}
+
+	return r.Reconcile()
+}
+
+func (r *roleController) Reconcile() workflow.Result {
+	currentCustomRoles, err := r.service.List(r.ctx.Context, r.projectID)
```
s-urbaniak, line 38: question: we can only list custom roles in Atlas, not get a concrete one?

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,105 @@
...
+
+func handleCustomRole(ctx *workflow.Context, akoCustomRole *akov2.AtlasCustomRole) workflow.Result {
+	ctx.Log.Debug("starting custom role processing")
+	defer ctx.Log.Debug("finished custom role processing")
+
+	r := roleController{
+		ctx:       ctx,
+		service:   customroles.NewCustomRoles(ctx.SdkClient.CustomDatabaseRolesApi),
+		projectID: akoCustomRole.Spec.ProjectIDRef.ID,
+		deleted:   !akoCustomRole.DeletionTimestamp.IsZero(),
```
s-urbaniak, line 30: stylistic nit: to reduce the amount of fields, I suggest to just inject `akoCustomRole` and derive `projectID` and `deleted` inside the `roleController`. Technically, there is no need to dependency-inject those and makes the struct unnecessary bigger.

pkg/operator/builder.go
```diff
@@ -7,6 +7,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlascustomrole"
```
s-urbaniak, line 10: package ordering

s-urbaniak:
first iteration looks great already! 🎉

pkg/controller/atlascustomrole/custom_role.go
igor-karpukhin: We can call `Get` but this would require to modify the TranslationLayer for the `Get` call (which I think is used somewhere already), as it doesn't return the `httpResponse` which is required in this case to check if role exists.

bundle.Dockerfile
igor-karpukhin: Reverted back to stable

bundle.Dockerfile
```diff
@@ -1,17 +1,13 @@
 FROM scratch
 
-LABEL com.redhat.openshift.versions="v4.8"
-LABEL com.redhat.delivery.backport=true
-LABEL com.redhat.delivery.operator.bundle=true
```
josvazg, line 5: Not needed any more?

pkg/api/v1/atlascustomrole_types.go
```diff
@@ -0,0 +1,56 @@
...
+	Items           []AtlasCustomRole `json:"items"`
+}
+
+// AtlasCustomRoleSpec defines the desired state of CustomRole in Atlas
+type AtlasCustomRoleSpec struct {
+	api.LocalCredentialHolder `json:",inline"`
+	Role                      CustomRole `json:"role"`
+	// ID of the Atlas Project this role is attached to
+	// +required
+	ProjectIDRef ExternalProjectReference `json:"projectIDRef"`
```
helderjs, line 55: The field is called `externalProjectRef` in all other already migrated CRDs, let's keep consistency.

pkg/api/v1/atlascustomrole_types.go
```diff
@@ -0,0 +1,56 @@
...
+// AtlasCustomRoleList contains a list of AtlasCustomRole
+// +kubebuilder:object:root=true
+type AtlasCustomRoleList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []AtlasCustomRole `json:"items"`
+}
+
+// AtlasCustomRoleSpec defines the desired state of CustomRole in Atlas
+type AtlasCustomRoleSpec struct {
```
helderjs, line 50: Reference for kube project is missing

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -0,0 +1,195 @@
...
+	return workflow.OK().ReconcileResult()
+}
+
+// notFound terminates the reconciliation silently(no updates on conditions) and without retry
+func (r *AtlasСustomRoleReconciler) notFound(req ctrl.Request) ctrl.Result {
+	r.Log.Infof("Object %s doesn't exist, was it deleted after reconcile request?", req.NamespacedName)
+	return workflow.TerminateSilently().WithoutRetry().ReconcileResult()
+}
+
+func (r *AtlasСustomRoleReconciler) SetupWithManager(mgr ctrl.Manager, skipNameValidation bool) error {
```
josvazg, line 189: I believe this lacks indexers for when the credentials change (when using IDs). Eg:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go#L223-L227

pkg/controller/atlascustomrole/custom_role.go
helderjs: I saw no one is actually calling `GET()` from the sub-resource code. You can modify it to fit your case without problem. Also, I'd return nil and no error when not found, so you know the resource doesn't exist in atlas.

helderjs (changes requested):
Looks good so far, but despite the comments I'm missing some parts:

*Add an index by project and check in the project to not allow deletion when there are refences pending.
* Validate by test (e2e) migration path. Does sub-resource code conflicts with CRD?

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,106 @@
...
+
+	r := roleController{
+		ctx:       ctx,
+		service:   customroles.NewCustomRoles(ctx.SdkClient.CustomDatabaseRolesApi),
+		role:      akoCustomRole,
+		dpEnabled: dpEnabled,
+	}
+
+	return r.Reconcile()
+}
```
josvazg, line 34: I am not sure why we need sub-reconciler for the custom role.

This seems to me like the regular inner reconcile logic we always have, after usual login that gets the Kubernetes object from the plain request, skips if needed, gets the Atlas status, etc. Not sure why we have to separate it here.

But even if we separate, we do not need a custom `roleController` for something that is request specific. Do we?

This separation makes more sense when you have a list of items at the top level controller, and then at the next level here there is the item to be reconciled.

josvazg (changes requested):
I think we need, at the very least, to add support for using the custom roles with a reference to the project, not just always as independent CRDs.

pkg/api/v1/externalreference.go
```diff
@@ -3,5 +3,6 @@ package v1
 type ExternalProjectReference struct {
 	// ID is the Atlas project ID
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern:=`^([a-f0-9]{24})$`
```
s-urbaniak, line 6: You cannot retrospectively tighten validation. This potentially can break existing users.

s-urbaniak: we can do this but only once CRD ratcheting is 100% supported by the minimal Kubernetes version, cc @roothorp

pkg/api/v1/atlascustomrole_types.go
s-urbaniak, line 55: and how do we reference projects internally? As far as I see this is missing:
```
Project *common.ResourceRefNamespaced `json:"projectRef,omitempty"
```

pkg/api/v1/externalreference.go
igor-karpukhin: This is required by Atlas, so it has to be in this format already. But I can remove it, no problem

pkg/api/v1/atlascustomrole_types.go
igor-karpukhin: Yes, good point, I think I missed that

igor-karpukhin:
> Three blockers from my side (in addition to the comments from others):
> 
> * referencing AtlasProject resources inside a cluster doesn't seem to be possible
> * Usage of `cmp` in production code
> * Tightening validation without ratcheting (and ratcheting won't be available for some time)

I think the `cmp` usage has to be removed from there as well: https://github.com/mongodb/mongodb-atlas-kubernetes/blob/2921cceadd0117063ed6ee4cd5c23d940e5e7412/pkg/controller/atlasproject/custom_roles.go#L177

pkg/controller/atlascustomrole/atlascustomrole_controller.go
igor-karpukhin: Thanks, fixed

pkg/api/v1/externalreference.go
igor-karpukhin: Removed

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,106 @@
...
+func (r *roleController) create(role customroles.CustomRole) workflow.Result {
+	err := r.service.Create(r.ctx.Context, r.projectID, role)
+	if err != nil {
+		return r.terminate(workflow.AtlasCustomRoleNotCreated, err)
+	}
+	return r.idle()
+}
+
+func (r *roleController) update(roleInAKO, roleInAtlas customroles.CustomRole) workflow.Result {
+	if cmp.Diff(roleInAKO, roleInAtlas, cmpopts.EquateEmpty()) == "" {
```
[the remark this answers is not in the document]
igor-karpukhin: We still use it in custom roles. Also, the panics are mitigated by `cmpopts.IgnoreUnexported()`

bundle.Dockerfile
igor-karpukhin: Looks like these removed by the new controller-runtime version. Anyway, the bundle is built by the RedHat. Any labels from the `bundle.Dockerfile` are replaced by the values from the `annotations.yaml`

pkg/api/v1/atlascustomrole_types.go
igor-karpukhin: Fixed

pkg/controller/atlascustomrole/custom_role.go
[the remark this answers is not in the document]
s-urbaniak: This is not only about panics, as per go doc, see https://pkg.go.dev/github.com/google/go-cmp/cmp:
>  It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal.

I highly discourage us from using it in new production code. While `reflect.DeepEquals` is not perfect and has pitfalls, we want to implement custom semantic `Equals` statements in the translation layer if necessary, cc @josvazg .

s-urbaniak (approved):
redacting "Request Changes" to "Approve" to not be a blocker as I am on the other side of the world but please address my concerns.

pkg/controller/atlascustomrole/custom_role.go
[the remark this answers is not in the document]
josvazg: Yes, please, no more `cmp.Diff` calls on new or touched code.

internal/translation/customroles/conversion.go
```diff
@@ -70,32 +70,38 @@ func toAtlasInheritedRoles(role *CustomRole) []admin.DatabaseInheritedRole {
...
+	var actions []akov2.Action
+	if role.Actions != nil {
+		actions = make([]akov2.Action, 0, len(*role.Actions))
+		for _, atlasAction := range *role.Actions {
+			var resources []akov2.Resource
+			if atlasAction.Resources != nil {
+				resources = make([]akov2.Resource, 0, len(*atlasAction.Resources))
+				for _, atlasResource := range *atlasAction.Resources {
+					resources = append(resources, akov2.Resource{
+						Cluster:    pointer.MakePtr(atlasResource.Cluster),
```
josvazg, line 93: Too much cyclomatic complexity here.

instead we could:
- Exit early returning the custom role without actions OR delegate the actions population to a function. 
- And/or have a function for each resource build.

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,121 @@
...
+	err := r.service.Create(r.ctx.Context, r.projectID, role)
+	if err != nil {
+		return r.terminate(workflow.AtlasCustomRoleNotCreated, err)
+	}
+	return r.idle()
+}
+
+func (r *roleController) update(roleInAKO, roleInAtlas customroles.CustomRole) workflow.Result {
+	// cmpopts.IgnoreUnexported() prevents panic
+	if cmp.Diff(roleInAKO, roleInAtlas, cmpopts.EquateEmpty(), cmpopts.IgnoreUnexported()) == "" {
```
josvazg, line 93: Please, no more of these on new code, it must be replaced by a explicit comparison, usually aided by the translation layer. Usually implies some normalisation and `reflect.DeepEquals`

josvazg (commented):
Do we plan to follow up with e2e tests?

internal/translation/customroles/conversion.go
igor-karpukhin: Agreed. But this complexity was there before. I just saved it from nil pointer deference that I got during testing. We can refactor this later if needed.

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	err := r.Client.Get(ctx, req.NamespacedName, atlasCustomRole)
+	objectNotFound := err != nil && apiErrors.IsNotFound(err)
+	failedToRetrieve := err != nil && !objectNotFound
+
+	switch {
+	case failedToRetrieve:
+		return r.fail(req, err), nil
+	case objectNotFound:
+		return r.notFound(req), nil
+	}
```
josvazg, line 93: I believe this is new and I do not see it is only applicable to this particular resource type.

If this is the way we want to go now, we should be consistent across all controllers, which means we need to have a task revisiting and retrofitting this instead of simply fail/terminate when resolving the request into a Kubernetes object.

I would, at the very least create a Jira task so we do not forget.

igor-karpukhin: I took it from the https://github.com/mongodb/mongodb-atlas-kubernetes/blob/ad40d6f408409362bda7742087bf3681cc5640f7/pkg/controller/atlasdatabaseuser/atlasdatabaseuser_controller.go#L92-L97 that you refactored

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	workflowCtx.SetConditionTrue(api.ResourceVersionStatus).SetConditionTrue(api.ValidationSucceeded)
+
+	if !r.AtlasProvider.IsResourceSupported(atlasCustomRole) {
+		return r.terminate(workflowCtx, atlasCustomRole,
+			api.ProjectCustomRolesReadyType, workflow.AtlasGovUnsupported,
+			false,
+			fmt.Errorf("the %T is not supported by Atlas for government", atlasCustomRole)), nil
+	}
+
+	credentials, err := selectCredentials(workflowCtx.Context, r.Client, atlasCustomRole)
```
josvazg, line 131: We should also be consistent with how we solve the credentials and populate the project ID from external reference or Kubernetes. I believe that is s single operation which is always teh same for all CRDs and controllers that implement independent resource usage.

The current way this is solved can be seen in the database users controller:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/pkg/controller/atlasdatabaseuser/databaseuser.go#L48

And also in the deployment controller:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/pkg/controller/atlasdeployment/atlasdeployment_controller.go#L137

I am proposing a simplification that would allow use to just use the same code centrally with minimal rewrite on each new controller:
https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1930/files#diff-d17d85797ec5b7b965ec7b3508aab8f74d8bc04e930828939a234525efcaaddcR108

This is aided by a reconciler base implementation that does the heavy lifting for all reconcilers using composition:
https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1930/files#diff-1d07cdba1ef7647d2e1ac39e6f7a861bbb19e44824dafd0f069d0abe96856e4dR40

But while we decide whether we want to go that way or not, I would suggest we stick with the schema we already merged in version 2.5 for this.

igor-karpukhin: I find this one a bit misleading. The methods are called `getProject` but they do more than that. That's why I made an alternative

josvazg: To get the project ID you need the credentials. And you also need the credentials to populate the services. Maybe the naming is not the best, but we already had a way, and we should be consistent.

We can have competing alternatives, but we should converge to a single way in the end.

My proposal in draft solves consistency by having a single implementation all controllers who need it can just use. We can argue on naming and refactoring, but I would rather converge on a single way that having one way per controller.

BTW, in here there is no extraction of the project ID, is it? Can you elaborate on that?

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -0,0 +1,245 @@
...
+		project := &akov2.AtlasProject{}
+		err := k8sClient.Get(ctx,
+			client.ObjectKey{Name: akoRole.Spec.ProjectRef.Name, Namespace: akoRole.Spec.ProjectRef.Namespace}, project)
+		if err != nil {
+			return nil, errors.Wrap(err, "can not read credentials from AtlasProject")
+		}
+		if project.Spec.ConnectionSecret == nil || project.Spec.ConnectionSecret.Name == "" {
+			return nil, errors.Wrapf(err, "credentials for AtlasProject '%s' are not configured", project.GetName())
+		}
+		return &client.ObjectKey{Name: project.Spec.ConnectionSecret.Name, Namespace: project.Spec.ConnectionSecret.Namespace}, nil
```
josvazg, line 176: [customresource.ComputeSecret](https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/pkg/controller/customresource/customresource.go#L142) covers this, mostly.

In the alternative proposal this moves into the common reconciler here:
https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1930/files#diff-1d07cdba1ef7647d2e1ac39e6f7a861bbb19e44824dafd0f069d0abe96856e4dR139

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -0,0 +1,245 @@
...
+			fmt.Errorf("unable to create atlas client: %s", err.Error())), nil
+	}
+	workflowCtx.SdkClient = atlasSdkClient
+	if res := handleCustomRole(workflowCtx, r.Client, atlasCustomRole, r.ObjectDeletionProtection); !res.IsOk() {
+		return r.fail(req, fmt.Errorf("%s", res.GetMessage())), nil
+	}
+	return r.idle(workflowCtx), nil
+}
+
+func selectCredentials(ctx context.Context, k8sClient client.Client, akoRole *akov2.AtlasCustomRole) (*client.ObjectKey, error) {
```
josvazg, line 152: One thing I like about this proposal is that it does not require any Atlas calls yet, only Kubernetes queries.

Maybe we could:
1) Get credentials first.
2) Populate the project ID using the credentials as an input, as they are needed when using Kubernetes references, and optionally to check the project actually exists with external references.

But I suggest we make this either common methods or functions we can reuse from anywhere, instead of repeating them on every new controller that uses the same references and credentials schema.

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		if project.Spec.ConnectionSecret == nil || project.Spec.ConnectionSecret.Name == "" {
+			return nil, errors.Wrapf(err, "credentials for AtlasProject '%s' are not configured", project.GetName())
+		}
+		return &client.ObjectKey{Name: project.Spec.ConnectionSecret.Name, Namespace: project.Spec.ConnectionSecret.Namespace}, nil
+	}
+	return nil, errors.New("either 'externalProjectIDRef' or 'projectRef' must be set for the AtlasCustomRole resource")
+}
+
+func (r *AtlasCustomRoleReconciler) terminate(
```
josvazg, line 181: Not a blocker for this, but I would shortly be proposing to move all these repetitive methods to common code we just use and stop repeating over an over on each controller.

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -0,0 +1,245 @@
+package atlascustomrole
```
josvazg, line 1: Test coverage on the controller package is very good at a glance! good job!

pkg/controller/atlascustomrole/atlascustomrole_controller_test.go
```diff
@@ -0,0 +1,556 @@
+package atlascustomrole
```
josvazg, line 1: Test coverage on the controller package is very good at a glance! good job!

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type roleController struct {
+	ctx       *workflow.Context
+	projectID string
+	service   customroles.CustomRoleService
+	role      *akov2.AtlasCustomRole
+	dpEnabled bool
+	k8sClient client.Client
+}
```
josvazg, line 23: This, more than a controller, is a reconciliation request object, becase it is only tied to a single request context.

Calling it a controller makes it also confusing is this a sub-controller/sub-reconciler or the custom roles reconciler?

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,119 @@
...
+		service:   customroles.NewCustomRoles(ctx.SdkClient.CustomDatabaseRolesApi),
+		role:      akoCustomRole,
+		dpEnabled: dpEnabled,
+		k8sClient: k8sClient,
+	}
+
+	return r.Reconcile()
+}
+
+func (r *roleController) Reconcile() workflow.Result {
```
josvazg, line 41: this could be instead something like:

```go
func (r *AtlasCustomRoleReconciler) handle(customRoleRequest *CustomRoleRequest) {...
 ```
 
 And the request has anything that you need that is NOT already in the top level reconciler:
 ```go
type CustomRoleRequest struct {
	ctx       *workflow.Context
	projectID string
	role      *akov2.AtlasCustomRole
}
// service   customroles.CustomRoleService is already in the top level reconciler
// dpEnabled is already in the top level reconciler
// k8sClient client.Client is already in top level the reconciler

Which renders a request with just 3 elements, that could be just a call with 3 arguments:
[func (r *Resource) handle(ctx, projectID, Resource)](https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1930/files#diff-d17d85797ec5b7b965ec7b3508aab8f74d8bc04e930828939a234525efcaaddcR125)
```

josvazg: On the other hand, a request might make sense after all. The service(s), which might be more than one, cannot be solved until reconciliation starts, as they depend on credentials that depend on Kubernetes configuration that needs to be parse while doing the reconciliation.

So being strict, the services part is part of the request, and should not be set in the reconciler struct... which means we might have to fix this retroactively in several places.

Going back here, it seems the handle method should be invoked like this:
```go

r.handle(&ResourceHandleRequest{
  Context: ctx
  ProjectID: projectID:
  Resource: resource
  ServiceA: serviceA,...)
```

Note I am using generic naming as I think this can and should be generalized.

pkg/controller/atlascustomrole/atlascustomrole_controller.go
josvazg: CLOUDP-284790

pkg/controller/atlascustomrole/custom_role.go
josvazg: CLOUDP-284792 about the services being a request context bound concept, not a reconciler field

pkg/controller/atlascustomrole/atlascustomrole_controller.go
josvazg: I created CLOUDP-284794 so we discuss design beyond this particular PR

pkg/controller/atlascustomrole/atlascustomrole_controller.go
helderjs: I think `terminate()` is one of the transitions that can't be "generic". Some resource needs to update their status when terminating, others update more than one condition. We might be able to come up with some generic signature but before hand, I'd say is not straightforward.

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	roleFoundInAtlas := false
+	roleInAtlas, err := r.service.Get(r.ctx.Context, r.projectID, r.role.Spec.Role.Name)
+	switch {
+	case err != nil:
+		return workflow.Terminate(workflow.ProjectCustomRolesReady, err.Error())
+	case err == nil && (roleInAtlas != customroles.CustomRole{}):
+		roleFoundInAtlas = true
+	}
```
helderjs, line 64: I think a simple if here makes things more readable.

```suggestion
	roleInAtlas, err := r.service.Get(r.ctx.Context, r.projectID, r.role.Spec.Role.Name)
	if err != nil {
		return workflow.Terminate(workflow.ProjectCustomRolesReady, err.Error())
	}
	
	roleFoundInAtlas :=  roleInAtlas != customroles.CustomRole{}
```

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,119 @@
...
+	return r.idle()
+}
+
+func (r *roleController) delete(roleInAtlas customroles.CustomRole) workflow.Result {
+	err := r.service.Delete(r.ctx.Context, r.projectID, roleInAtlas.Name)
+	if err != nil {
+		return r.terminate(workflow.AtlasCustomRoleNotDeleted, err)
+	}
+	return r.idle()
```
helderjs, line 106: Does the resource become idle when deleted? The operator should stop managing the resource,  remove finalizers, and let the resource be removed.

helderjs (changes requested):
Almost there. Beyond the 2 comments I left, I'm missing 2 things.

1. Finalizer handling, I haven't see it in the code
2. A project being referenced by a custom role can still be deleted, letting the resource orphan and impossible to reconcile.

pkg/controller/atlascustomrole/atlascustomrole_controller.go
josvazg, line 176: This particular bit here is wrong, because it will prevent users from falling back to global credentials, which is the default we need to mainaint. Instead I suggest:
```go
		return project.ConnectionSecretObjectKey(), nil
```

`project.ConnectionSecretObjectKey()` areadu has the logic needed here:
```go
func (p *AtlasProject) ConnectionSecretObjectKey() *client.ObjectKey {
	if p.Spec.ConnectionSecret != nil {
		var key client.ObjectKey
		if p.Spec.ConnectionSecret.Namespace != "" {
			key = kube.ObjectKey(p.Spec.ConnectionSecret.Namespace, p.Spec.ConnectionSecret.Name)
		} else {
			key = kube.ObjectKey(p.Namespace, p.Spec.ConnectionSecret.Name)
		}
		return &key
	}
	return nil
}
```

pkg/controller/atlascustomrole/custom_role.go
igor-karpukhin: Thanks! Fixed

pkg/controller/atlascustomrole/custom_role.go
igor-karpukhin: Thanks, fixed!

pkg/indexer/atlascustomrolesprojects.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != nil {
+			a.logger.Errorf("unable to find project to index: %s", err)
+
+			return nil
+		}
+
+		if project.ID() != "" {
+			return []string{project.ID()}
+		}
+	}
```
helderjs, line 62: It should be only by project ref. External project ID doesn't make sense here, this is to avoid deleting the Project resource and leaving orphans dependent resources that will not be able to reconcile once the project is gone.

Note: A custom role with an external project ref will always have a local credentials set and, therefore, be able to reconcile. If the external project doesn't exist anymore, the resource should be allowed to be deleted. (good candidate for test case)

helderjs (changes requested):

pkg/indexer/atlascustomrolesprojects.go
igor-karpukhin: Good point. I removed this part. As well as I marked the similar part for the `atlasdabaseuser` indexer to be removed later

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (r *roleController) update(roleInAKO, roleInAtlas customroles.CustomRole) workflow.Result {
+	if reflect.DeepEqual(roleInAKO, roleInAtlas) {
+		return r.idle()
+	}
+	err := r.service.Update(r.ctx.Context, r.projectID, roleInAKO.Name, roleInAKO)
+	if err != nil {
+		return r.terminate(workflow.AtlasCustomRoleNotUpdated, err)
+	}
+	return r.idle()
```
helderjs, line 112: you need to call `managed()` here too for the case of sync with existing resource

helderjs (changes requested):

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -0,0 +1,245 @@
...
+type AtlasCustomRoleReconciler struct {
+	Client                      client.Client
+	Log                         *zap.SugaredLogger
+	Scheme                      *runtime.Scheme
+	EventRecorder               record.EventRecorder
+	AtlasProvider               atlas.Provider
+	GlobalPredicates            []predicate.Predicate
+	ObjectDeletionProtection    bool
+	SubObjectDeletionProtection bool
+	independentSyncPeriod       time.Duration
```
s-urbaniak, line 48: non-blocking question: why is this private and the other fields public?

pkg/controller/atlascustomrole/atlascustomrole_controller.go
josvazg: yes, I chose the wrong one probably, but most of the rest are pretty repetitive.

Anyway, closing for now.

pkg/controller/atlascustomrole/atlascustomrole_controller.go
igor-karpukhin: It is the same for `AtlasDatabaseUser` and the `AtlasDeployment` controllers. Others don't have this field at all

josvazg (approved):
LGTM with minor nits

pkg/indexer/atlascustomrolesprojects.go
```diff
@@ -0,0 +1,61 @@
...
+		err := a.client.Get(a.ctx, *role.Spec.ProjectRef.GetObject(role.Namespace), project)
+		if err != nil {
+			a.logger.Errorf("unable to find project to index: %s", err)
+
+			return nil
+		}
+
+		if project.ID() != "" {
+			return []string{project.ID()}
+		}
```
helderjs, line 57: @igor-karpukhin, just noticed when rebasing that you are still indexing by ID when it should be by namespace/name.
which also means you do not need context and client.

igor-karpukhin: Fixed by https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1939

PR #1916, opened 2024-11-08
CLOUDP-280222: Private Endpoint controller

helderjs:
### All Submissions:


pkg/controller/atlasprivateendpoint/atlasprivateendpoint_controller.go
```diff
@@ -0,0 +1,304 @@
...
+	akoPrivateEndpoint := akov2.AtlasPrivateEndpoint{}
+	result := customresource.PrepareResource(ctx, r.Client, req, &akoPrivateEndpoint, r.Log)
+	if !result.IsOk() {
+		return result.ReconcileResult(), errors.New(result.GetMessage())
+	}
+
+	return r.ensureCustomResource(ctx, &akoPrivateEndpoint)
+}
+
+func (r *AtlasPrivateEndpointReconciler) ensureCustomResource(ctx context.Context, akoPrivateEndpoint *akov2.AtlasPrivateEndpoint) (ctrl.Result, error) {
```
josvazg, line 83: Why the ensure... separation here? regular top level CRDs do not follow this pattern.

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+
+		atlasPEService, err = r.deletePrivateEndpoint(ctx.Context, projectID, atlasPEService)
+		if err != nil {
+			return r.terminate(ctx, akoPrivateEndpoint, atlasPEService, api.PrivateEndpointServiceReady, workflow.PrivateEndpointFailedToDelete, err)
+		}
+	}
+
+	switch atlasPEService.Status() {
+	case privateendpoint.StatusInitiating:
```
josvazg, line 67: I am confused here. Maybe we should document this state machine somewhere else.

The first switch handles synchronising the existence of the private endpoint service between Atlas and the Kubernetes spec. It is not very explicit when that exists early or when it falls through the next step, which is the switch that handles the atlas PE status, which should happen when an atlas PE existed or gets created.

Can we point to the state machine spec somewhere?

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,220 @@
...
+		return r.inProgress(ctx, akoPrivateEndpoint, atlasPEService, api.PrivateEndpointServiceReady, workflow.PrivateEndpointServiceDeleting, "Private Endpoint is being deleted")
+	}
+
+	ctx.SetConditionTrue(api.PrivateEndpointServiceReady)
+	r.EventRecorder.Event(akoPrivateEndpoint, "Normal", string(workflow.PrivateEndpointServiceCreated), "Private Endpoint Service is available")
+
+	return r.handlePrivateEndpointInterface(ctx, projectID, akoPrivateEndpoint, akoPEService, atlasPEService)
+}
+
+func (r *AtlasPrivateEndpointReconciler) handlePrivateEndpointInterface(
```
josvazg, line 83: The prefix **handle** would made me thing we have another state machine here, but this is more like a **sync**, missing endpoints get created and endpoints not in the spec get removed.

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	if !inProgress && reflect.DeepEqual(akoPEService.EndpointInterfaces(), atlasPEService.EndpointInterfaces()) {
+		return r.ready(ctx, akoPrivateEndpoint, atlasPEService)
+	}
+
+	return r.inProgress(ctx, akoPrivateEndpoint, atlasPEService, api.PrivateEndpointReady, workflow.PrivateEndpointUpdating, "Private Endpoints are being updated")
+}
+
+func (r *AtlasPrivateEndpointReconciler) getPrivateEndpointService(
```
josvazg, line 150: This is a bit too magic right?
It is searching by Service ID, if set, otherwise listing and matching by region. Is that second match alwasy accurate?

I would rename the method to be more explicit:

```suggestion
func (r *AtlasPrivateEndpointReconciler) getPrivateEndpointServiceByIDorRegion(
```

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	return nil, nil
+}
+
+func (r *AtlasPrivateEndpointReconciler) deletePrivateEndpoint(
+	ctx context.Context,
+	projectID string,
+	atlasPEService privateendpoint.EndpointService,
+) (privateendpoint.EndpointService, error) {
```
josvazg, line 177: why does a deletion return an object from a different type? It is confusing

pkg/controller/atlasproject/project.go
```diff
@@ -168,5 +168,21 @@ func (r *AtlasProjectReconciler) hasDependencies(ctx *workflow.Context, project
...
+			client.ObjectKeyFromObject(project).String(),
+		),
+	}
+	err = r.Client.List(ctx.Context, privateEndpoints, listOps)
+	if err != nil {
+		return false, err
+	}
+
+	return len(privateEndpoints.Items) > 0, nil
 }
```
josvazg, line 188: So... we were ignoring the private endpoints as dependencies before?

pkg/controller/workflow/reason.go
```diff
@@ -131,3 +131,16 @@ const (
 	StreamConnectionNotRemoved    ConditionReason = "StreamConnectionNotRemoved"
 	StreamConnectionNotUpdated    ConditionReason = "StreamConnectionNotUpdated"
 )
+
+const (
+	PrivateEndpointServiceCreated           ConditionReason = "PrivateEndpointServiceCreated"
```
josvazg, line 136:
```suggestion
)

// Atlas Private Endpoint reasons
const (
	PrivateEndpointServiceCreated           ConditionReason = "PrivateEndpointServiceCreated"
```

pkg/controller/atlasprivateendpoint/atlasprivateendpoint_controller.go
helderjs: This was mostly to separate scopes more clearly. It fits inside the caller with no concerns, I just think this way looks better. i.e. my taste.

pkg/controller/atlasprivateendpoint/privateendpoint.go
helderjs: Sorry, but I don't get where these assumptions come from. Am I missing some agreement on vocabulary?

pkg/controller/atlasprivateendpoint/privateendpoint.go
helderjs: the `Get` semantics for me is that it is supposed to return a single match or error the internal search is less relevant.
If I would be accepting an ID or region as parameters, then the renaming would make sense to me.

pkg/controller/atlasproject/project.go
helderjs: no, because it was a subresource, and when it is a sub-resource, it's done in a different place.

pkg/controller/atlasprivateendpoint/privateendpoint.go
josvazg: Maybe it is just me then

pkg/controller/atlasprivateendpoint/privateendpoint.go
josvazg: Well I think this method does too much, I would rather have a search by id then a filter by region explicitly on the caller. So that is more obvious what is going on there. The code in the caller is still small (one if and else clause) and at this level we have the find by id and list and filter by region as 2 smaller functions.

Basically having methods that do one thing as said in the method name.

But maybe that is just me.

pkg/controller/atlasprivateendpoint/privateendpoint.go
josvazg: I read it wrong, but I still do not understand why we need to pass the `atlasPEService privateendpoint.EndpointService` and return a `privateendpoint.EndpointService`, such service should be part of `r *AtlasPrivateEndpointReconciler` already no?
And even if it were not, why are we retuning it later?

I would expect this signature instead.

```go
func (r *AtlasPrivateEndpointReconciler) deletePrivateEndpoint(ctx context.Context, projectID string) error {
```

But I might be mossing something here.

helderjs: Got you now. The reason here is, one more time, there's a special case added by "yet another" Atlas strange behavior.
1. We have two parts that need to be deleted and both need to wait until finished (status `DELETING`)
2. For AWS, deleting the interface also deletes a service automatically, for the other providers we delete the interface and then delete the service.

Because of these reasons, we return the entity so we can watch its status until the action is completed.

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,220 @@
...
+	inProgress := false
+	for _, akoPEInterface := range akoPEService.EndpointInterfaces() {
+		atlasPEInterfaces := atlasPEService.EndpointInterfaces()
+		atlasPEInterface := atlasPEInterfaces.Get(akoPEInterface.InterfaceID())
+		existInAtlas := atlasPEInterface != nil
+		inProgress = isInterfaceInProgress(akoPEInterface.Status())
+		var err error
+
+		if !existInAtlas {
+			gcpProjectID := getGCPProjectID(akoPrivateEndpoint, akoPEInterface.InterfaceID())
```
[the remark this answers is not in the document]
helderjs: Here is because of Atlas (again). We need this to configure a GCP "interface" but Atlas doesn't give this field back to us.
We send it when creating, but when getting the resource we don't have this field, so we always send it to the translation layer even when it is empty.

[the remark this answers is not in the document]
josvazg: Ok, that is weird alright, but still the original question is, why is this code **not** guarded to run only when the provider is Google?

pkg/controller/atlasprivateendpoint/privateendpoint.go
josvazg: ok

pkg/controller/atlasprivateendpoint/privateendpoint.go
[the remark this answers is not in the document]
helderjs: Because there are no separate flows per provider, there's a creation flow and all the data should be there depending on the provider.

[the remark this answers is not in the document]
josvazg: Ok I see the guard inside the function:
```go
func getGCPProjectID(...) string {
	if akoPrivateEndpoint.Spec.Provider == privateendpoint.ProviderGCP && len(akoPrivateEndpoint.Spec.GCPConfiguration) > 0 {
...
	}

	return ""
}
```

pkg/controller/atlas/provider.go
```diff
@@ -76,7 +76,8 @@ func (p *ProductionProvider) IsResourceSupported(resource api.AtlasCustomResourc
 		*akov2.AtlasDatabaseUser,
 		*akov2.AtlasSearchIndexConfig,
 		*akov2.AtlasBackupCompliancePolicy,
-		*akov2.AtlasFederatedAuth:
+		*akov2.AtlasFederatedAuth,
+		*akov2.AtlasPrivateEndpoint:
```
josvazg, line 80: @igor-karpukhin we probably need to add custom roles in your case and network peering in mine here.
And also make sure to include some testing about gov support.

pkg/controller/atlasprivateendpoint/atlasprivateendpoint_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/customresource"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/statushandler"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/workflow"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/indexer"
+)
+
+// AtlasPrivateEndpointReconciler reconciles a AtlasPrivateEndpoint object
+type AtlasPrivateEndpointReconciler struct {
+	Scheme                 *runtime.Scheme
```
josvazg, line 54: We do not seem to be using this anywhere. I think we probably don't use it in most other reconcilers either.

pkg/controller/atlasprivateendpoint/atlasprivateendpoint_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	var atlasProject *project.Project
+	var err error
+	if akoPrivateEndpoint.Spec.ExternalProject != nil {
+		atlasProject, err = r.getProjectFromAtlas(ctx, akoPrivateEndpoint)
+	} else {
+		atlasProject, err = r.getProjectFromKube(ctx, akoPrivateEndpoint)
+	}
+	if err != nil {
+		return r.terminate(workflowCtx, akoPrivateEndpoint, nil, api.ReadyType, workflow.AtlasAPIAccessNotConfigured, err)
+	}
```
josvazg, line 112: Not a blocker for this PR, but I think the approach by @igor-karpukhin on getting the credentials first, then solving the reference is better than trying to do both at once as done here. I would argue we should eventually converge in that direction.

pkg/controller/atlasprivateendpoint/atlasprivateendpoint_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	akoPrivateEndpoint := akov2.AtlasPrivateEndpoint{}
+	result := customresource.PrepareResource(ctx, r.Client, req, &akoPrivateEndpoint, r.Log)
+	if !result.IsOk() {
+		return result.ReconcileResult(), errors.New(result.GetMessage())
+	}
+
+	return r.ensureCustomResource(ctx, &akoPrivateEndpoint)
+}
+
+func (r *AtlasPrivateEndpointReconciler) ensureCustomResource(ctx context.Context, akoPrivateEndpoint *akov2.AtlasPrivateEndpoint) (ctrl.Result, error) {
```
josvazg, line 85: I like this ensure separation, this regularizes the preamble in the following steps:
- **r.Reconcile** just unwraps the request into a full blown resource object.
- **r.ensure...** takes care of skipping, condition preparation, validation, checks for support and solves any parent reference. I wonder if we can make it generic.
- **r.handle...** evaluates the state machine.

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,236 @@
...
+	case existInAtlas && wasDeleted:
+		if customresource.IsResourcePolicyKeepOrDefault(akoPrivateEndpoint, r.ObjectDeletionProtection) {
+			return r.unmanage(ctx, akoPrivateEndpoint)
+		}
+
+		atlasPEService, err = r.deletePrivateEndpoint(ctx.Context, projectID, atlasPEService)
+		if err != nil {
+			return r.terminate(ctx, akoPrivateEndpoint, atlasPEService, api.PrivateEndpointServiceReady, workflow.PrivateEndpointFailedToDelete, err)
+		}
+	}
```
josvazg, line 63: Nit

Why not?

```suggestion
	case !existInAtlas && !wasDeleted:
		return r.create(ctx.Context, projectID, akoPEService)
	case !existInAtlas && wasDeleted:
		return r.unmanage(ctx, akoPrivateEndpoint)
	case existInAtlas && wasDeleted:
		return r.delete(ctx, akoPrivateEndpoint)
	default:
	       return r.terminate(ctx...)
	}
```

whether or not the action has a follow up is part of the action taken. That way this level of the state machine is simpler, more readable and also more regular.

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,236 @@
...
+	case privateendpoint.StatusFailed, privateendpoint.StatusRejected:
+		return r.terminate(ctx, akoPrivateEndpoint, atlasPEService, api.PrivateEndpointServiceReady, workflow.PrivateEndpointServiceFailedToConfigure, errors.New(atlasPEService.ErrorMessage()))
+	case privateendpoint.StatusDeleting:
+		return r.inProgress(ctx, akoPrivateEndpoint, atlasPEService, api.PrivateEndpointServiceReady, workflow.PrivateEndpointServiceDeleting, "Private Endpoint is being deleted")
+	}
+
+	ctx.SetConditionTrue(api.PrivateEndpointServiceReady)
+	r.EventRecorder.Event(akoPrivateEndpoint, "Normal", string(workflow.PrivateEndpointServiceCreated), "Private Endpoint Service is available")
+
+	return r.handlePrivateEndpointInterface(ctx, projectID, akoPrivateEndpoint, akoPEService, atlasPEService)
```
josvazg, line 79: This to me would much the updating status of the state machine.

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,236 @@
...
+		return r.waitForConfiguration(ctx, akoPrivateEndpoint, atlasPEService)
+	}
+
+	// we want to sync all interface, if any of them is in progress, after all we transition to in progress
+	pendingResources := false
+	for _, compositeInterfaceMap := range compositeInterfacesMap {
+		inAKO := compositeInterfaceMap.AKO != nil
+		inAtlas := compositeInterfaceMap.Atlas != nil
+		inProgress := isInterfaceInProgress(compositeInterfaceMap.Atlas)
+		failed := hasInterfaceFailed(compositeInterfaceMap.Atlas)
```
josvazg, line 101: We have 4 flags, which could render 16 combinations, but only 6 transitions in the state machine. It would be good to point here to the state machine diagram or summarise it.

pkg/controller/atlasprivateendpoint/privateendpoint.go
helderjs: I didn't get what you mean here.

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,236 @@
...
+				api.PrivateEndpointReady,
+				workflow.PrivateEndpointFailedToConfigure,
+				errors.New(compositeInterfaceMap.Atlas.ErrorMessage()),
+			)
+		case inProgress:
+			pendingResources = true
+			continue
+		case inAKO && !inAtlas:
+			gcpProjectID := getGCPProjectID(akoPrivateEndpoint, compositeInterfaceMap.AKO.InterfaceID())
+			_, err := r.privateEndpointService.CreatePrivateEndpointInterface(
```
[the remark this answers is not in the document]
helderjs: Sorry, but I also didn't get what you mean here. Could you improve the example?

pkg/controller/atlasprivateendpoint/privateendpoint.go
helderjs: The number of flags is not directly proportional to the number of transitions, once some applies only on a specific state.
Anyway, I'm gonna to add a doc block here explaining it

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return nil, nil
+}
+
+func (r *AtlasPrivateEndpointReconciler) deletePrivateEndpoint(
+	ctx context.Context,
+	projectID string,
+	atlasPEService privateendpoint.EndpointService,
+) (privateendpoint.EndpointService, error) {
+	if len(atlasPEService.EndpointInterfaces()) == 0 && atlasPEService.Status() != privateendpoint.StatusDeleting {
```
josvazg, line 182: Why would it be already deleting when we try to delete?

pkg/controller/atlasprivateendpoint/privateendpoint.go
```diff
@@ -0,0 +1,236 @@
...
+		err := r.privateEndpointService.DeleteEndpointService(ctx, projectID, atlasPEService.Provider(), atlasPEService.ServiceID())
+		if err != nil {
+			return atlasPEService, err
+		}
+	}
+
+	for _, i := range atlasPEService.EndpointInterfaces() {
```
josvazg, line 189: Why is this not before the service deletion? don't we need the endpoints to be deleted first?

josvazg (commented):
LGTM with minor nits, but will wait for CI to go green

pkg/controller/atlasprivateendpoint/privateendpoint.go
josvazg: This code seems to be for checking in progress status only, but is right after the state machine switch. I would see it more readable if the state machine was terminal on this method and the in progress check is an explicit transition or a follow up of create or delete transitions. That way each method deals with a different level of abstraction.

But maybe that is just me.

pkg/controller/atlasprivateendpoint/privateendpoint.go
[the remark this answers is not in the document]
josvazg: You could call a createPEI method (name is just an example) which is already multi-provider aware so we do not leak the GCP specific query at this method, but hide it within that method. I think it would make more sense that way, for readability as that call works at a lower abstraction level than this logic here.

pkg/controller/atlasprivateendpoint/privateendpoint.go
helderjs: It's not an atomic action operation and as we might be deleting one interface among other that are also configured, we watch for the status and skip items while they are being deleted in Atlas(avoid API errors).

pkg/controller/atlasprivateendpoint/privateendpoint.go
helderjs: As there is two pieces moving here and different behaviors depending on the configuration scenario and provider.
This was the best way to handle all case while keep implementation complexity low.
For example, here we ensure we delete a service only when there are not more interfaces pending, moving as you suggest it would require extra conditions in both part to interrupt (early return) the flow.

pkg/indexer/atlasprivateendpoints.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (a *AtlasPrivateEndpointByProjectIndexer) Keys(object client.Object) []string {
+	pe, ok := object.(*akov2.AtlasPrivateEndpoint)
+	if !ok {
+		a.logger.Errorf("expected *akov2.AtlasPrivateEndpoint but got %T", object)
+		return nil
+	}
+
+	if pe.Spec.Project == nil {
```
s-urbaniak, line 39:
```suggestion
	if pe.Spec.Project == nil || pe.Spec.Project.Name == "" {
```

pkg/indexer/atlasprivateendpoints.go
```diff
@@ -0,0 +1,50 @@
...
+		return nil
+	}
+
+	if pe.Spec.Project.Name == "" {
+		return nil
+	}
+
+	key := pe.Spec.Project.GetObject(pe.GetNamespace())
+
+	return []string{key.String()}
```
s-urbaniak, line 49:
```suggestion
	return []string{pe.Spec.Project.GetObject(pe.GetNamespace()).String()}
```

s-urbaniak (approved):
two nits regarding error handling (please fix this as part of THIS PR), else lgtm

internal/collection/collection.go
```diff
@@ -21,3 +21,11 @@ func Keys[K comparable, V any](m map[K]V) []K {
 
 	return s
 }
+
+func FirstKeyFromMap[K comparable, V any](m map[K]V) K {
```
s-urbaniak, line 25: @helderjs is this used anywhere? Besides, this is dangerous as the iteration order of keys inside of a map is not guaranteed. There is semantically not such a thing like "first key from map" in Golang.

internal/translation/privateendpoint/conversion.go
```diff
@@ -473,3 +473,30 @@ func interfaceCreateToAtlas(peInterface EndpointInterface, gcpProjectID string)
 
 	return nil
 }
+
+type CompositeEndpointInterface struct {
```
s-urbaniak, line 477: This name is too general, the composited types are concrete while the name is abstract. Just call the type very concrete, i.e. "Endpoints", see https://github.com/mongodb/mongodb-atlas-kubernetes/blob/46a504c6fa326037607bcc51e1518c3ede092330/internal/translation/datafederation/conversion_endpoints.go#L69-L71

pkg/controller/atlasprivateendpoint/atlasprivateendpoint_controller.go
```diff
@@ -0,0 +1,343 @@
...
+		peList := &akov2.AtlasPrivateEndpointList{}
+		listOpts := &client.ListOptions{
+			FieldSelector: fields.OneTermEqualSelector(
+				indexer.AtlasPrivateEndpointByProjectIndex,
+				client.ObjectKeyFromObject(atlasProject).String(),
+			),
+		}
+		err := r.Client.List(ctx, peList, listOpts)
+		if err != nil {
+			r.Log.Errorf("failed to list AtlasPrivateEndpoint: %e", err)
```
s-urbaniak, line 303: @helderjs is this a typo? `%e` is for scientific notation.

pkg/controller/atlasprivateendpoint/atlasprivateendpoint_controller.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+func (r *AtlasPrivateEndpointReconciler) getProjectFromAtlas(ctx context.Context, akoPrivateEndpoint *akov2.AtlasPrivateEndpoint) (*project.Project, error) {
+	sdkClient, _, err := r.AtlasProvider.SdkClient(
+		ctx,
+		&client.ObjectKey{Namespace: akoPrivateEndpoint.Namespace, Name: akoPrivateEndpoint.Credentials().Name},
+		r.Log,
+	)
+	if err != nil {
+		return nil, err
```
s-urbaniak, line 124: Please wrap errors here and elsewhere. Debugging this like so in error cases will not reveal stack traces.

pkg/controller/atlasprivateendpoint/atlasprivateendpoint_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	r.privateEndpointService = privateendpoint.NewPrivateEndpointAPI(sdkClient.PrivateEndpointServicesApi)
+
+	return project.NewProject(atlasProject, orgID), nil
+}
+
+func (r *AtlasPrivateEndpointReconciler) skip(ctx context.Context, akoPrivateEndpoint *akov2.AtlasPrivateEndpoint) (ctrl.Result, error) {
+	r.Log.Infow(fmt.Sprintf("-> Skipping AtlasPrivateEndpoint reconciliation as annotation %s=%s", customresource.ReconciliationPolicyAnnotation, customresource.ReconciliationPolicySkip), "spec", akoPrivateEndpoint.Spec)
+	if !akoPrivateEndpoint.GetDeletionTimestamp().IsZero() {
```
s-urbaniak, line 161: `skip` is a transition method while this code is asserting whether another transition is supposed to be executed. `skip` should only contain setting skip-state, not transition to other state. Checking deletion timestamp has to be done outside this method.

internal/collection/collection.go
s-urbaniak: I see this is being used in tests, please move it there and rename it to `randomKeyFromMap` and not export it.

internal/translation/privateendpoint/conversion.go
helderjs: @s-urbaniak the interface suffix, given the Go reserved word, is probably making you think this is an abstraction, but it's not. They are, indeed, the interface part of the private endpoint, that's why it's named like that.
With this, are you okay keeping the name or prefer to keep your suggestion?
Note: The example from data federation differ from here because the interface there is predefined and not a concept in the API

internal/collection/collection.go
s-urbaniak: @helderjs no, i think this method will be tempted to be used internally and I would like to prevent that.

internal/translation/privateendpoint/conversion.go
s-urbaniak: lol, indeed, resolving

s-urbaniak (approved):

josvazg (approved):

PR #987, opened 2023-06-07
Fixed encryption at REST aws.roleID

igor-karpukhin:
### All Submissions:

The field `encryptionAtRest.awskms.roleId` can now accept both: AtlasObjectID or AWS ARN

* [x] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if there is one).
* [ ] Update docs/release-notes/release-notes.md if your changes should be included in the release notes for the next release.

PR #1919, opened 2024-11-11
CLOUDP-228588: Migrate Encryption at Rest

roothorp:
### All Submissions:


internal/translation/encryptionatrest/conversion.go
```diff
@@ -0,0 +1,175 @@
...
+type EncryptionAtRest struct {
+	AWS   AwsKms
+	Azure AzureKeyVault
+	GCP   GoogleCloudKms
+}
+
+type AwsKms struct {
+	akov2.AwsKms
+	AccessKeyID                  string
+	SecretAccessKey              string
```
josvazg, line 20: These 2 are never sent nor received, are they needed?

internal/translation/encryptionatrest/conversion.go
```diff
@@ -0,0 +1,175 @@
...
+		KeyVersionResourceID: &g.KeyVersionResourceID,
+	}
+}
+
+func NewEncryptionAtRest(project *akov2.AtlasProject) *EncryptionAtRest {
+	ear := &EncryptionAtRest{}
+	spec := project.Spec.EncryptionAtRest
+	if spec == nil {
+		return nil
+	}
```
josvazg, line 115: nit

```suggestion
	spec := project.Spec.EncryptionAtRest
	if spec == nil {
		return nil
	}
	
	ear := &EncryptionAtRest{}
```

internal/translation/encryptionatrest/conversion.go
```diff
@@ -0,0 +1,175 @@
...
+
+func toAtlas(spec *EncryptionAtRest) *admin.EncryptionAtRest {
+	if spec == nil {
+		return nil
+	}
+
+	return &admin.EncryptionAtRest{
+		AwsKms:         spec.AWS.ToAtlas(),
+		AzureKeyVault:  spec.Azure.ToAtlas(),
+		GoogleCloudKms: spec.GCP.ToAtlas(),
```
josvazg, line 142: Why did we chose to go for a conversion method here rather than a function as usual?
I would prefer to stick to toAtlas/fromAtlas functions for consistency.

internal/translation/encryptionatrest/conversion_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	fuzz "github.com/google/gofuzz"
+	"github.com/stretchr/testify/require"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1/common"
+)
+
+func TestRoundtrip_EncryptionAtRest(t *testing.T) {
```
josvazg, line 13: Sample for making this < go native Fuzz test while still using `gofuzz`:

https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1884/files#diff-845d71c11ed89ac707e9ff64ff4273b11ab5f62125de863d4c0af3317e8f5086R23

pkg/controller/atlasproject/encryption_at_rest.go
```diff
@@ -140,57 +139,32 @@ func readSecretData(ctx context.Context, kubeClient client.Client, res common.Re
...
-		AzureKeyVault:  getAzureKeyVault(project),
-		GoogleCloudKms: getGoogleCloudKms(project),
-	}
-
-	if err := normalizeAwsKms(ctx, projectID, &requestBody.AwsKms); err != nil {
-		return err
+	if encRest != nil {
+		if err := normalizeAwsKms(ctx, projectID, &encRest.AWS); err != nil {
+			return workflow.Terminate(workflow.Internal, err.Error())
+		}
```
josvazg, line 157: Why only AWS needs this normalization?

pkg/controller/atlasproject/encryption_at_rest.go
```diff
@@ -206,14 +180,14 @@ func normalizeAwsKms(ctx *workflow.Context, projectID string, awsKms *mongodbatl
 	}
 
 	// assume that role ID is set as AWS ARN
-	resp, _, err := ctx.Client.CloudProviderAccess.ListRoles(ctx.Context, projectID)
+	resp, _, err := ctx.SdkClient.CloudProviderAccessApi.ListCloudProviderAccessRoles(ctx.Context, projectID).Execute()
```
josvazg, line 183: This call should also be wrapped within the translation layer, don't you think?

josvazg (commented):
Looking more tidy after the migration, good work!

I still have some questions here and there, though

internal/translation/encryptionatrest/conversion.go
roothorp: These are fields that Atlas seems to expect, but we don't seem to currently include. I don't want to add these fields (especially if they are mandatory) without further discussion with the team, but they exist in the API/SDK and are 'sensitive' fields.

internal/translation/encryptionatrest/conversion.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil
+	}
+
+	if *spec.AwsKms.Enabled {
+		ear.AWS.AwsKms = spec.AwsKms
+		for _, role := range project.Status.CloudProviderIntegrations {
+			if role.ProviderName == "AWS" {
+				ear.AWS.CloudProviderIntegrationRole = role.RoleID
+			}
+		}
```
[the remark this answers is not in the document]
roothorp: I'm not 100% sure - this is just taken from the current implementation, just this happens in the controller. I think we take the IAM Role from Cloud Provider Integrations if one is not explicitly defined in the encryption at rest CRD. I agree it's not ideal, but I'm trying to be cautious about rewriting or changing stuff like this at risk of changing the behaviour.

internal/translation/encryptionatrest/conversion.go
josvazg: I see, I think this merits a comment explaining the situation and maybe even a link to theAPI endpoint docs. Because looking at the rest of the code it is not obvious why these are here.

internal/translation/encryptionatrest/conversion.go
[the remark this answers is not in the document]
josvazg: Where did this came from initially? did you git blame it and check for context with the original author? (if that is possible)

If we cannot get context on why is this done in this way, I would at the very least add a comment, as it is another non obvious part of the conversion.

internal/translation/encryptionatrest/conversion.go
helderjs: Setups by access/secret keys are discouraged and I believe API doesn't accept it anymore, it's was kept to avoid breaking change, if I'm not wrong.

pkg/controller/atlasproject/encryption_at_rest.go
roothorp: Cloud Provider Access is a separate API which was migrated to the new SDK without adding a translation layer (in `cloud_provider_integration.go`). Adding a translation layer for this seems beyond the scope of this PR

s-urbaniak: agreed, translation layer was skipped for https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1402

pkg/controller/atlasproject/encryption_at_rest.go
s-urbaniak: This special case was implemented as part of https://github.com/mongodb/mongodb-atlas-kubernetes/pull/987. Maybe @igor-karpukhin has more historical context.

pkg/controller/atlasproject/encryption_at_rest.go
josvazg: Ok, understood. Could we make sure we have a Jira ticket for tackling this in the future?

pkg/controller/atlasproject/encryption_at_rest.go
```diff
@@ -44,27 +43,27 @@ func (r *AtlasProjectReconciler) ensureEncryptionAtRest(workflowCtx *workflow.Co
...
 
-func readEncryptionAtRestSecrets(kubeClient client.Client, service *workflow.Context, encRest *akov2.EncryptionAtRest, parentNs string) error {
+func readEncryptionAtRestSecrets(kubeClient client.Client, service *workflow.Context, encRest *encryptionatrest.EncryptionAtRest, parentNs string) error {
 	if encRest == nil {
 		return nil
 	}
 
-	if encRest.AwsKms.Enabled != nil && *encRest.AwsKms.Enabled && encRest.AwsKms.SecretRef.Name != "" {
-		err := readAndFillAWSSecret(service.Context, kubeClient, parentNs, &encRest.AwsKms)
+	if encRest.AWS.Enabled != nil && *encRest.AWS.Enabled && encRest.AWS.SecretRef.Name != "" {
```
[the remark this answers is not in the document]
s-urbaniak: We have to be careful with introducing custom public methods in the public types but this is "innocent" enough and I like it :-)

TLDR: fixed ;-)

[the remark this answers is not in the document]
s-urbaniak: in https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1919/commits/661c238449d34f43965ae0122ea2d0828589df63

internal/translation/encryptionatrest/conversion.go
s-urbaniak: good one, added 👍

pkg/controller/atlasproject/encryption_at_rest.go
s-urbaniak: that's a great idea, added https://jira.mongodb.org/browse/CLOUDP-285127

internal/translation/encryptionatrest/conversion.go
s-urbaniak: this is fixed

helderjs (approved):

internal/translation/encryptionatrest/conversion.go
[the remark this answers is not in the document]
josvazg: Should we leave a comment explaining why this is needed? I fear we will forget over time.

internal/translation/encryptionatrest/conversion.go
```diff
@@ -0,0 +1,176 @@
...
+	if ear.HasGoogleCloudKms() {
+		out.GCP.GoogleCloudKms = akov2.GoogleCloudKms{
+			Enabled: ear.GoogleCloudKms.Enabled,
+		}
+	}
+	return out
+}
+
+func EqualSpecs(spec, atlas *EncryptionAtRest) bool {
+	// Retracted secrets mean that this will never be equal
```
josvazg, line 174:
```suggestion
	// Redacted secrets mean that this will never be equal
```

josvazg: If this is true, why are we even doing the comparison?

s-urbaniak: huh, that's a good question, let me debug

josvazg (approved):
LGTM, just a few questions and I would probably add some clarifying comments on some special conversion cases around AWS.

internal/translation/encryptionatrest/conversion.go
s-urbaniak: debugged, this lacks normalization. we have to:
1. deepcopy the arguments
2. clear out (normalize) write only fields
3. then do a reflect.DeepEquals

igor-karpukhin (approved):

igor-karpukhin (approved):

PR #1920, opened 2024-11-11
CLOUDP-280723: Migrate Global Cluster

roothorp:
### All Submissions:


s-urbaniak:
this looks great so far, why is it a Draft? :-)

internal/translation/deployment/deployment.go
```diff
@@ -27,24 +28,33 @@ type AtlasDeploymentsService interface {
...
 	UpdateProcessArgs(ctx context.Context, cluster *Cluster) error
+
+	GetCustomZones(ctx context.Context, projectID, clusterName string) (map[string]string, error)
+	CreateCustomZones(ctx context.Context, projectID, clusterName string, mappings []akov2.CustomZoneMapping) (map[string]string, error)
+	DeleteCustomZones(ctx context.Context, projectID, clusterName string) error
+	GetZoneMappingMap(ctx context.Context, projectID, deploymentName string) (map[string]string, error)
+	GetManagedNamespaces(ctx context.Context, projectID, clusterName string) ([]akov2.ManagedNamespace, error)
+	CreateManagedNamespace(ctx context.Context, projectID, clusterName string, ns *akov2.ManagedNamespace) error
+	DeleteManagedNamespace(ctx context.Context, projectID, clusterName string, ns *akov2.ManagedNamespace) error
 }
```
josvazg, line 39: Not a blocker, but I'd rather have smaller more focused interface definitions. Note that interfaces can also be composed with embedding, so you can still get the mega-interface at the end. For instance, this interface is stating to look like a Java one, but it could instead be something like:

```go
type AtlasDeploymentsService interface {
	DeploymentsService
	ClustersService
	ProcessesService
	CustomZonesService
	ManagedNamespacesService
}
```

internal/translation/deployment/deployment.go
```diff
@@ -27,24 +28,33 @@ type AtlasDeploymentsService interface {
 	DeleteDeployment(ctx context.Context, deployment Deployment) error
 	ClusterWithProcessArgs(ctx context.Context, cluster *Cluster) error
 	UpdateProcessArgs(ctx context.Context, cluster *Cluster) error
+
+	GetCustomZones(ctx context.Context, projectID, clusterName string) (map[string]string, error)
+	CreateCustomZones(ctx context.Context, projectID, clusterName string, mappings []akov2.CustomZoneMapping) (map[string]string, error)
+	DeleteCustomZones(ctx context.Context, projectID, clusterName string) error
+	GetZoneMappingMap(ctx context.Context, projectID, deploymentName string) (map[string]string, error)
```
josvazg, line 35: Simplify, reduce rendundancy naming:

```suggestion
	GetZoneMapping(ctx context.Context, projectID, deploymentName string) (map[string]string, error)
```

internal/translation/deployment/conversion.go
```diff
@@ -915,3 +915,63 @@ func set[T any](nameFn func(T) string, lists ...[]T) []T {
 	}
 	return result
 }
+
+func customZonesFromAtlas(gs *admin.GeoSharding) *[]akov2.CustomZoneMapping {
```
josvazg, line 919: This one seems only used so that fuzzing round trips can be achieved, right?

josvazg (approved):
LGTM, great work!

All my other comments are non blocking

helderjs (approved):
LGTM. Added 2 suggestions

internal/translation/deployment/deployment.go
```diff
@@ -215,3 +232,69 @@ func (ds *ProductionAtlasDeployments) UpdateProcessArgs(ctx context.Context, clu
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get cluster: %w", err)
+	}
+	result := make(map[string]string, len(cluster.GetReplicationSpecs()))
+	for _, rc := range cluster.GetReplicationSpecs() {
+		result[rc.GetId()] = rc.GetZoneName()
+	}
+	return result, nil
+}
+
```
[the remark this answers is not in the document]
roothorp: If we'd prefer it converted I can do so, but since the Atlas API returns Custom Zones as a map I figured we could just stick with a native type. I appreciate this isn't strictly adhering to the idea behind the translation layer, but it seemed easier

internal/translation/deployment/deployment.go
```diff
@@ -29,22 +35,33 @@ type AtlasDeploymentsService interface {
 	UpdateProcessArgs(ctx context.Context, cluster *Cluster) error
 }
 
+type GlobalClusterService interface {
```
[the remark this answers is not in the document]
roothorp: I'm happy to do that - I'm not sure how we want to organise the translation layer for the larger CRDs/APIs (Deployment, Project, etc.) and if we want to have everything under one roof or spin them out into individual parts

internal/translation/deployment/deployment.go
roothorp: I've tried this locally, and it looks like mockery generates duplicate mock files - one for the 'mega-interface' and then again for each of the sub-interfaces (which contain the same method). Maybe I'm missing a difference, and maybe it doesn't matter, but I thought I would flag it before adding it

PR #1928, opened 2024-11-13
CLOUDP-283049: Fix Data Federation private endpoint reconciliation

helderjs:
### All Submissions:

* [x] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [ ] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if there is one).
* [ ] Update docs/release-notes/release-notes-template.md if your changes should be included in the release notes for the next release.

PR #1929, opened 2024-11-14
CLOUDP-283049: fix data federation endpoint reconciliation

s-urbaniak:
This is an alternative to #1928 and introduces a state machine handling.

It also prevents accidental deletion of data federation private endpoints that are not tracked by AKO.

### All Submissions:

### DIFF
--- a/internal/translation/datafederation/conversion_endpoints.go
+++ b/internal/translation/datafederation/conversion_endpoints.go
@@ -18,7 +18,7 @@ type DatafederationPrivateEndpointEntry struct {
 	ProjectID string
 }
 
-func NewDatafederationPrivateEndpointEntry(projectID string, pe *akov2.DataFederationPE) *DatafederationPrivateEndpointEntry {
+func NewDataFederationPrivateEndpointEntry(projectID string, pe *akov2.DataFederationPE) *DatafederationPrivateEndpointEntry {
 	if pe == nil {
 		return nil
 	}
@@ -92,11 +92,11 @@ func MapDatafederationPrivateEndpoints(projectID string, df *akov2.AtlasDataFede
 	}
 	for _, pe := range df.Spec.PrivateEndpoints {
 		addEndpoint(pe.EndpointID)
-		result[pe.EndpointID].AKO = NewDatafederationPrivateEndpointEntry(projectID, &pe)
+		result[pe.EndpointID].AKO = NewDataFederationPrivateEndpointEntry(projectID, &pe)
 	}
 	for _, pe := range lastApplied.Spec.PrivateEndpoints {
 		addEndpoint(pe.EndpointID)
-		result[pe.EndpointID].LastApplied = NewDatafederationPrivateEndpointEntry(projectID, &pe)
+		result[pe.EndpointID].LastApplied = NewDataFederationPrivateEndpointEntry(projectID, &pe)
 	}
 
 	return result, nil
--- a/internal/translation/datafederation/conversion_endpoints.go
+++ b/internal/translation/datafederation/conversion_endpoints.go
@@ -1,13 +1,16 @@
 package datafederation
 
 import (
+	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"go.mongodb.org/atlas-sdk/v20231115008/admin"
 
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/cmp"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/pointer"
 	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/customresource"
 )
 
 type DatafederationPrivateEndpointEntry struct {
@@ -15,11 +18,15 @@ type DatafederationPrivateEndpointEntry struct {
 	ProjectID string
 }
 
-func NewDatafederationPrivateEndpointEntry(pe *akov2.DataFederationPE, projectID string) *DatafederationPrivateEndpointEntry {
+func NewDatafederationPrivateEndpointEntry(projectID string, pe *akov2.DataFederationPE) *DatafederationPrivateEndpointEntry {
 	if pe == nil {
 		return nil
 	}
 	return &DatafederationPrivateEndpointEntry{DataFederationPE: pe, ProjectID: projectID}
+}
+
+func (e *DatafederationPrivateEndpointEntry) EqualsTo(target *DatafederationPrivateEndpointEntry) bool {
+	return reflect.DeepEqual(e.DataFederationPE, target.DataFederationPE)
 }
 
 func endpointsFromAtlas(endpoints []admin.PrivateNetworkEndpointIdEntry, projectID string) ([]*DatafederationPrivateEndpointEntry, error) {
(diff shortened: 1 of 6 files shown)


internal/translation/datafederation/conversion_endpoints.go
```diff
@@ -58,3 +65,39 @@ func endpointToAtlas(ep *DatafederationPrivateEndpointEntry) *admin.PrivateNetwo
...
+}
+
+func MapDatafederationPrivateEndpoints(projectID string, df *akov2.AtlasDataFederation, atlasEndpoints []*DatafederationPrivateEndpointEntry) (map[string]*DataFederationPrivateEndpoint, error) {
+	var lastApplied akov2.AtlasDataFederation
+	if ann, ok := df.GetAnnotations()[customresource.AnnotationLastAppliedConfiguration]; ok {
+		err := json.Unmarshal([]byte(ann), &lastApplied.Spec)
+		if err != nil {
+			return nil, fmt.Errorf("error reading data federation from last applied annotation: %w", err)
+		}
+	}
```
josvazg, line 80: nit

Being purist I would do this outside of the translation layer. Basically the translation layer should be Kubernetes unaware. We could pass this as an argument.

Not blockling the PR due to this, but design wise I think is better to not tangle the translation with the concept of annotations in kube objects.

internal/translation/datafederation/conversion_endpoints.go
```diff
@@ -58,3 +65,39 @@ func endpointToAtlas(ep *DatafederationPrivateEndpointEntry) *admin.PrivateNetwo
...
+
+	result := make(map[string]*DataFederationPrivateEndpoint)
+	addEndpoint := func(endpointID string) {
+		if _, ok := result[endpointID]; !ok {
+			result[endpointID] = &DataFederationPrivateEndpoint{}
+		}
+	}
```
josvazg, line 87: Why was a closure preferred to a simple external private function such as this?
```go
func(result map[string]*DataFederationPrivateEndpoint, endpointID string) {
	if _, ok := result[endpointID]; !ok {
		result[endpointID] = &DataFederationPrivateEndpoint{}
	}
}
```


Closures are required when the closure needs to travel to another context, bringing its context with it. This is not the case.

s-urbaniak: This closes over `result`

igor-karpukhin (approved):

internal/translation/datafederation/conversion_endpoints.go
s-urbaniak: I actually don't agree in this case.

This is definitely part of the translation layer, it introspects public types and its internal state, in this case annotation, and generates internal types to be consumed by the controller, in this case a map.

pkg/controller/atlasdatafederation/datafederation_controller.go
```diff
@@ -116,8 +116,8 @@ func (r *AtlasDataFederationReconciler) Reconcile(context context.Context, req c
 		return result.ReconcileResult(), nil
 	}
 
-	if result = r.ensurePrivateEndpoints(ctx, project, dataFederation, endpointService); !result.IsOk() {
-		ctx.SetConditionFromResult(api.DataFederationPEReadyType, result)
+	if result = r.ensurePrivateEndpoints(ctx, endpointService, project, dataFederation); !result.IsOk() {
```
josvazg, line 119: why the order changes?

pkg/controller/atlasdatafederation/datafederation_controller.go
```diff
@@ -116,8 +116,8 @@ func (r *AtlasDataFederationReconciler) Reconcile(context context.Context, req c
...
+		ctx.SetConditionFromResult(api.DataFederationReadyType, result)
```
josvazg, line 120: This was another bug?

pkg/controller/atlasdatafederation/datafederation_controller.go
```diff
@@ -145,7 +145,7 @@ func (r *AtlasDataFederationReconciler) Reconcile(context context.Context, req c
 		return r.handleDelete(ctx, log, dataFederation, project, dataFederationService).ReconcileResult(), nil
 	}
 
-	err = customresource.ApplyLastConfigApplied(context, project, r.Client)
+	err = customresource.ApplyLastConfigApplied(context, dataFederation, r.Client)
```
josvazg, line 148: Seems like another bug fixed here right?

internal/translation/datafederation/conversion_endpoints.go
s-urbaniak: > the translation layer should be Kubernetes unaware. 

I agree but this is not the case here. We are introspecting the structure of the public type. The annotation is a simple map, we encode previous state in here using JSON, there is nothing Kubernetes specific in here.

pkg/controller/atlasdatafederation/private_endpoint.go
```diff
@@ -1,64 +1,103 @@
...
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/datafederation"
-
-	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/set"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api"
 	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/workflow"
 )
 
-func (r *AtlasDataFederationReconciler) ensurePrivateEndpoints(ctx *workflow.Context, project *akov2.AtlasProject, dataFederation *akov2.AtlasDataFederation, service datafederation.DatafederationPrivateEndpointService) workflow.Result {
+func (r *AtlasDataFederationReconciler) ensurePrivateEndpoints(ctx *workflow.Context, service datafederation.DatafederationPrivateEndpointService, project *akov2.AtlasProject, dataFederation *akov2.AtlasDataFederation) workflow.Result {
```
josvazg, line 12: I would not change order of params, it is an internal API and does not matter much, but still there is no reason for the change AFAIK

pkg/controller/atlasdatafederation/datafederation_controller.go
s-urbaniak: yes, it's part of the ticket being solved.

pkg/controller/atlasdatafederation/private_endpoint.go
s-urbaniak: see https://github.com/mongodb/mongodb-atlas-kubernetes/pull/1929#discussion_r1846084044

s-urbaniak: I actually suggest to change the variable ordering, it makes the signature clearer.

pkg/controller/atlasdatafederation/private_endpoint.go
```diff
@@ -1,64 +1,103 @@
...
+	ctx.UnsetCondition(api.DataFederationPEReadyType)
+	return workflow.OK()
+}
+
+type PrivateEndpointReconciler struct {
+	service  datafederation.DatafederationPrivateEndpointService
+	endpoint *datafederation.DataFederationPrivateEndpoint
+}
+
+func (r *PrivateEndpointReconciler) Reconcile(ctx *workflow.Context) error {
```
josvazg, line 60: It would be good to point to the state machine diagram doc somehow

s-urbaniak: you mean the dot plot? I think this is a very simple state machine, I suggest to address this as a follow-up if you believe it is worth doing.

internal/translation/datafederation/conversion_endpoints.go
josvazg: Actually, you are right, we are not using any Kubernetes interfaces or APIs here. Annotations are just maps, so yes, I guess is fine.

PR #1930, opened 2024-11-14
CLOUDP-280230: Network peering controller and CRD

josvazg:
Network peering support is sliced into 3 logical changes:

1. Fixing to the already merged translation layer.
2. Adding the CRD types and manifest files.
3. Adding the controller code and tests.
4.  🚧 **(WIP) Unit tests**

### All Submissions:


config/crd/bases/atlas.mongodb.com_atlasprivateendpoints.yaml
```diff
@@ -106,8 +106,8 @@ spec:
                 - name
                 type: object
               externalProjectRef:
-                description: ExternalProject holds the Atlas project ID the user belongs
-                  to
```
josvazg, line 110: cc @helderjs have you find this typo in your change? The description name should match the actual field name

josvazg:
Helm test fail because the CRDs in the helm package still do not include network peering.

api/v1/atlasnetworkpeering_types_test.go
```diff
@@ -0,0 +1,19 @@
+package v1 // nolint: dupl
+
+import (
+	"testing"
+)
+
+func TestNetworkPeeringProjectRefCELValidations(t *testing.T) {
+	launchProjectRefCELTests(
```
s-urbaniak, line 8: q: what is being asserted here? `launchProjectRefCELTests` has no indication what test cases are being executed.

api/v1/atlasnetworkpeering_types.go
```diff
@@ -16,6 +16,58 @@ limitations under the License.
...
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.status`
+// +kubebuilder:printcolumn:name="Container Provisioned",type=string,JSONPath=`.status.containerProvisioned`
+// +kubebuilder:subresource:status
+// +groupName:=atlas.mongodb.com
+// +kubebuilder:resource:categories=atlas,shortName=anp
+type AtlasNetworkPeering struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   AtlasNetworkPeeringSpec          `json:"spec,omitempty"`
```
s-urbaniak, line 45: q: are we following the rules described in https://ahmet.im/blog/crd-generation-pitfalls/index.html#explicit-required-or-optional-on-every-field for all spec fields?

internal/operator/builder.go
```diff
@@ -219,6 +220,18 @@ func (b *Builder) Build(ctx context.Context) (cluster.Cluster, error) {
 		return nil, fmt.Errorf("unable to create indexers: %w", err)
 	}
 
+	npeReconciler := atlasnetworkpeering.NewAtlasNetworkPeeringsReconciler(
```
s-urbaniak, line 223: blocking: this needs to be in the controller registry https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/internal/controller/registry.go

internal/controller/atlasnetworkpeering/atlasnetworkpeering_controller.go
```diff
@@ -0,0 +1,433 @@
...
+	var atlasPeer *networkpeering.NetworkPeer
+	if req.networkPeering.Status.ID != "" {
+		peer, err := req.service.GetPeer(req.workflowCtx.Context, req.projectID, req.networkPeering.Status.ID)
+		if err != nil && !errors.Is(err, networkpeering.ErrNotFound) {
+			return r.terminate(req.workflowCtx, req.networkPeering, workflow.Internal, err)
+		}
+		atlasPeer = peer
+	}
+	inAtlas := atlasPeer != nil
+	deleted := req.networkPeering.DeletionTimestamp != nil
```
s-urbaniak, line 168: blocking: this doesn't catch the case of deletion timestamp being zero. Please use the `IsZero()` method which asserts both nil and zero time.

internal/indexer/indexer.go
```diff
@@ -18,9 +18,9 @@ type Indexer interface {
 // RegisterAll registers all known indexers to the given manager.
 // It uses the given logger to create a new named "indexer" logger,
 // passing that to each indexer.
-func RegisterAll(ctx context.Context, c cluster.Cluster, logger *zap.Logger) error {
+func RegisterAll(ctx context.Context, mgr manager.Manager, logger *zap.Logger) error {
```
s-urbaniak, line 21: blocking: undo this change, we need to retain `c cluster.Cluster` as a parameter.

internal/controller/atlasnetworkpeering/atlasnetworkpeering_controller.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+	req.workflowCtx.EnsureStatusOption(updatePeeringStatusOption(newPeer))
+	return workflow.InProgress(
+		workflow.NetworkPeeringConnectionCreating,
+		fmt.Sprintf("Network Peering Connection %s is %s",
+			req.networkPeering.Status.ID, req.networkPeering.Status.Status),
+	).ReconcileResult()
+}
+
+func (r *AtlasNetworkPeeringReconciler) sync(req *reconcileRequest, atlasPeer *networkpeering.NetworkPeer) ctrl.Result {
```
s-urbaniak, line 202: blocking: what is being `sync`'ed here? as I read the code, updates in the spec are silently ignored. If the spec is immutable, we should talk w/ PM whether it makes sense to make the whole spec immutable (guarded via CEL) if Atlas doesn't support updates to network peers.

s-urbaniak (changes requested):
a couple of comments, great work! 👏

internal/controller/atlasnetworkpeering/atlasnetworkpeering_controller.go
s-urbaniak: I see we have updates in Atlas for both connections https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Network-Peering/operation/updatePeeringConnection and containers https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Network-Peering/operation/updatePeeringContainer.

Why aren't we leveraging that?

josvazg:
Making this a draft again, as we plan to re-design the Peering Connection - Network Container relatinship, to make Networks a first class CRD as well.

Still I would like to address most of the comments as I will be using this branch for continuing work and this and will need it to be rebase-able as close to main as possible.

api/v1/atlasnetworkpeering_types.go
josvazg: I will double check that for when we surface this again, both for the NetworkContainer and the PeeringConnection CRDs.

api/v1/atlasnetworkpeering_types_test.go
josvazg: This is interesting, because all cases were always the same for all the users. The thing is, the only change was the dual reference was used by one or another underlying CRD type, but that does not change the behaviour or correctness of how the dual ref is handled.


We could fix this in a separate PR, I think we can have a single test for this in a single place and just keep on adding underlaying CRDs to test as carriers of the dual reference to be tested.
WDYT?

internal/controller/workflow/reason.go
```diff
@@ -151,3 +151,11 @@ const (
 	PrivateEndpointFailedToConfigure        ConditionReason = "PrivateEndpointFailedToConfigure"
 	PrivateEndpointFailedToDelete           ConditionReason = "PrivateEndpointFailedToDelete"
 )
+
+// Atlas Network Peering reasons
+const (
+	NetworkPeeringConnectionCreating  ConditionReason = "NetworkPeeringConnectionCreating"
+	NetworkPeeringConnectionResetting ConditionReason = "NetworkPeeringConnectionResetting"
```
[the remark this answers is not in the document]
josvazg: will remove

internal/indexer/indexer.go
josvazg: Right! I missed this detail on the rebase, I will fix

internal/operator/builder.go
josvazg: Right! yep I also need to fix this for comply with the new **Cluster** support

internal/controller/workflow/reason.go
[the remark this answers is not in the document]
josvazg: done in next update

josvazg:
This has been split into 2 sequential PRs: #2101 & #2103
Once I am sure all the relevant bits have been moved to those, I shall close this PR.

josvazg:
No longer required, replaced by #2100 & #2103

PR #1955, opened 2024-11-26
CLOUDP-286703: Fix bogus Reaps

### LINKED ISSUE #1954: AtlasDatabaseUser connection secrets removed after upgrading
**What did you do to encounter the bug?**
We upgraded mongodb-atlas-kubernetes to version 2.5.0 (we came from version 2.3.1). This failed pretty hard and caused downtime for our customers. Somehow the connection secrets were removed for most of our AtlasDatabaseUser objects. This caused our depending deployments to not work anymore. Reverting to 2.3.1 resolved the issues.

**What did you expect?**

The first reconciliation with 2.5.0 removed random connection secrets. We have a setup where:

* We use serverless instances.
* We have one AtlasDeployment per namespace
* We have multiple AtlasDatabaseUser crds in our namespace. One for each application, which also gives the needed rights.
* We expect the same number of connection secrets in our namespace as AtlasDatabaseUsers.
* We have a lot off AtlasDatabaseUser objects over the whole cluster: so the syncing takes it's time (no problem but it made it more difficult to pinpoint).
* This all works fine with 2.3.1

**What happened instead?**

We somehow only had one connection secret per namespace after 2.5.0 started (instead of 5): and it showed weird behavior. It looked like the secrets where deleted. Enabling the debug logs showed that this was indeed right (see next Alinea).

**Screenshots**
image

**Operator Information**
Operator 2.3.1 does work (we downgraded). Operator 2.5.0 does not.

We run Kubernetes 1.30 in EKS.

**Probable cause**

* I think the bug is in this [piece of code](https://github.com/mongodb/mongodb-atlas-kubernetes/blob/7b11f5360f756d430af389334fad9e4535382bb5/pkg/controller/atlasdatabaseuser/databaseuser.go#L185C1-L201C3). The ListClusters is using [this library under the hood](https://github.com/mongodb/atlas-sdk-go/blob/74bb99188a3949dcc09aea44b5137176d58abcf6/admin/api_clusters.go#L211C305-L211C360). Which explicitly states that serverless instances are not returned in this call.
* In our situation - we only have serverless clusters - this means that the list is empty and that all connection secrets are marked as orphaned.

If this is all true, then I do have a request: could you please add a log line at warning level when connection secrets are deleted. I have carefully tested this upgrade on my test cluster and I would have seen these lines.

josvazg:
Closes #1954

**Check the last commit for the fix**, specifically the main fix is in `internal/translation/deployment/deployment.go` making the list of cluster names actually be a list of deployments in general, hence the rename on the method and thus the refactors and mockery changes in the first commit. The list now includes the serverless deployments, so it detects properly that secrets attached to serverless instances must not be reaped.

Commits breakdown:
- `Use same default operator namespace`: Small fix to use the same namespace in contract test and other local testing.
- `Fix contract testing controller runtime logging`: Fix to set the logger to avoid controller runtime complains about it.
- `Test list includes clusters and serverless`: Reproduction error failing before the fix.
- `Refactor mocks and add mockery to devbox`: Refactors before fix to mockery and devbox.
- `Fix list to include serverless as well as clusters`: The actual fix code isolated.

### All Submissions:

* [X] Put `closes #1954` in your comment to auto-close the issue that your PR fixes (if there is one).


josvazg:
Verified the bug is caught by the new test before the fix:
```shell
        	Error:      	Not equal: 
        	            	expected: []string{"cluster0", "serverless-name"}
        	            	actual  : []string{"cluster0"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,3 @@
        	            	-([]string) (len=2) {
        	            	- (string) (len=8) "cluster0",
        	            	- (string) (len=15) "serverless-name"
        	            	+([]string) (len=1) {
        	            	+ (string) (len=8) "cluster0"
        	            	 }

```

s-urbaniak (approved):

s-urbaniak (approved):

PR #1956, opened 2024-11-26
CLOUDP-286489: Atlas Search now works with the new API

igor-karpukhin:
### All Submissions:

Atlas Search now works with the new API


pkg/controller/atlas/provider.go
```diff
@@ -30,10 +32,16 @@ const (
...
 	SdkClient(ctx context.Context, secretRef *client.ObjectKey, log *zap.SugaredLogger) (*admin.APIClient, string, error)
+	SdkClientSet(ctx context.Context, secretRef *client.ObjectKey, log *zap.SugaredLogger) (*ClientSet, string, error)
 	IsCloudGov() bool
 	IsResourceSupported(resource api.AtlasCustomResource) bool
 }
 
+type ClientSet struct {
+	//SdkClient20231115008 *admin.APIClient
+	SdkClient20241113001 *adminv20241113001.APIClient
+}
```
josvazg, line 43: So the plan is to keep removing clients from the Provider, such as `Client` and `SdkClient` and keep on managing the new ones only on the `ClientSet`?

I guess we do that to avoid having to change the client usages on exiting code.

internal/mocks/atlas/provider.go
```diff
@@ -3,6 +3,8 @@ package atlas
 import (
 	"context"
 
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/atlas"
```
josvazg, line 6: Please put this one with the other import  from `github.com/mongodb/mongodb-atlas-kubernetes/v2` so that goimports can order it properly.

Canonical imports structure we would like to converge to is:
- Stdlib packages at the top.
- External dependencies in the middle.
- Internal dependencies within AKO at the bottom.

internal/translation/searchindex/searchindex.go
```diff
@@ -323,12 +323,14 @@ func (s *SearchIndex) Normalize() (*SearchIndex, error) {
 }
 
 // toAtlas converts internal SearchIndex representation to the Atlas structure used for API calls
-func (s *SearchIndex) toAtlas() (*admin.ClusterSearchIndex, error) {
-	convertJSONToListOfMaps := func(in *apiextensionsv1.JSON) (*[]map[string]interface{}, error) {
+//
+//nolint:dupl
```
josvazg, line 327: This here might indicate a code smell, such as we are duplicating code that could be avoided. Probably sliced into smaller reusable functions?

igor-karpukhin:
> LGTM. we are attaching a new dependency to workflow.Context but it should be okay until we have a final plan to it.
Yes, I think it could be a good discussion topic for the kick-off's optional part

pkg/controller/atlas/provider.go
igor-karpukhin: Yes, I think having a clientSet with a group of all clients could be a good solution, instead of extending the `Provider` interface every time we add a new version of Go SDK

internal/mocks/atlas/provider.go
igor-karpukhin: This is done with the goimport and linted with the golangci lint. I didn't put it there manually.

josvazg: This happens to me all the time, you slip an extra end of line, create a new group. Then `goimports` respects all the divisions you do. If you have 5 divisions it will sort each of the 5 divisions. But we only want to have 3. @s-urbaniak wanted to start using https://github.com/daixiang0/gci so that there is no ambiguity on our imports ordering and layout.

internal/translation/searchindex/searchindex.go
josvazg: In general the idiom `funcName := func(...) {...}` to make a closure does not usually pay off, you can always add the argument to de function and make it a regular internal function of the module. That also makes the caller method shorter and easier to read.

In this example I would just have instead elsewhere:
```go
func convertJSONToListOfMaps(in *apiextensionsv1.JSON /*add any extra args needed here*/) (*[]map[string]interface{}, error) {
...
}
```

Why are the named closures a better fit here?
Hopefully they are not repeated across functions, cause in that case all the more reason to define them once externally.

internal/mocks/atlas/provider.go
s-urbaniak: agreed with @josvazg  but not a blocker for this PR

s-urbaniak (approved):

PR #1971, opened 2024-12-05
CLOUDP-280230: Refactor adds Project Dual Reference

josvazg:
While the refactor change accept many files, there are only a few key places that are key:
- The new **ProjectDualReference** at `pkg/api/v1/projectref.go` encapsulates together the 3 fields that are required for all CRDs that depend on a project and have dual mode of operation with independent IDs and Kubernetes references.
- The credentials types at `pkg/api/credentials.go` are simplifies all types but **CredentialsProvider** interface disappear and a new **ObjectwithCredentials** allows indexers to define kubertenes objects that carry credentials, such is the case of these dual references.
- All CRDs supporting independent usage are changed to use the new **ProjectDualReference**.

All the rest of changes are derived from there:
- Some CRD descriptions are changes to become consistent under `config/crd/bases`. **There are no changes to the YAML**, this is a refactor.
- Some reference namings are corrected to be consistent across all types now.
- All tests need to be fixed with the new type shape, fields are moved around, and sometimes renamed.

This PR contains 5 commits, the new code infrastructure and 4 more for each type migrated to use it, db users, custom roles, private endpoints and deployments.

### All Submissions:

### All Submissions:


pkg/api/v1/projectref.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// ProjectDualReference encapsulates together the frequnet construct to refer to
+// a parent projet by a Kubernetes references, or an external ID, which also
+// reqwuires access credentials in such case
+type ProjectDualReference struct {
+	// "projectRef" is a reference to the parent AtlasProject resource.
+	// Mutually exclusive with the "externalProjectRef" field
+	// +kubebuilder:validation:Optional
+	Project *common.ResourceRefNamespaced `json:"projectRef,omitempty"`
```
s-urbaniak, line 15: Non-blocking (just a nit)
```suggestion
	ProjectRef *common.ResourceRefNamespaced `json:"projectRef,omitempty"`
```
I suggest keeping the field names and the json serializer names in sync unless it is criticially necessary to deviate, otherwise this can easily cause confusion.

s-urbaniak (approved):
one non-blocking comment. feel free to ignore if you prefer the current field name. Beware though that changing public field names does change our API surface so I suggest keeping factors like this to a minimum in the future.

As a reminder: everything in `pkg/api/...` **is** part of our API surface.

josvazg:
> one non-blocking comment. feel free to ignore if you prefer the current field name. Beware though that changing public field names does change our API surface so I suggest keeping factors like this to a minimum in the future.
> 
> As a reminder: everything in `pkg/api/...` **is** part of our API surface.

True. I am fixing the name, but be ware that before this "refactor", we were already inconsistent in that naming across CRDs, the field name, not the YAML name, luckily.

josvazg:
The e2e tests are quite broken by this. I am going to slice the change further, not only to easy to review, but also to more easily detect where e2e tests got broken.

pkg/controller/reconciler/reconciler.go
```diff
@@ -0,0 +1,91 @@
...
+
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+)
+
+type Reconciler struct {
+	Client client.Client
+	Log    *zap.SugaredLogger
+}
+
+func (r *Reconciler) SolveProject(ctx context.Context, sdkClient *admin.APIClient, pro akov2.ProjectReferrerObject, orgID string) (*project.Project, error) {
```
s-urbaniak, line 21:
```suggestion
func (r *Reconciler) ResolveProject(ctx context.Context, sdkClient *admin.APIClient, pro akov2.ProjectReferrerObject, orgID string) (*project.Project, error) {
```

pkg/controller/reconciler/reconciler.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, fmt.Errorf("failed to get Project from Kubernetes: %w", err)
+	}
+	return project.NewProject(kubeProject, orgID), nil
+}
+
+func (r *Reconciler) projectFromAtlas(ctx context.Context, projectsService project.ProjectService, pdr *akov2.ProjectDualReference) (*project.Project, error) {
+	return projectsService.GetProject(ctx, pdr.ExternalProjectRef.ID)
+}
+
+func (r *Reconciler) SolveCredentials(ctx context.Context, pro akov2.ProjectReferrerObject) (*client.ObjectKey, error) {
```
s-urbaniak, line 50:
```suggestion
func (r *Reconciler) ResolveProjectCredentials(ctx context.Context, pro akov2.ProjectReferrerObject) (*client.ObjectKey, error) {
```

pkg/api/v1/projectref.go
```diff
@@ -0,0 +1,40 @@
...
+	ProjectDualRef() *ProjectDualReference
+}
+
+// +k8s:deepcopy-gen=false
+
+// ProjectReferrerObject is an project referrer that is also an Kubernetes Object
+type ProjectReferrerObject interface {
+	client.Object
+	ProjectReferrer
+}
```
s-urbaniak, line 40: **question**: Note that these interfaces all become public API surface. Do we really want to give the API stability guarantees here? I would rather suggest to put these in the translation layer.

pkg/controller/reconciler/reconciler.go
```diff
@@ -0,0 +1,91 @@
...
+	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+	"go.uber.org/zap"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/project"
+
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/api/v1"
+)
+
+type Reconciler struct {
```
s-urbaniak, line 16: **question**: the type implements project specific reconciliation bits, that is resolving atlas projects via kube or atlas and resolving credentials for a project however the type is very abstract named "Reconciler" and clashes with controller-runtime concepts. I am much rather inclined to name it concretely:

```suggestion
type AtlasProjectReconciler struct {
```

s-urbaniak (changes requested):
one API concern and naming suggestion, else LGTM

pkg/controller/reconciler/reconciler.go
josvazg: Ok will apply with the IDE

pkg/controller/reconciler/reconciler.go
josvazg: same, from IDE

pkg/api/v1/projectref.go
josvazg: I can move this interface inside

pkg/controller/reconciler/reconciler.go
josvazg: uhm but ti is not an atlas project reconciler 

What about `AtlasReconciler` instead?

s-urbaniak: much better 👍

josvazg: renamed like that

pkg/api/v1/projectref.go
josvazg: done

pkg/controller/reconciler/reconciler.go
josvazg: done

pkg/controller/reconciler/reconciler.go
josvazg: done

s-urbaniak (approved):

helderjs (changes requested):

pkg/controller/atlascustomrole/custom_role.go
```diff
@@ -11,55 +11,43 @@ import (
...
+		service:   service,
 		role:      akoCustomRole,
 		dpEnabled: dpEnabled,
 		k8sClient: k8sClient,
 	}
 
-	result := r.Reconcile()
-	ctx.SetConditionFromResult(api.ReadyType, result)
-	return result
+	return r.Reconcile()
```
helderjs, line 41: Same here, removal `ctx.SetConditionFromResult(api.ReadyType, result)` makes the resource never be ready. I believe it also would feat better into the `idle()` transition instead of here. But, please, make sure move it to other place fix the problem.

helderjs (changes requested):

pkg/controller/atlascustomrole/custom_role.go
josvazg: This refactor is not about any changes to the reconciliation logic. I will just fix the bug to leave the logic as it was with the new dual ref in place. We can make a separate PR if we need to fix the logic as well.

pkg/controller/atlascustomrole/atlascustomrole_controller.go
```diff
@@ -224,7 +194,7 @@ func (r *AtlasCustomRoleReconciler) notFound(req ctrl.Request) ctrl.Result {
 func (r *AtlasCustomRoleReconciler) SetupWithManager(mgr ctrl.Manager, skipNameValidation bool) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("AtlasCustomRole").
-		For(&akov2.AtlasCustomRole{}, builder.WithPredicates(r.GlobalPredicates...)).
+		For(&akov2.AtlasCustomRole{}).
```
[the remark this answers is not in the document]
josvazg: fixed

pkg/controller/atlascustomrole/custom_role.go
josvazg: updated

PR #1998, opened 2024-12-17
CLOUDP-289150: Fix private endpoint conflict between sub-resource and new CRD

helderjs:
### All Submissions:

### DIFF
--- a/pkg/api/v1/atlasproject_types.go
+++ b/pkg/api/v1/atlasproject_types.go
@@ -17,9 +17,6 @@ limitations under the License.
 package v1
 
 import (
-	"encoding/json"
-	"fmt"
-
 	"go.uber.org/zap/zapcore"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -296,22 +293,6 @@ func (p *AtlasProject) WithIPAccessList(ipAccess project.IPAccessList) *AtlasPro
 func (p *AtlasProject) WithMaintenanceWindow(window project.MaintenanceWindow) *AtlasProject {
 	p.Spec.MaintenanceWindow = window
 	return p
-}
-
-func (p *AtlasProject) LastSpecFrom(annotation string) (*AtlasProjectSpec, error) {
-	var lastApplied AtlasProject
-	ann, ok := p.GetAnnotations()[annotation]
-
-	if !ok {
-		return nil, nil
-	}
-
-	err := json.Unmarshal([]byte(ann), &lastApplied.Spec)
-	if err != nil {
-		return nil, fmt.Errorf("error reading AtlasProject Spec from annotation [%s]: %w", annotation, err)
-	}
-
-	return &lastApplied.Spec, nil
 }
 
 func DefaultProject(namespace, connectionSecretName string) *AtlasProject {
--- a/pkg/api/v1/atlasproject_types.go
+++ b/pkg/api/v1/atlasproject_types.go
@@ -17,6 +17,9 @@ limitations under the License.
 package v1
 
 import (
+	"encoding/json"
+	"fmt"
+
 	"go.uber.org/zap/zapcore"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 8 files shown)


pkg/api/v1/atlasproject_types.go
```diff
@@ -295,6 +298,22 @@ func (p *AtlasProject) WithMaintenanceWindow(window project.MaintenanceWindow) *
 	return p
 }
 
+func (p *AtlasProject) LastSpecFrom(annotation string) (*AtlasProjectSpec, error) {
```
s-urbaniak, line 301: I would really like us to avoid attaching methods in the public API surface. This nicely fits into the translation layer.

pkg/controller/atlasproject/private_endpoint.go
```diff
@@ -516,9 +530,16 @@ func terminateWithError(ctx *workflow.Context, conditionType api.ConditionType,
...
 var notReadyInterfaceResult = workflow.InProgress(workflow.ProjectPEInterfaceIsNotReadyInAtlas, "Interface Private Endpoint is not ready")
 
-func getEndpointsNotInSpec(specPEs []akov2.PrivateEndpoint, atlasPEs []atlasPE) []atlasPE {
-	uniqueItems, _ := getUniqueDifference(atlasPEs, specPEs)
-	return uniqueItems
+func getEndpointsNotInSpec(specPEs []akov2.PrivateEndpoint, atlasPEs []atlasPE, lastAppliedPEs map[string]akov2.PrivateEndpoint) []atlasPE {
+	notInSpecItems, _ := getUniqueDifference(atlasPEs, specPEs)
+	toDelete := make([]atlasPE, 0, len(notInSpecItems))
+	for _, item := range notInSpecItems {
+		if _, ok := lastAppliedPEs[item.Identifier().(string)]; ok {
```
s-urbaniak, line 537: this uses the deprecated `DeprecatedIdentifiable` interface. we should move this sort of comparisons in the translation layer and not rely on deprecated functionality.

s-urbaniak (changes requested):
two changes which I would like to see before we merge:
- Don't expose translation layer specific functions in the public API surface
- Don't use `DeprecatedIdentifiable` which is ... deprecated and we know it has rough edges

pkg/api/v1/atlasproject_types.go
helderjs: It cannot be on translation layer because is to be used on the sub-resources, that are not using translation layer, at least PEs are not. Anyway, I will move it to the controller instead.

pkg/controller/atlasproject/private_endpoint.go
helderjs: Again, this sub-resource doesn't use translation layer. Unless we migrate everything to translation layer, it's not possible to get rid of `DeprecatedIdentifiable`

pkg/api/v1/atlasproject_types.go
s-urbaniak: In that case let's at least have this method **not** in the public API surface.

pkg/controller/atlasproject/private_endpoint.go
s-urbaniak: indeed, resolving, thank you!

s-urbaniak (approved):

s-urbaniak:
final result looks great! 🎉

pkg/controller/atlasproject/private_endpoint.go
```diff
@@ -585,3 +598,31 @@ func hasSkippedPrivateEndpointConfiguration(atlasProject *akov2.AtlasProject) (b
...
+}
+
+func hasManagedPrivateEndpoints(specPEs []akov2.PrivateEndpoint, atlasPEs []atlasPE, lastAppliedPEs map[string]akov2.PrivateEndpoint) bool {
+	for _, pe := range atlasPEs {
+		if _, ok := lastAppliedPEs[pe.Identifier().(string)]; ok {
+			return true
+		}
+	}
+
+	return len(set.DeprecatedDifference(specPEs, atlasPEs)) == 0
```
josvazg, line 627: why are we using a deprecated diff on new code?

helderjs: As mentioned in previous comments, while there's new addition to this part of the code, it still depends on deprecated types/interfaces unless fully refactor to use translation layer, which is not a goal.

josvazg (approved):
LGTM, but would prefer not to keep on using deprecated calls on new changes

PR #2016, opened 2025-01-02
pkg, internal: add pagination to list requests

s-urbaniak:
### All Submissions:


internal/translation/paging/page_request_test.go
```diff
@@ -0,0 +1,146 @@
...
+			wantResult: []string{"a", "b"},
+		},
+		{
+			name: "multiple results with empty results",
+			pages: []*page{
+				{results: []string{"a", "b"}},
+				{results: []string{}},
+				{results: []string{"c", "d"}},
+			},
+			wantResult: []string{"a", "b"},
```
roothorp, line 133: To be clear; are we expecting to ignore the 3rd page since when we see an empty page we assume that we have run out of results?

roothorp (approved):

internal/translation/paging/page_request_test.go
s-urbaniak: Correct, that's the current implementation, taken over from terraform. Semantically it is debatable what is more correct, however I don't expect results like these from Atlas. Checking the expected behavior is beneficial though.

PR #2026, opened 2025-01-08
CLOUDP-280012: Add http dry run transport for Atlas

s-urbaniak:
This adds a dry run transport towards Atlas calls. Note that this contribution is a no-op for now. The operator starts in non-dry-run mode.

### All Submissions:

### DIFF
--- a/internal/controller/atlas/provider.go
+++ b/internal/controller/atlas/provider.go
@@ -8,16 +8,18 @@ import (
 	"runtime"
 	"strings"
 
-	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+	"github.com/mongodb-forks/digest"
+	adminv20231115008 "go.mongodb.org/atlas-sdk/v20231115008/admin"
 	adminv20241113001 "go.mongodb.org/atlas-sdk/v20241113001/admin"
 	"go.mongodb.org/atlas/mongodbatlas"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	//v20241113001
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/api"
 	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/dryrun"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/httputil"
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/version"
 )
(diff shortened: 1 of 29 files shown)


helderjs (approved):

internal/controller/atlas/provider.go
```diff
@@ -8,16 +8,18 @@ import (
 	"runtime"
 	"strings"
 
-	"go.mongodb.org/atlas-sdk/v20231115008/admin"
+	"github.com/mongodb-forks/digest"
+	adminv20231115008 "go.mongodb.org/atlas-sdk/v20231115008/admin"
```
josvazg, line 12: nit

do we need to distinguish up to the hour, I think year moth is probably enough right?

helderjs: it's has nothing with hour, sdk versioning has an incremental 3 digits composing the major version

josvazg: true, my bad, still I do not think we need anything beyond yyyymm internally

s-urbaniak: I'd leave as is and not deviate from the SDK pkg names

helderjs: adding to Serge's comment, I do see the possibility of having adminv20231115007 and adminv20231115008 at the same time, for example. Better to match pkg name pattern and be on the safe side of integration/compatibility.

josvazg: ok, then agreed

igor-karpukhin (approved):

roothorp (approved):
Looks good! Just to clarify my understanding of the feature; this is going to be used to we can create all of the Atlas API calls we _would_ make, but without actually sending them, letting us preview what a change would do?

internal/controller/atlas/provider.go
```diff
@@ -121,48 +127,58 @@ func (p *ProductionProvider) Client(ctx context.Context, secretRef *client.Objec
 	return c, secretData.OrgID, err
 }
 
-func (p *ProductionProvider) SdkClient(ctx context.Context, secretRef *client.ObjectKey, log *zap.SugaredLogger) (*admin.APIClient, string, error) {
-	secretData, err := getSecrets(ctx, p.k8sClient, secretRef, &p.globalSecretRef)
+func (p *ProductionProvider) SdkClient(ctx context.Context, secretRef *client.ObjectKey, log *zap.SugaredLogger) (*adminv20231115008.APIClient, string, error) {
```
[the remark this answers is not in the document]
s-urbaniak: I think i did in the comment

internal/controller/atlas/provider.go
```diff
@@ -121,48 +127,58 @@ func (p *ProductionProvider) Client(ctx context.Context, secretRef *client.Objec
...
+	secretData, err := getSecrets(ctx, p.k8sClient, secretRef, &p.globalSecretRef)
 	if err != nil {
 		return nil, "", err
 	}
 
-	return c, secretData.OrgID, nil
-}
+	var transport http.RoundTripper = digest.NewTransport(secretData.PublicKey, secretData.PrivateKey)
+	transport = p.newDryRunTransport(transport)
+	transport = httputil.NewLoggingTransport(log, false, transport)
```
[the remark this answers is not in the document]
s-urbaniak: Yes, i think so!

PR #2071, opened 2025-01-27
CLOUDP-280230: Refactor dual ref CEL validations

josvazg:
Group all dual references CEL tests in a single test battery.

Related to works on CLOUDP-280230 Network peering CRD support

### All Submissions:


s-urbaniak (changes requested):
blocking: this is missing a ticket

josvazg:
CLOUDP-280230

api/v1/project_reference_cel_test.go
```diff
@@ -1,130 +1,170 @@
...
+const (
+	crdBasePath = "../../config/crd/bases/"
+)
 
-type celValidationTestCase struct {
-	title          string
-	obj, old       AtlasCustomResource
-	expectedErrors []string
+type projectReferrer interface {
+	ProjectDualRef() *ProjectDualReference
```
s-urbaniak, line 23: **nit**: (almost blocking but ok) interfaces are not about data but about behaviour in Go. It is not idiomatic to use interfaces as an abstraction of "getters", see https://go.dev/doc/effective_go#interfaces_and_types:

> Interfaces in Go provide a way to specify the **behavior** of an object: if something can do _this_, then it can be used _here_. 

I can dig up more blogs/videos about this but we should avoid this pattern in general.

s-urbaniak (approved):

api/v1/project_reference_cel_test.go
josvazg: I disagree. This `projectReferrer` interface is a local version of `project.ProjectReferrer` which is crucial to be able to uniformly handle all CRDs that provide a dual project reference. This is also not a new pattern as all our CRDs require several methods that in the end just provide access to status data, or other bits like that, and are needed to conform to custom resources interface.

Even more, the Kubernetes Object interface also uses that pattern to access things like the name and namespace of CRDs uniformly.

helderjs (approved):

api/v1/project_reference_cel_test.go
s-urbaniak: yes, Kubernetes deviates from this Go idiom as well.

josvazg: Ok, so this is not a new deviation.

In any case, I would love to hear alternatives that allow for using the dual refs uniformly across several CRDs without falling into this.

s-urbaniak: totally fine to go in as is 👍 i'll experiment locally maybe to have a better solution.

api/v1/project_reference_cel_test.go
```diff
@@ -1,130 +1,174 @@
...
-type CELTestObjectGenFunc func(*ProjectDualReference) AtlasCustomResource
+const (
+	crdBasePath = "../../config/crd/bases/"
+)
 
-type celValidationTestCase struct {
-	title          string
-	obj, old       AtlasCustomResource
-	expectedErrors []string
+type projectReferrer interface {
```
s-urbaniak, line 22: keeping it package is the way to go 👏

api/v1/project_reference_cel_test.go
```diff
@@ -1,130 +1,174 @@
...
-			unstructuredOldObject, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&tc.old)
-			require.NoError(t, err)
-
-			validator, err := cel.VersionValidatorFromFile(t, crdPath, "v1")
-			assert.NoError(t, err)
-			errs := validator(unstructuredObject, unstructuredOldObject)
-
-			require.Equal(t, len(tc.expectedErrors), len(errs))
+type projectReferrerObject interface {
+	client.Object
```
s-urbaniak, line 27: nit: `runtime.Object` is good enough as it satisfies `DeepCopyObject()`

s-urbaniak: `client.Object` has accessors (getters/setters) to metadata which is not needed here.

api/v1/project_reference_cel_test.go
```diff
@@ -1,130 +1,174 @@
...
-
-			unstructuredOldObject, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&tc.old)
-			require.NoError(t, err)
-
-			validator, err := cel.VersionValidatorFromFile(t, crdPath, "v1")
-			assert.NoError(t, err)
-			errs := validator(unstructuredObject, unstructuredOldObject)
-
-			require.Equal(t, len(tc.expectedErrors), len(errs))
+type projectReferrerObject interface {
```
s-urbaniak, line 26: nit:
```suggestion
type projectReferrerObject interface {
	runtime.Object
	ProjectDualRef() *ProjectDualReference
}
```
and remove `projectReferrer`

roothorp (approved):

api/v1/project_reference_cel_test.go
josvazg: Let me try, I believe that is not as easy and that idiom is a contract interface, not a plain interface? will check

josvazg: Nope, I was wrong, this works fine. Next update will use this.

josvazg: Done

s-urbaniak (approved):

igor-karpukhin (approved):

PR #2080, opened 2025-01-30
CLOUDP-280258: add For() method in all reconcilers for dry-run

s-urbaniak:
We have to make sure our custom resources controllers support dry run capabilities. We have the necessary metadata already available already during registration in the regular Manager object, namely the runtime object a reconciler is responsible for.

We need the runtime object for the same reasons controller-runtime needs it in the manager. We need to list resources and act on them during dry-run so the runtime object is needed.

This adds a `For()` method for every reconciler to be used during dry-run.

**Note** This is a no-op, however it's fine to merge this after the 2.7 release.

### All Submissions:

### DIFF
--- a/internal/controller/atlasbackupcompliancepolicy/atlasbackupcompliancepolicy_controller.go
+++ b/internal/controller/atlasbackupcompliancepolicy/atlasbackupcompliancepolicy_controller.go
@@ -96,10 +96,14 @@ func (r *AtlasBackupCompliancePolicyReconciler) ensureAtlasBackupCompliancePolic
 	return r.release(workflowCtx, bcp)
 }
 
+func (r *AtlasBackupCompliancePolicyReconciler) For() (client.Object, builder.Predicates) {
+	return &akov2.AtlasBackupCompliancePolicy{}, builder.WithPredicates(r.GlobalPredicates...)
+}
+
 func (r *AtlasBackupCompliancePolicyReconciler) SetupWithManager(mgr ctrl.Manager, skipNameValidation bool) error {
 	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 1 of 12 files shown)


josvazg (commented):
The description does explain this refactor is useful for dry-run but it does not explain why. This is important as it will not be obvious until we see it used on a subsequent PR.

Code LGTM, BTW.

s-urbaniak:
> The description does explain this refactor is useful for dry-run but it does not explain why. This is important as it will not be obvious until we see it used on a subsequent PR.

It does:

> We have to make sure our custom resources controllers support dry run capabilities. We have the necessary metadata already available already during registration in the regular Manager object, namely the runtime object a reconciler is responsible for.

igor-karpukhin (approved):

s-urbaniak:
added some more clarification:
> We need the runtime object for the same reasons controller-runtime needs it in the manager. We need to list resources and act on them during dry-run so the runtime object is needed.

josvazg:
> > The description does explain this refactor is useful for dry-run but it does not explain why. This is important as it will not be obvious until we see it used on a subsequent PR.
> 
> It does:
> 
> > We have to make sure our custom resources controllers support dry run capabilities. We have the necessary metadata already available already during registration in the regular Manager object, **namely the runtime object a reconciler is responsible for.**

I guess the missing piece of info is dry run does **not** use the same **SetupWithManager** method as the regular mode of working does, therefore it needs some info that was only available there, such as the For piece.

This might be obvious to people working on the feature and known the code that is coming in the future, but it is not generally available knowledge.

josvazg (approved):
LGTM

helderjs (approved):

PR #2085, opened 2025-02-03
CLOUDP-280012: internal/dryrun/transport: simplify

s-urbaniak:
Good news: we can nicely simplify the dry run transport, there is no need to dependency inject the event recorder to it. As a consequence, we also don't need to bubble down the runtime object any more which is needed for recording events.

It turned out this is all doable in the dry-run manager itself.

**Note** this has to be merged after the 2.7 release only.

### All Submissions:


internal/dryrun/transport.go
```diff
@@ -37,24 +30,13 @@ func (t *DryRunTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
-
 		verb, ok := verbMap[req.Method]
 		if !ok {
 			verb = "execute " + req.Method
 		}
 		msg := "Would %v %v"
-		t.Recorder.Eventf(obj, v1.EventTypeNormal, DryRunReason, msg, verb, req.URL.Path)
 
-		return nil, NewDryRunError(obj.GetObjectKind(), meta, v1.EventTypeNormal, DryRunReason, msg, verb, req.URL.Path)
+		return nil, NewDryRunError(nil, nil, "", "", msg, verb, req.URL.Path)
```
josvazg, line 39: **q:** If these are always going to be `nil, nil, "", ""`do they even make sense at all as `NewDryRunError` args?

s-urbaniak: no, since this is a stacked PR, a follow-up is going to simplify the dry run error constructor as well 👍

igor-karpukhin, line 39: nit: feels like way to many parameters. Consider using builder. Not critical

igor-karpukhin (approved):
LGTM 👍

josvazg (approved):
LGTM

(given we will simplify the dry-run error constructor in a follow-up)

internal/dryrun/transport.go
s-urbaniak: see https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2085#discussion_r1939463080, this is going away.

PR #2102, opened 2025-02-06
CLOUDP-297404: Handle deployment features in non-blocking way

igor-karpukhin:
### All Submissions:

Handle deployment features in non-blocking way

### DIFF
--- a/internal/controller/atlasdeployment/advanced_deployment.go
+++ b/internal/controller/atlasdeployment/advanced_deployment.go
@@ -69,18 +69,15 @@ func (r *AtlasDeploymentReconciler) handleAdvancedDeployment(ctx *workflow.Conte
 			return r.terminate(ctx, workflow.DeploymentConnectionSecretsNotCreated, err)
 		}
 
+		var results []workflow.Result
 		if !r.AtlasProvider.IsCloudGov() {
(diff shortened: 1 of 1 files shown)


s-urbaniak (approved):

internal/controller/atlasdeployment/advanced_deployment.go
```diff
@@ -101,10 +96,13 @@ func (r *AtlasDeploymentReconciler) handleAdvancedDeployment(ctx *workflow.Conte
 			akoCluster.GetCustomResource().Spec.DeploymentSpec.ManagedNamespaces,
 			akoCluster.GetName(),
 		)
-		if transition = r.transitionFromResult(ctx, deploymentService, akoCluster.GetProjectID(), akoCluster.GetCustomResource(), result); transition != nil {
-			return transition(workflow.Internal)
-		}
+		results = append(results, result)
 
+		for i := range results {
+			if !results[i].IsOk() {
```
josvazg, line 102: **q:** Instead of failing on the first error, why not join all errors into one with `errors.Join`?
I see we are not handling errors directly here, but there might be a way to extract those errors.

igor-karpukhin: It is the same as with the atlas project. We can do that, but we fail for first error, saying what went wrong. Anyway, status should be updated already. This is just for the reconcile result.

PR #2103, opened 2025-02-06
CLOUDP-280230: Add Network peering CRD and controller

josvazg:
This adds the new Network Peering CRD and controller support, together with the Network Container CRD, this completely deprecated the Network Peering Connection sub-resources within the Atlas Project CRD under the project controller.

The PR consists on 3 steps completing the functionality, to be included as individual commits:

- Add AtlasNetworkPeering CRD.
- Wire the helm chart changes to support the new CRD.
- Add Network Peering controller.

### All Submissions:


config/rbac/kustomization.yaml
```diff
@@ -50,3 +50,7 @@ resources:
 - atlasproject_viewer_role.yaml
 - atlasdeployment_editor_role.yaml
 - atlasdeployment_viewer_role.yaml
+- atlasnetworkpeering_editor_role.yaml
+- atlasnetworkpeering_viewer_role.yaml
+
+
```
s-urbaniak, line 56: **nit**: superfluous newlines

internal/controller/atlasnetworkpeering/state.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (r *AtlasNetworkPeeringReconciler) handle(workflowCtx *workflow.Context, req *reconcileRequest) (ctrl.Result, error) {
+	r.Log.Infow("handling network peering reconcile request",
+		"service set", (req.service != nil), "projectID", req.projectID, "networkPeering", req.networkPeering)
+	container, err := r.getContainer(workflowCtx.Context, req)
+	if err != nil {
+		return r.terminate(workflowCtx, req.networkPeering, workflow.Internal, err)
+	}
+	if container == nil {
```
s-urbaniak, line 76: **q**: a container *MUST* exist for a network peer, right? At least that is what this code is enforcing.

internal/controller/atlasnetworkpeering/state.go
```diff
@@ -0,0 +1,138 @@
...
+		err := fmt.Errorf("container not found for reference %v", req.networkPeering.Spec.ContainerRef)
+		return r.terminate(workflowCtx, req.networkPeering, workflow.NetworkPeeringMissingContainer, err)
+	}
+	var atlasPeer *networkpeering.NetworkPeer
+	if req.networkPeering.Status.ID != "" {
```
s-urbaniak, line 81: **q**: how do we handle the import case? That is, associating existing network peers via dry-run or CLI's `atlas kubernetes config generate`?

config/rbac/kustomization.yaml
josvazg: will clear

internal/controller/atlasnetworkpeering/state.go
josvazg: Yes, the peering connection is just that a connection. Without the "container", which is basically a virtual network, there is nothing to connect to. So the container is a hard pre-requisite, it must exist for the peering to be possible.

josvazg: Do you think we should add a comment here to further clarify?

internal/controller/atlasnetworkpeering/state.go
josvazg: The ID is Atlas given, autogenerated and there is no name to be matched. Just like for the container.

We could add an ID which will be used for that import case but be optional otherwise. I need to double check how that affects the sate machine. I guess it will be similar to the container, when the ID is set the peering is expected to exist already and not finding it is a blocking error. Also the ID once set is immutable.

s-urbaniak (commented):
Leaving in idle state as there is one blocking question regarding the "import" case, else LGTM.

internal/controller/atlasnetworkpeering/state.go
s-urbaniak: I'd say so yes, for the future reader.

internal/controller/atlasnetworkpeering/state.go
s-urbaniak: right but in the container case we can reference existing containers in Atlas by ID, we cannot do the same for network peers.

s-urbaniak: > We could add an ID which will be used for that import case but be optional otherwise.

yeah, pretty similar semantics as for the container case. I believe the old subresource code does matching by name somehow?

josvazg: Yeah, the old matching was brittle. Let's go for the same solution as the container which matches the fact the API does not provider any unique naming, just unique IDs.

internal/translation/networkcontainer/conversion.go
```diff
@@ -73,3 +99,32 @@ func fromAtlasNoStatus(container *admin.CloudProviderContainer) *NetworkContaine
...
 	}
 }
+
+func fromAtlasAWSStatus(container *admin.CloudProviderContainer) *AWSContainerStatus {
+	if container.VpcId == nil {
+		return nil
+	}
+	return &AWSContainerStatus{
+		VpcID: container.GetVpcId(),
+	}
```
helderjs, line 109: Just a hint on how SDK expect such cases to be handled.
```suggestion
	if ok, vpcID := container.GetVpcIdOk(); ok {
		return &AWSContainerStatus{
		    VpcID: vpcID,
	        }
	}
	return nil
```

helderjs (approved):
LGTM

api/v1/atlasnetworkpeering_types.go
```diff
@@ -16,14 +16,76 @@ limitations under the License.
...
+// AtlasNetworkPeeringList contains a list of AtlasNetworkPeering
+type AtlasNetworkPeeringList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []AtlasNetworkPeering `json:"items"`
+}
+
+// +kubebuilder:validation:XValidation:rule="(has(self.externalProjectRef) && !has(self.projectRef)) || (!has(self.externalProjectRef) && has(self.projectRef))",message="must define only one project reference through externalProjectRef or projectRef"
+// +kubebuilder:validation:XValidation:rule="(has(self.externalProjectRef) && has(self.connectionSecret)) || !has(self.externalProjectRef)",message="must define a local connection secret when referencing an external project"
+// +kubebuilder:validation:XValidation:rule="(has(self.containerRef.name) && !has(self.containerRef.id)) || (!has(self.containerRef.name) && has(self.containerRef.id))",message="must either have a container Atlas id or Kubernetes name, but not both (or neither)"
```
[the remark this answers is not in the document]
josvazg: done! thanks for the heads up!

roothorp (approved):

config/rbac/kustomization.yaml
igor-karpukhin, line 56: **non-blocking**: redundant new line

config/samples/atlas_v1_atlasnetworkpeering.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: atlasnetworkpeering-sample
+spec:
+  projectRef:
+    name: atlas-project
+    namespace: namespace
+  provider: AWS
+  containerId: "623412394512350"
+  awsConfiguration:
+    accepterRegionName: US_EAST_1 # TODO: double check sample values
```
igor-karpukhin, line 12: **q**: Is this TODO still needed?

internal/controller/atlasnetworkpeering/state.go
```diff
@@ -0,0 +1,146 @@
...
+		service:          networkpeering.NewNetworkPeeringServiceFromClientSet(sdkClientSet),
+		containerService: networkcontainer.NewNetworkContainerServiceFromClientSet(sdkClientSet),
+		projectID:        project.ID,
+		networkPeering:   networkPeering,
+	})
+}
+
+func (r *AtlasNetworkPeeringReconciler) handle(workflowCtx *workflow.Context, req *reconcileRequest) (ctrl.Result, error) {
+	r.Log.Infow("handling network peering reconcile request",
+		"service set", (req.service != nil), "projectID", req.projectID, "networkPeering", req.networkPeering)
```
igor-karpukhin, line 71:
```suggestion
		"service set", req.service != nil, "projectID", req.projectID, "networkPeering", req.networkPeering)
```

internal/controller/atlasnetworkpeering/state.go
```diff
@@ -0,0 +1,146 @@
...
+func (r *AtlasNetworkPeeringReconciler) getContainer(ctx context.Context, req *reconcileRequest) (*networkcontainer.NetworkContainer, error) {
+	id := req.networkPeering.Spec.ContainerRef.ID
+	if req.networkPeering.Spec.ContainerRef.ID == "" { // Name should be non nil instead
+		var err error
+		id, err = getContainerIDFromKubernetes(ctx, r.Client, req.networkPeering)
+		if err != nil {
+			return nil, fmt.Errorf("failed to solve Network Container id from Kubernetes: %w", err)
+		}
+		if id == "" {
+			return nil, fmt.Errorf("container %s has no id, is it still to be created?",
```
igor-karpukhin, line 116:
```suggestion
			return nil, fmt.Errorf("container %s has no id, waiting for it to be created",
```

internal/translation/networkpeering/networkpeering.go
```diff
@@ -2,141 +2,102 @@ package networkpeering
...
+	ErrNotFound = errors.New("not found")
 
-type PeeringContainerService interface {
-	CreateContainer(ctx context.Context, projectID string, container *ProviderContainer) (*ProviderContainer, error)
-	GetContainer(ctx context.Context, projectID, containerID string) (*ProviderContainer, error)
-	ListContainers(ctx context.Context, projectID, providerName string) ([]ProviderContainer, error)
-	DeleteContainer(ctx context.Context, projectID, containerID string) error
-}
+	// ErrContainerInUse is a failure to remove a containe still in use
+	ErrContainerInUse = errors.New("container still in use")
```
igor-karpukhin, line 19: **non-blocking**: this seems to be unused

igor-karpukhin (approved):
Great job! See some non-blocking comments

internal/translation/networkpeering/networkpeering.go
josvazg: Good catch! Thanks!

api/v1/project_reference_cel_test.go
```diff
@@ -165,7 +173,14 @@ func TestProjectDualReferenceCELValidations(t *testing.T) {
 				assert.NoError(t, err)
 				errs := validator(unstructuredObject, unstructuredOldObject)
 
-				require.Equal(t, tc.expectedErrors, cel.ErrorListAsStrings(errs))
+				for i, err := range errs {
+					fmt.Printf("%s error %d: %v\n", title, i, err)
```
s-urbaniak, line 177: **nit**
```suggestion
					t.Logf("%s error %d: %v\n", title, i, err)
```

internal/controller/atlasnetworkcontainer/state.go
```diff
@@ -16,14 +15,17 @@ import (
...
+)
+
 type reconcileRequest struct {
 	projectID        string
 	networkContainer *akov2.AtlasNetworkContainer
 	service          networkcontainer.NetworkContainerService
 }
 
 func (r *AtlasNetworkContainerReconciler) handleCustomResource(ctx context.Context, networkContainer *akov2.AtlasNetworkContainer) (ctrl.Result, error) {
-	typeName := reflect.TypeOf(*networkContainer).Name()
```
s-urbaniak, line 26: wonderful 👏

PR #2142, opened 2025-02-20
CLOUDP-301800: dryrun: add support for namespace filtering

s-urbaniak:
Currently, Atlas Operator respects the “WATCH_NAMESPACE” environment variable which lets you filter namespaces the operator is watching.

We need to retain this behavior for the dry-run as well as the current code base is wiring cache setup with namespace bindings so we have to respect that in the dry-run manager implementation as well, otherwise behavior will cause errors during the dry-run.

### All Submissions:


igor-karpukhin (approved):
LGTM

PR #2158, opened 2025-02-26
CLOUDP-294105: remove kube lookups in Atlas Provider

s-urbaniak:
This finalizes the transition to remove Kubernetes lookups in the Atlas Provider and just use client sets in the translation layer.

Note: we want to merge this **after the 2.8 release**

### All Submissions:


josvazg (commented):
LGTM, but this now needs rebasing

josvazg:
lint and unit tests are unhappy

internal/controller/atlas/provider.go
```diff
@@ -43,24 +38,28 @@ type ClientSet struct {
...
-	dryRun          bool
+	domain string
+	dryRun bool
 }
 
-type credentialsSecret struct {
+// Credentials is the type that holds credentials to authenticate against the Atlas API.
+// Currently, only API keys are support but more credential types could be added,
+// see https://www.mongodb.com/docs/atlas/configure-api-access/.
+type Credentials struct {
```
helderjs, line 48: I'd have one struct (merge `Credentials` and `ApiKeys`) and remove `OrgID` field.
The org id is not needed to setup the client and the function would be only giving it back to the caller.

internal/controller/reconciler/credentials.go
```diff
@@ -0,0 +1,100 @@
...
+	key := client.ObjectKeyFromObject(pro)
+	pdr := pro.ProjectDualRef()
+	if pdr.ConnectionSecret == nil {
+		return nil
+	}
+	key.Name = pdr.ConnectionSecret.Name
+	return &key
+}
+
+func GetAtlasCredentials(ctx context.Context, k8sClient client.Client, secretRef, fallbackRef *client.ObjectKey) (*atlas.Credentials, error) {
```
helderjs, line 57: Why isn't it part of reconciler struct? where kube client is already available

helderjs, line 57: Given my suggestion on the provider, I'd return the org id separate from credentials.
```suggestion
func GetAtlasCredentials(ctx context.Context, k8sClient client.Client, secretRef, fallbackRef *client.ObjectKey) (string, *atlas.Credentials, error) {
```

helderjs (changes requested):

internal/controller/atlas/provider.go
s-urbaniak: No, i disagree, we have to be open to add additional key types in the future. This struct prepares us to be ready for that.

internal/controller/reconciler/credentials.go
s-urbaniak: because not all reconcilers implement the `AtlasReconciler` where this is intrinsic.

s-urbaniak: No, the orgId is specifically part of the public/private key tuple, this might be different for other types of credentials, i.e. OAuth opaque access tokens where we have no information about the encoded orgID.

internal/controller/atlas/provider.go
s-urbaniak: but, I think you have one good point though, I believe we can remove the `orgID` at all from the credentials setup 🤔  Let me check if that is possible 👍

internal/controller/atlasproject/team_reconciler.go
```diff
@@ -58,14 +53,8 @@ func (r *AtlasProjectReconciler) teamReconcile(
...
-		atlasClient, orgID, err := r.AtlasProvider.SdkClientSet(teamCtx.Context, connectionSecretKey, log)
-		if err != nil {
-			result := workflow.Terminate(workflow.AtlasAPIAccessNotConfigured, err)
-			setCondition(teamCtx, api.ReadyType, result)
-			return result.ReconcileResult(), nil
-		}
-		teamCtx.OrgID = orgID
-		teamCtx.SdkClient = atlasClient.SdkClient20231115008
+		teamCtx.OrgID = workflowCtx.OrgID
+		teamCtx.SdkClient = workflowCtx.SdkClientSet.SdkClient20231115008
```
helderjs, line 57: You're passing nil here because the client was set to a different field inn the project controller:
https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2158/files#diff-0d383863ebbfca7afb47ff12d141447ce823fbcd48870db9722d7b5a53cb3e15R165

This is making the Gov (because of teams) and Teams tests to fail

```suggestion
		teamCtx.SdkClient = workflowCtx.SdkClientSet.SdkClient
```

helderjs (changes requested):

josvazg (approved):
LGTM!

helderjs (approved):

PR #2159, opened 2025-02-26
CLOUDP-303005: Allow Serverless Alerts to specify decimal thresholds

tvdw-mongodb:
This fixes the IsEqual function to compare using floats instead of integers, so that the operator will reconcile values that change by less than 1 unit.


igor-karpukhin:
Looks like we have different CI process for PRs from forks. Closed in favor of: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2160

PR #2192, opened 2025-03-11
CLOUDP-299772: Remove last skipped config

josvazg:
Replace it by:
- Clearing the resource lists for migrated independent CRDs at skip time. So that they will not eagerly try to take over  the full resource list when skip is removed and potentially conflict with existing independent CRs.
- Without last applied config, non greediness behaviour kicks in and the legacy embedded resource controller will not delete resources from Atlas that does not know about.

Testing: Added one test verifying that migrated resource lists get cleared when the project reconcile is skipped, and several tests cases per migrated resource to verify non greedy behaviour conforms to expectations.

### All Submissions:

### DIFF
--- a/internal/controller/atlasproject/atlasproject_controller.go
+++ b/internal/controller/atlasproject/atlasproject_controller.go
@@ -378,7 +378,7 @@ func lastAppliedSpecFrom(atlasProject *akov2.AtlasProject) (*akov2.AtlasProjectS
 }
 
 func (r *AtlasProjectReconciler) clearLastAppliedMigratedResources(ctx context.Context, atlasProject *akov2.AtlasProject) error {
-	lastCfg, err := customresource.ParseLastConfigApplied(&akov2.AtlasProjectSpec{}, atlasProject)
+	lastCfg, err := customresource.ParseLastConfigApplied[akov2.AtlasProjectSpec](atlasProject)
 	if err != nil {
 		return fmt.Errorf("failed to parse last applied config annotation: %w", err)
 	}
--- a/internal/controller/atlasproject/atlasproject_controller.go
+++ b/internal/controller/atlasproject/atlasproject_controller.go
@@ -388,7 +388,12 @@ func (r *AtlasProjectReconciler) clearLastAppliedMigratedResources(ctx context.C
 	lastCfg.PrivateEndpoints = nil
 	lastCfg.ProjectIPAccessList = nil
 	lastCfg.NetworkPeers = nil
-	if err := customresource.ApplyLastConfigApplied(ctx, &akov2.AtlasProject{Spec: *lastCfg}, r.Client); err != nil {
+	shallowCopy := akov2.AtlasProject{
+		TypeMeta:   atlasProject.TypeMeta,
+		ObjectMeta: atlasProject.ObjectMeta,
+		Spec:       *lastCfg,
+	}
+	if err := customresource.ApplyLastConfigApplied(ctx, &shallowCopy, r.Client); err != nil {
 		return fmt.Errorf("failed to apply last applied config annotation: %w", err)
 	}
 	return nil
--- a/internal/controller/atlasproject/atlasproject_controller.go
+++ b/internal/controller/atlasproject/atlasproject_controller.go
@@ -378,23 +378,19 @@ func lastAppliedSpecFrom(atlasProject *akov2.AtlasProject) (*akov2.AtlasProjectS
 }
 
 func (r *AtlasProjectReconciler) clearLastAppliedMigratedResources(ctx context.Context, atlasProject *akov2.AtlasProject) error {
-	lastCfg, err := customresource.ParseLastConfigApplied[akov2.AtlasProjectSpec](atlasProject)
+	clearedCfg, err := customresource.ParseLastConfigApplied[akov2.AtlasProjectSpec](atlasProject)
 	if err != nil {
 		return fmt.Errorf("failed to parse last applied config annotation: %w", err)
 	}
 	// clear all resources migrated as independent CRDs to avoid eager
 	// reconciliation that might conflict with independent CRs and apply
-	lastCfg.CustomRoles = nil
-	lastCfg.PrivateEndpoints = nil
-	lastCfg.ProjectIPAccessList = nil
-	lastCfg.NetworkPeers = nil
-	shallowCopy := akov2.AtlasProject{
-		TypeMeta:   atlasProject.TypeMeta,
-		ObjectMeta: atlasProject.ObjectMeta,
-		Spec:       *lastCfg,
-	}
-	if err := customresource.ApplyLastConfigApplied(ctx, &shallowCopy, r.Client); err != nil {
-		return fmt.Errorf("failed to apply last applied config annotation: %w", err)
+	clearedCfg.CustomRoles = nil
+	clearedCfg.PrivateEndpoints = nil
+	clearedCfg.ProjectIPAccessList = nil
+	clearedCfg.NetworkPeers = nil
+
+	if err := customresource.PatchLastConfigApplied(ctx, r.Client, atlasProject, clearedCfg); err != nil {
+		return fmt.Errorf("failed to clear migrated resources in last applied config annotation: %w", err)
 	}
 	return nil
 }
--- a/internal/controller/atlasproject/atlasproject_controller.go
+++ b/internal/controller/atlasproject/atlasproject_controller.go
@@ -382,6 +382,9 @@ func (r *AtlasProjectReconciler) clearLastAppliedMigratedResources(ctx context.C
 	if err != nil {
 		return fmt.Errorf("failed to parse last applied config annotation: %w", err)
 	}
+	if clearedCfg == nil { // nothing to patch
+		return nil
+	}
 	// clear all resources migrated as independent CRDs to avoid eager
 	// reconciliation that might conflict with independent CRs and apply
 	clearedCfg.CustomRoles = nil
--- a/internal/controller/atlasproject/atlasproject_controller.go
+++ b/internal/controller/atlasproject/atlasproject_controller.go
@@ -110,7 +110,11 @@ func (r *AtlasProjectReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			}
 		}
 
-		// TODO: clear last applied for type
+		if err := r.clearLastAppliedMigratedResources(ctx, atlasProject); err != nil {
+			result = workflow.Terminate(workflow.Internal, err)
+			log.Errorw("Failed to clear migrated independent resources", "error", err)
+			return result.ReconcileResult(), nil
+		}
 
 		return workflow.OK().ReconcileResult(), nil
 	}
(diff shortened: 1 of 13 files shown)


josvazg:
Last test cases added uncovered some inconsistencies in non greedy behaviour on IP access lists, expected to be more greedy when active, and network peering, expected to be less greedy when active.

internal/controller/atlasproject/atlasproject_controller.go
```diff
@@ -348,18 +347,24 @@ func logIfWarning(ctx *workflow.Context, result workflow.Result) {
...
+	lastCfg.CustomRoles = nil
+	lastCfg.PrivateEndpoints = nil
+	lastCfg.ProjectIPAccessList = nil
+	lastCfg.NetworkPeers = nil
+	shallowCopy := akov2.AtlasProject{
+		TypeMeta:   atlasProject.TypeMeta,
+		ObjectMeta: atlasProject.ObjectMeta,
+		Spec:       *lastCfg,
+	}
+	if err := customresource.ApplyLastConfigApplied(ctx, &shallowCopy, r.Client); err != nil {
```
helderjs, line 366: You started using the full resource instead of only the spec as it id currently. This will break existing resources.

helderjs (changes requested):

internal/controller/atlasproject/atlasproject_controller.go
josvazg: Note this `ApplyLastConfigApplied` was always receiving the full `customresource`, not just the `spec`. Then inside the method only the spec is saved:

```go
func ApplyLastConfigApplied(ctx context.Context, resource api.AtlasCustomResource, k8sClient client.Client) error {
	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(resource)
	if err != nil {
		return err
	}

	js, err := json.Marshal(uObj["spec"])
	if err != nil {
		return err
	}
...
	annotations[AnnotationLastAppliedConfiguration] = string(js)
	resourceCopy.SetAnnotations(annotations)
	err = k8sClient.Patch(ctx, resourceCopy, client.MergeFrom(resource))
	if err != nil {
		return err
	}

	// retains current behavior
	resource.SetAnnotations(resourceCopy.GetAnnotations())
	return nil
}
```

internal/controller/customresource/protection.go
```diff
@@ -79,6 +71,19 @@ func applyLastSpec(ctx context.Context, resource api.AtlasCustomResource, k8sCli
 	return nil
 }
 
+func ParseLastConfigApplied[S any](spec *S, resource api.AtlasCustomResource) (*S, error) {
```
s-urbaniak, line 74: **blocking**: this is missing a unit test

internal/controller/atlasproject/atlasproject_controller.go
```diff
@@ -348,18 +347,24 @@ func logIfWarning(ctx *workflow.Context, result workflow.Result) {
...
-func lastSpecFrom(atlasProject *akov2.AtlasProject, annotation string) (*akov2.AtlasProjectSpec, error) {
-	var lastApplied akov2.AtlasProject
-	ann, ok := atlasProject.GetAnnotations()[annotation]
-
-	if !ok {
-		return nil, nil
-	}
-
-	err := json.Unmarshal([]byte(ann), &lastApplied.Spec)
+func (r *AtlasProjectReconciler) clearLastAppliedMigratedResources(ctx context.Context, atlasProject *akov2.AtlasProject) error {
```
s-urbaniak, line 350: **blocking** i really would like to understand the semantics here before I can approve this pull request.

internal/controller/atlasproject/atlasproject_controller.go
josvazg: Pointed out by @s-urbaniak I will be revamping this invocation, we cannot use `ApplyLastConfigApplied` for this as it touches more things than just the annotation. For the new method I will pass the spec alone most likely.

internal/controller/atlasproject/atlasproject_controller.go
josvazg: Clarified offline, this call needs to avoid using `ApplyLastConfigApplied` because that method does more than just applying a change to the last applied annotation.

Instead I should:

1. Extract the last applied state from the current object, just as done now.
2. Clear the fields we need to clear. Still no changes.
3. Avoid any copies (shallow or deep) and just pass the new spec to be applied as a new last applied annotation to a new method called `PatchLastApplied` which will **only** patch the last applied annotation and nothing else.

internal/controller/customresource/protection.go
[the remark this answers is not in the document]
josvazg: coming on next update

internal/controller/atlasproject/atlasproject_controller.go
josvazg: Done, there is ad-hoc `customresource` method that just applies the annotation change alone. No changes to the spec neither for the etcd side nor locally.

internal/controller/customresource/protection.go
```diff
@@ -79,6 +71,52 @@ func applyLastSpec(ctx context.Context, resource api.AtlasCustomResource, k8sCli
...
+func ParseLastConfigApplied[S any](resource api.AtlasCustomResource) (*S, error) {
+	var spec S
+	lastAppliedJSON, ok := resource.GetAnnotations()[AnnotationLastAppliedConfiguration]
+	if !ok {
+		return nil, nil
+	}
+
+	err := json.Unmarshal([]byte(lastAppliedJSON), &spec)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing JSON annotation value [%s] into a %T: %w", lastAppliedJSON, spec, err)
```
s-urbaniak, line 115: **blocking** I am sceptical we should add the whole JSON to the error. If parsing last applied fails, the whole JSON will be added to the AtlasProject status, blowing its size. We can add a dedicated (structured!) log output in this place if this information is useful but if this error happens, the resource can be introspected in kube in the first place.

internal/controller/customresource/protection.go
```diff
@@ -79,6 +71,52 @@ func applyLastSpec(ctx context.Context, resource api.AtlasCustomResource, k8sCli
...
+
+	err = k8sClient.Patch(ctx, copy, client.MergeFrom(resource))
+	if err != nil {
+		return fmt.Errorf("failed to patch resource: %w", err)
+	}
+	return nil
+}
+
+func setLastAppliedConfigAnnotation[S any](resource api.AtlasCustomResource, spec *S) (api.AtlasCustomResource, error) {
+	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(spec)
```
s-urbaniak, line 88: **blocking** why are we using unstructured here? This is a lot of code solving an extremely simple problem, patching the resource.

internal/controller/customresource/protection.go
```diff
@@ -79,6 +71,52 @@ func applyLastSpec(ctx context.Context, resource api.AtlasCustomResource, k8sCli
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get unstructured object from spec %v: %w", spec, err)
+	}
+	js, err := json.Marshal(uObj)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal %v: %w", uObj, err)
+	}
+	copy := resource.DeepCopyObject().(api.AtlasCustomResource)
```
s-urbaniak, line 96: **blocking** this collides with the `copy` builtin, please use another name.

internal/controller/customresource/protection.go
s-urbaniak: converting to unstructured is not necessary, you can marshal `spec` directly.

s-urbaniak:
```
func setLastAppliedConfigAnnotation[S any](resource api.AtlasCustomResource, spec *S) (api.AtlasCustomResource, error) {
	if spec == nil {
		return nil, fmt.Errorf("spec is nil")
	}
	js, err := json.Marshal(spec)
	if err != nil {
		return nil, fmt.Errorf("failed to marshal %v: %w", spec, err)
	}
	copy := resource.DeepCopyObject().(api.AtlasCustomResource)
	annotations := copy.GetAnnotations()
	if annotations == nil {
		annotations = map[string]string{}
	}
	annotations[AnnotationLastAppliedConfiguration] = string(js)
	copy.SetAnnotations(annotations)
	return copy, nil
}
```

s-urbaniak: another suggestion for readability: inline `setLastAppliedConfigAnnotation` in the caller. This method is not used anywhere else and hides the implementation details of making a deep copy of the resource being passed.

This is very serial and easy code, there is no necessity to add additional abstractions.

internal/controller/customresource/protection.go
```diff
@@ -79,6 +71,52 @@ func applyLastSpec(ctx context.Context, resource api.AtlasCustomResource, k8sCli
...
+	annotations := copy.GetAnnotations()
+	if annotations == nil {
```
s-urbaniak, line 98: the nil check is not necessary, every resource in kube has annotations.

s-urbaniak: This whole method can be as simple as:
```
func PatchLastConfigApplied[S any](ctx context.Context, k8sClient client.Client, resource api.AtlasCustomResource, spec *S) error {
	if spec == nil {
		return fmt.Errorf("spec is nil")
	}

	js, err := json.Marshal(spec)
	if err != nil {
		return fmt.Errorf("failed to marshal %v: %w", spec, err)
	}

	resourceCopy := resource.DeepCopyObject().(api.AtlasCustomResource)
	resourceCopy.GetAnnotations()[AnnotationLastAppliedConfiguration] = string(js)

	err = k8sClient.Patch(ctx, resourceCopy, client.MergeFrom(resource))
	if err != nil {
		return fmt.Errorf("failed to patch resource: %w", err)
	}
	return nil
}
```

s-urbaniak (changes requested):
a couple of suggestions in the `PatchLastConfigApplied` method.

internal/controller/customresource/protection.go
s-urbaniak: actually, here, a little helper method acutally makes sense if we want to preserve nil-safety:

s-urbaniak: actually, the current method in `main` is perfectly fine, the unstructured stuff can be removed: https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/internal/controller/customresource/protection.go#L51-L80

internal/controller/customresource/protection.go
s-urbaniak: actually, the current method in `main` is perfectly fine, the unstructured stuff can be removed: https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/internal/controller/customresource/protection.go#L51-L80

s-urbaniak:
```
func PatchLastConfigApplied[S any](ctx context.Context, k8sClient client.Client, resource api.AtlasCustomResource, spec *S) error {
	if spec == nil {
		return fmt.Errorf("spec is nil")
	}

	js, err := json.Marshal(spec)
	if err != nil {
		return fmt.Errorf("failed to marshal %v: %w", spec, err)
	}

	resourceCopy := resource.DeepCopyObject().(api.AtlasCustomResource)
	annotations := resourceCopy.GetAnnotations()
	if annotations == nil {
		annotations = map[string]string{}
	}
	annotations[AnnotationLastAppliedConfiguration] = string(js)
	resourceCopy.SetAnnotations(annotations)

	err = k8sClient.Patch(ctx, resourceCopy, client.MergeFrom(resource))
	if err != nil {
		return fmt.Errorf("failed to patch resource: %w", err)
	}
	return nil
}
```

s-urbaniak (approved):
looks great now, great job! 🎉

helderjs (approved):

internal/controller/customresource/protection.go
```diff
@@ -79,6 +71,45 @@ func applyLastSpec(ctx context.Context, resource api.AtlasCustomResource, k8sCli
...
+		return fmt.Errorf("failed to patch resource: %w", err)
+	}
+	return nil
+}
+
+func ParseLastConfigApplied[S any](resource api.AtlasCustomResource) (*S, error) {
+	var spec S
+	lastAppliedJSON, ok := resource.GetAnnotations()[AnnotationLastAppliedConfiguration]
+	if !ok {
+		return nil, nil
```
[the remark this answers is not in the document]
s-urbaniak: concretely at https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2192/files#diff-0d383863ebbfca7afb47ff12d141447ce823fbcd48870db9722d7b5a53cb3e15R375

s-urbaniak (changes requested):
one potential nil panic

internal/controller/customresource/protection.go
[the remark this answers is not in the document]
josvazg: fixed: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2192/commits/bbb4d407e19da3a04821b1c213d857301281c2ed

s-urbaniak (approved):

roothorp (approved):

PR #2199, opened 2025-03-17
CLOUDP-306040: Manually upgrade to controller-runtime 0.20

josvazg:
### All Submissions:


internal/controller/atlasdatafederation/datafederation_controller_test.go
```diff
@@ -237,7 +237,11 @@ func TestDeleteConnectionSecrets(t *testing.T) {
 				gotSecrets.Items[i].ResourceVersion = ""
 			}
 			assert.NoError(t, err)
-			assert.Equal(t, tc.wantSecrets, gotSecrets.Items)
+			wantSecrets := tc.wantSecrets
+			if wantSecrets == nil {
```
helderjs, line 241: Test case should set `nil` instead of have logic around it.

helderjs (changes requested):

internal/controller/atlasdatafederation/datafederation_controller_test.go
josvazg: Tried to do that, but the linter was complaining about duplicated code between test cases... I can give it another go.

josvazg: Hey, turns out the test case was really duplicated!gt

internal/controller/atlasdatafederation/datafederation_controller_test.go
```diff
@@ -91,34 +93,8 @@ func TestDeleteConnectionSecrets(t *testing.T) {
 				},
 				Spec: akov2.DataFederationSpec{Project: common.ResourceRefNamespaced{Name: "fooProject"}},
 			},
-			wantResult: workflow.OK(),
-		},
-		{
-			name: "federation object without secrets",
```
josvazg, line 97: This test case was duplicated.

PR #2211, opened 2025-03-18
CLOUDP-304211: Only accept pull_request_target with safe-to-test label

josvazg:
This PR uses a new small `trigger.yml` workflow to handle the logic for both `pull_request` & `pull_request_target` separately of other ways to trigger the `test.yml` workflow. There are only 2 scenarios allowed:

1. The `pull_request` events without the `safe-to-test` label, intended for code owner contributions. Uses the incoming change repo creds, which is the same, as the PR comes from a branch in the same repo.
2. The `pull_request_target` events with `safe-to-test` label that a code owner have just set. Uses the target repo creds, that is the AKO repo creds, and is intended to allow contributions to be tested once a code owner decides it is safe to do so.

### All Submissions:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,24 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  trace:
+    name: Traces
+    runs-on: ubuntu-latest
+    steps:
+    - name: Show
+      id: show
+      env:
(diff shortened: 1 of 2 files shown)


josvazg:
The only way to activate this change and test it is by merging it first, unfortunately. This is because the `pull_request_target` we need to test runs on the base branch (`main`) context, and will not have the new trigger logic until it is merged in.

josvazg:
CLOUDP-304211

s-urbaniak (approved):

PR #2240, opened 2025-03-31
CLOUDP-296897: Allow to select tests to run

igor-karpukhin:
# Summary

This PR introduces a new workflow that allows to selectively run tests based on Github labels
Examples:
* `test/e2e/*` - runs all e2e tests
* `test/e2e/Atlas*` - runs e2e tests whose Ginkgo test labels starts with `Atlas`
* `test/e2e/AtlasBackupSchedule` - runs specific e2e test with Ginkgo label `AtlasBackupSchedule`
* `test/int/*` - runs all integration
* `test/int/alert-*` - runs integration tests whose Ginkgo test labels starts with `alert-`
* `test/int/alert-config` - runs specific integration test with Ginkgo label `alert-config`

The workflow is triggered for opened, labeled and modified PR. The workflow is triggered for every labels change.

This PR doesn't break existing workflow. The previous workflow can still be triggered by applying `cloud-tests` label.

## Proof of Work

See the workflow results

## Checklist


.github/workflows/tests-selectable.yaml
```diff
@@ -0,0 +1,372 @@
+name: Tests-selectable
+
+# This a prototype for running tests based on PR labels
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled]
```
josvazg, line 7: **comment:** Note this will never work for external contributors. For security reasons, the `pull_request` event uses the credentials from the contributor's fork repo, not the AKO repo. For accepting external contributions `pull_request_target` needs to be used instead, but in such case, when the contribution is external, a gating mechanism needs to be set in place to avoid external forks to access and peek or abuse AKO creds without us even been able to stop it (for instance, if external, the safe-to-test label must also be present).

BTW, one current issue I had when tying to fix these external workflows is that if you have `pull_request` & `pull_request_target` both set, they will trigger and your workflows might duplicate. I guess the best way is, either make sure only one can fire at a time (`pull_request` only for owners and `pull_request_target` only for contributors) or use only `pull_request_target` with gating for non owners.

**q:** Was intended to skip external contributions? It is fine if not as the CI is still broken in general for that use case.

.github/workflows/tests-selectable.yaml
```diff
@@ -0,0 +1,372 @@
...
+        run: go install github.com/onsi/ginkgo/v2/ginkgo@latest
+
+      - name: Get test labels from PR or input
+        env:
+          TEST_LABELS: ${{ github.event.inputs.testLabels }}
+        id: get-labels
+        uses: actions/github-script@v7
+        with:
+          script: |
+            if (context.eventName === 'pull_request') {
```
josvazg, line 52: **comment:** remember this if `pull_request_target` is added in the PR.

.github/workflows/tests-selectable.yaml
```diff
@@ -0,0 +1,372 @@
...
+          E2E_LABELS: ${{ env.e2e_labels }}
+          USE_JSON: true
+        run: |
+          make compute-labels
+          ./bin/ginkgo-labels > result.json
+          echo "Int tests to execute $(cat result.json | jq -c .int)"
+          echo "E2E tests to execute $(cat result.json | jq -c .e2e)"
+          
+          echo "int_matrix=$(cat result.json | jq -c .int)" >> $GITHUB_OUTPUT
+          echo "e2e_matrix=$(cat result.json | jq -c .e2e)" >> $GITHUB_OUTPUT
```
josvazg, line 92: **comment:** we'll need to take this into account when we split helm chart tests from e2e

.github/workflows/tests-selectable.yaml
```diff
@@ -0,0 +1,372 @@
...
+
+  prepare-e2e:
+    name: Prepare E2E configuration and image
+    if: ${{ fromJSON(needs.detect-tests.outputs.e2e_matrix) != '[]' }}
+    environment: release
+    runs-on: ubuntu-latest
+    env:
+      REPOSITORY: ${{ github.repository_owner }}/mongodb-atlas-kubernetes-operator-prerelease
+    steps:
+      - if: ${{ inputs.forked == false }}
```
josvazg, line 119: **q:** given we do not accept external contribs, why is this here?

.github/workflows/tests-selectable.yaml
```diff
@@ -0,0 +1,372 @@
...
+          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+          AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          GCP_SA_CRED: ${{ secrets.GCP_SA_CRED }}
+          DATADOG_KEY: ${{ secrets.DATADOG_KEY }}
+          PAGER_DUTY_SERVICE_KEY: ${{ secrets.PAGER_DUTY_SERVICE_KEY }}
+        run: devbox run -- ./scripts/launch-ci-e2e.sh
+
+#        run: |
+#          cd ./test/e2e
+#          ginkgo --label-filter="${{ matrix.test }}" -r --fail-fast
```
josvazg, line 372: **q:** interesting this `--fail-fast`, should we be more explicit about this? I mean this means when the PR fails there might be more errors to check on the same label that failed in the CI
But I agree we probably want this to be less wasteful with CI tests.

Makefile
```diff
@@ -151,6 +151,12 @@ help: ## Show this help screen
 .PHONY: all
 all: manager ## Build all binaries
 
+
+.PHONY: compute-labels
+compute-labels:
+	mkdir -p bin
+	go build -o bin/ginkgo-labels tools/compute-test-labels/main.go
```
josvazg, line 158: **q:** Not for this PR, but a thought in general. Should we group all AKO dev tools in a single tool repo?

tools/compute-test-labels/main.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+)
+
+func jsonDump(data interface{}) string {
+	r, _ := json.Marshal(data)
+	return string(r)
+}
```
josvazg, line 14: **comment:** I use this exact same code a lot when debugging, we might want to have it in a library somewhere in internal. It will be a different copy here, though.

josvazg (commented):
Waiting on a few questions, but LGTM overall.
Great job!

.github/workflows/tests-selectable.yaml
igor-karpukhin: Ok, I think we can drop the CI runs for PRs from forked repos for now. I'll create a follow up ticket

Makefile
igor-karpukhin: We can think about that if there are good reasons for that. So far, nothing stops us for having tools in the `/tools` folder.

.github/workflows/tests-selectable.yaml
igor-karpukhin: This is fixed by executing the devbox

.github/workflows/tests-selectable.yaml
```diff
@@ -0,0 +1,345 @@
...
+      - name: Compute Test Matrix
+        id: set-matrix
+        env:
+          PR_LABELS: ${{ steps.get-labels.outputs.result }}
+          INT_LABELS: ${{ env.int_labels }}
+          E2E_LABELS: ${{ env.e2e_labels }}
+          USE_JSON: true
+        run: |
+          make compute-labels
+          ./bin/ginkgo-labels > result.json
```
s-urbaniak, line 82: **q** why not just `go run ....`?

s-urbaniak (approved):

.github/workflows/tests-selectable.yaml
igor-karpukhin: Also possible, but I decided to go with the `go build` option

josvazg (approved):
LGTM! great job!

helderjs (approved):

PR #2313, opened 2025-04-30
CLOUDP-316083: Integration Controller

josvazg:
# Summary

Internal-only implementation for the integrations controller

## Proof of Work

TBD including new tests

## Checklist

  * **q**: Clarifying question.


internal/controller/atlasthirdpartyintegrations/setup.go
```diff
@@ -0,0 +1,196 @@
...
+					mckpredicate.AnnotationChanged("mongodb.com/reapply-period"),
+					predicate.GenerationChangedPredicate{},
+				),
+				mckpredicate.IgnoreDeletedPredicate[client.Object](),
+			),
+		).
+		Watches(
+			&akov2.AtlasProject{},
+			handler.EnqueueRequestsFromMapFunc(h.integrationForProjectMapFunc()),
+			builder.WithPredicates(integrationPredicates),
```
s-urbaniak, line 102: we want to watch for generation changes only here.

s-urbaniak, line 102: the global variable `integrationPredicates` obfuscates what predicates we're having so reading through it makes it hard to reason about what predicates are actually applied. I suggest using `predicate.GenerationChangedPredicate{}` and `predicate.ResourceVersionChangedPredicate{}` to make it explicit and readable to the user.

In addition you risk faulty behaviour in case controller-runtime decides to have internal state inside the predicate struct. I'd be careful reusing it.

internal/controller/atlasthirdpartyintegrations/setup.go
```diff
@@ -0,0 +1,196 @@
...
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(h.integrationForCredentialMapFunc()),
+			builder.WithPredicates(integrationPredicates),
+		).
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(h.integrationForSecretMapFunc()),
+			builder.WithPredicates(integrationPredicates),
+		).
```
s-urbaniak, line 113: really just a nit: rather than watching twice for `Secret` i'd suggest using `predicate.Or`, it makes it more explicit.

internal/controller/atlasthirdpartyintegrations/handler.go
```diff
@@ -0,0 +1,199 @@
...
+			currentState,
+			fmt.Errorf("Error evaluating secret changes for %s Atlas Integration for project %s: %w",
+				integrationSpec.Type, req.Project.ID, err),
+		)
+	}
+	if secretChanged || !reflect.DeepEqual(atlas, spec) {
+		return h.update(ctx, currentState, req, integrationSpec)
+	}
+	return result.NextState(
+		state.StateCreated,
```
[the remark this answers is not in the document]
s-urbaniak: alternative: pass a `nextState` var as well, callers would be then:
```
func (h *AtlasThirdPartyIntegrationHandler) HandleInitial(ctx context.Context, integration *akov2next.AtlasThirdPartyIntegration) (ctrlstate.Result, error) {
	return h.upsert(ctx, state.StateInitial, state.StateCreated, integration)
}

func (h *AtlasThirdPartyIntegrationHandler) HandleCreated(ctx context.Context, integration *akov2next.AtlasThirdPartyIntegration) (ctrlstate.Result, error) {
	return h.upsert(ctx, state.StateCreated, state.StateUpdated, integration)
}

func (h *AtlasThirdPartyIntegrationHandler) HandleUpdated(ctx context.Context, integration *akov2next.AtlasThirdPartyIntegration) (ctrlstate.Result, error) {
	return h.upsert(ctx, state.StateUpdated, state.StateUpdated, integration)
}
```

internal/controller/atlasthirdpartyintegrations/handler.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (h *AtlasThirdPartyIntegrationHandler) delete(ctx context.Context, req *reconcileRequest, integrationType string) (ctrlstate.Result, error) {
+	err := req.Service.Delete(ctx, req.Project.ID, integrationType)
+	if errors.Is(err, thirdpartyintegration.ErrNotFound) {
+		return h.unmanage(integrationType)
+	}
+	if err != nil {
+		return result.Error(
+			state.StateInitial,
```
s-urbaniak, line 150:
```suggestion
			state.StateDeletionRequested
```

s-urbaniak (commented):
unified state machine for the win 👏 this looks great!

internal/controller/atlasthirdpartyintegrations/setup.go
josvazg: Those secrets are semantically different. One if for operator credentials, and the other are for integration credentials. The indexers are totally different, even if they map secret<-> integration.

I will try to combine, but it might not be that easy.

internal/controller/atlasthirdpartyintegrations/handler.go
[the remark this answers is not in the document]
josvazg: I think I go for the simpler static fix.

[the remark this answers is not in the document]
josvazg: Actually, the next step is best, as if it was already created and it did not change, we want it to stay saying created.

test/helper/e2e2/operator/operator.go
```diff
@@ -108,13 +143,38 @@ func (o *Operator) Stop(t testingT) {
...
+		t.Errorf("error stopping operator terminated=%v : %+#v", terminated, err)
+	}
+}
+
+func operatorBinary() string {
+	return envVarOrDefault("AKO_BINARY", filepath.Join("bin", "manager"))
+}
+
+func delveListenFlag() string {
+	return fmt.Sprintf("--listen=%s", envVarOrDefault("DELVE_LISTEN", DefaultDelveListen))
```
s-urbaniak: inline this function, it provides no value other than the reader scrolling down

test/helper/e2e2/operator/operator.go
```diff
@@ -108,13 +143,38 @@ func (o *Operator) Stop(t testingT) {
...
+					return // ignore sigterm if we sent SIGTERM ourselves
+				}
+			}
+		}
+		t.Errorf("error stopping operator terminated=%v : %+#v", terminated, err)
+	}
+}
+
+func operatorBinary() string {
+	return envVarOrDefault("AKO_BINARY", filepath.Join("bin", "manager"))
```
s-urbaniak: inline this function, it provides no value other than the reader scrolling down

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+			return fmt.Errorf("failed to create %s: %w", key, err)
+		}
+	default:
+		return fmt.Errorf("failed to apply %s: %w", key, err)
+	}
+	return nil
+}
+
+func CreateNamespace(ctx context.Context, kubeClient client.Client, namespace string) error {
+	return kubeClient.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
```
s-urbaniak: inline this function, it provides no value other than the reader must introspect a different package.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func WipeNamespace(ctx context.Context, kubeClient client.Client, namespace string) error {
+	return kubeClient.Delete(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
```
s-urbaniak: inline this function, it provides no value other than the reader must introspect a different package.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func HasNamespace(ctx context.Context, kubeClient client.Client, namespace string) bool {
+	return kubeClient.Get(ctx, client.ObjectKey{Name: namespace}, &corev1.Namespace{}) == nil
```
s-urbaniak: inline this function, it provides no value other than the reader must introspect a different package.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func apply(ctx context.Context, kubeClient client.Client, obj client.Object) error {
+	key := client.ObjectKeyFromObject(obj)
+	old := obj.DeepCopyObject().(client.Object)
+	err := kubeClient.Get(ctx, key, old)
+	switch {
+	case err == nil:
+		obj = obj.DeepCopyObject().(client.Object)
+		obj.SetResourceVersion(old.GetResourceVersion())
```
s-urbaniak: don't mutate resource version, its a no-no-no-go.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+		apiextensionsv1.AddToScheme,
+		akov2.AddToScheme,
+		akov2next.AddToScheme)
+	if err != nil {
+		return nil, fmt.Errorf("failed to setup Kubernetes test env scheme: %w", err)
+	}
+	return getKubeClient(testScheme)
+}
+
+func Apply(ctx context.Context, kubeClient client.Client, objs ...client.Object) error {
```
s-urbaniak: i am questioning the usefulness of this function. it has no retries, it is not clear whether it is a create or update. did you take this from kubectl code?

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+
+func apply(ctx context.Context, kubeClient client.Client, obj client.Object) error {
+	key := client.ObjectKeyFromObject(obj)
+	old := obj.DeepCopyObject().(client.Object)
+	err := kubeClient.Get(ctx, key, old)
+	switch {
+	case err == nil:
+		obj = obj.DeepCopyObject().(client.Object)
+		obj.SetResourceVersion(old.GetResourceVersion())
+		if err := kubeClient.Update(ctx, obj); err != nil {
```
s-urbaniak: this is missing retries on conflict.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+	err := kubeClient.Get(ctx, key, obj)
+	if err != nil {
+		return false, fmt.Errorf("failed to get object %v: %w", key, err)
+	}
+	return true, nil
+}
+
+func SetNamespace(obj client.Object, ns string) client.Object {
+	renamed := obj.DeepCopyObject().(client.Object)
+	renamed.SetNamespace(ns)
```
s-urbaniak: what value does this function provide? the function indicates you are mutating an object, however you are returning a copy. I would inline this so it is explicit to the reader what is happening.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	for _, condition := range obj.GetConditions() {
+		if condition.Type == "Ready" && condition.Status == metav1.ConditionTrue {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func AssertObjExists(ctx context.Context, kubeClient client.Client, key client.ObjectKey, obj client.Object) (bool, error) {
```
s-urbaniak: i don't think this function is useful. The name implies it is a read-only function but it in fact also mutates the given `obj` parameter.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+func HasNamespace(ctx context.Context, kubeClient client.Client, namespace string) bool {
+	return kubeClient.Get(ctx, client.ObjectKey{Name: namespace}, &corev1.Namespace{}) == nil
+}
+
+func AssertObjReady(ctx context.Context, kubeClient client.Client, key client.ObjectKey, obj ObjectWithStatus) (bool, error) {
+	err := kubeClient.Get(ctx, key, obj)
+	if err != nil {
+		return false, fmt.Errorf("failed to get object %v: %w", key, err)
+	}
+	for _, condition := range obj.GetConditions() {
```
s-urbaniak: instead of iterating, you can use a metav1 helper function, we can use apimachinery helpers for that.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+
+func WipeNamespace(ctx context.Context, kubeClient client.Client, namespace string) error {
+	return kubeClient.Delete(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
+}
+
+func HasNamespace(ctx context.Context, kubeClient client.Client, namespace string) bool {
+	return kubeClient.Get(ctx, client.ObjectKey{Name: namespace}, &corev1.Namespace{}) == nil
+}
+
+func AssertObjReady(ctx context.Context, kubeClient client.Client, key client.ObjectKey, obj ObjectWithStatus) (bool, error) {
```
s-urbaniak: i am doubting the usefulness of this function, it mutates the object parameter rather than just asserting. just inline this at the call side and use `meta.FindStatusCondition` in the apimachinery library.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, fmt.Errorf("failed to get Kubernetes config (is cluster configured?): %w", err)
+	}
+	kubeClient, err := client.New(restCfg, client.Options{Scheme: scheme})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Kubernetes client (is cluster up?): %w", err)
+	}
+	return kubeClient, nil
+}
+
+func assertCRD(ctx context.Context, kubeClient client.Client, targetCRD string) error {
```
s-urbaniak: what is `targetCRD`? is it a GVK, a GVR, something else? I suggest passing the concrete type.

test/helper/e2e2/kube/kube.go
```diff
@@ -0,0 +1,178 @@
...
+	return true, nil
+}
+
+func SetNamespace(obj client.Object, ns string) client.Object {
+	renamed := obj.DeepCopyObject().(client.Object)
+	renamed.SetNamespace(ns)
+	return renamed
+}
+
+func getTestScheme(addToSchemeFunctions ...func(*runtime.Scheme) error) (*runtime.Scheme, error) {
```
s-urbaniak: `s/getTestScheme/newScheme`

this is not getting but creating a new scheme

test/helper/e2e2/operator/operator.go
```diff
@@ -25,6 +25,10 @@ import (
 	"syscall"
 )
 
+const (
+	DefaultDelveListen = ":2345"
```
s-urbaniak: `DefaultDelveListenPort`

test/helper/e2e2/kube/kube.go
josvazg: Its for ergonomy, it does what it says on the tin but you do not need to wrap keys etc.

still, will do

test/helper/e2e2/yml/yml.go
```diff
@@ -0,0 +1,159 @@
...
+
+func MustParseCRs(ymls io.Reader) []client.Object {
+	objs, err := ParseCRs(ymls)
+	if err != nil {
+		panic(fmt.Errorf("Fatal: could not parse CRs: %w", err))
+	}
+	return objs
+}
+
+func ParseCRs(ymls io.Reader) ([]client.Object, error) {
```
s-urbaniak: s/ParseCRs/ParseObject

"CR" is not a thing in kube source code lingua. You are parsing a client object, hence let's call the function appropriately.

test/helper/e2e2/yml/yml.go
```diff
@@ -0,0 +1,159 @@
...
+		obj, err := DecodeCR(buffer.Bytes())
+		if err != nil && !errors.Is(err, ErrNoCR) {
+			return nil, err
+		}
+		objs = append(objs, obj)
+	}
+	return objs, nil
+}
+
+func DecodeCR(content []byte) (client.Object, error) {
```
s-urbaniak: see above

test/helper/e2e2/kube/kube.go
josvazg: how to support update then?

test/helper/e2e2/kube/kube.go
josvazg: nope, but I wanted to mimic kubectl apply behaviour

test/helper/e2e2/kube/kube.go
josvazg: I see, we did have a helper for that already right?

test/helper/e2e2/kube/kube.go
josvazg: What about renaming to `CloneWithNamespace` ? that removes ambiguity.

The reason I do not want to inline this is because this is a widespread helper service all tests would need:
- Test objects will NOT know the namespace until runtime, as it is a random name.
- Test objects should also pick a random project name to avoid atlas collisions.
- The support should be seamless whether you chose to use a go struct for your client.Object or it is coming from a YAML.

test/helper/e2e2/kube/kube.go
josvazg: the mutation can be removing making a clone inside.

The thing is I do not want the test to become less readable and us having to repeat the same constructs in all tests all the same.

test/helper/e2e2/kube/kube.go
josvazg: ergonomics again, there must be some middle ground to having to repeat yourself every time on every test

test/helper/e2e2/kube/kube.go
josvazg: will check

test/helper/e2e2/kube/kube.go
s-urbaniak: as discussed: use a one-liner using server side apply and preach the knowledge.

test/e2e2/e2e2_suite_test.go
```diff
@@ -0,0 +1,78 @@
...
+	logger := zaptest.NewLogger(t)
+	logrLogger := zapr.NewLogger(logger)
+	ctrllog.SetLogger(logrLogger.WithName("test"))
+}
+
+func runTestAKO(ns string, deletionprotection bool) *operator.Operator {
+	args := []string{
+		"--log-level=-9",
+		"--global-api-secret-name=mongodb-atlas-operator-api-key",
+		`--atlas-domain=https://cloud-qa.mongodb.com`,
```
s-urbaniak, line 83: let's use the json encoder so we are fit for structured logging integration

s-urbaniak: --log-encoder=json

test/helper/e2e2/kube/kube.go
josvazg: will remove the whole apply helper and instead use server side apply patching

test/helper/e2e2/kube/kube.go
josvazg: server side apply coming up to replace this

test/helper/e2e2/kube/kube.go
josvazg: will remove this one and instead make it composable. Eg:
```go
fixedObj := RandomizeAtlasProjectName(ReplaceNamespace(obj, ns))
```

test/helper/e2e2/kube/kube.go
josvazg: will remove and replace by regular controller-runtime functions, see how it looks like inline afterwards

test/helper/e2e2/kube/kube.go
josvazg: will remove and replace by regular controller-runtime functions, see how it looks like inline afterwards

test/helper/e2e2/kube/kube.go
josvazg: will use the type instead of the name

internal/controller/atlasthirdpartyintegrations/handler.go
```diff
@@ -0,0 +1,179 @@
...
+const (
+	AnnotationContentHash = "mongodb.com/content-hash"
+)
+
+func (h *AtlasThirdPartyIntegrationHandler) HandleInitial(ctx context.Context, integration *akov2next.AtlasThirdPartyIntegration) (ctrlstate.Result, error) {
+	return h.upsert(ctx, state.StateInitial, state.StateCreated, integration)
+}
+
+func (h *AtlasThirdPartyIntegrationHandler) HandleCreated(ctx context.Context, integration *akov2next.AtlasThirdPartyIntegration) (ctrlstate.Result, error) {
+	return h.upsert(ctx, state.StateCreated, state.StateCreated, integration)
```
josvazg: created -> updated

internal/controller/atlasthirdpartyintegrations/secret.go
```diff
@@ -0,0 +1,143 @@
...
+	secretName, err := secretName(integration) // at this point the secret should have filed at populateIntegration
+	if err != nil {
+		return false, fmt.Errorf("failed to check for secret changes: %w", err)
+	}
+	secret, err := fetchSecret(ctx, h.Client, secretName, integration.Namespace)
+	if err != nil {
+		return false, fmt.Errorf("failed to retrieve secret %s to evaluate changes: %w", secretName, err)
+	}
+	currentValue := hashSecret(secret.Data)
+	for k, v := range secret.Annotations {
```
s-urbaniak:
```suggestion
	if v, ok := secret.GetAnnotations()[AnnotationContentHash]; ok {
		if v == currentValue {
			return false, nil
		}
	}
```

s-urbaniak (approved):
looking great, just nits

roothorp (approved):

PR #2359, opened 2025-05-30
CLOUDP-316083: Add Unified State Machine support libraries

josvazg:
# Summary

Add support for the Unified State Machine. 

## Proof of Work

This has been tested in #2313 

## Checklist

### DIFF
--- a/pkg/controller/state/reconciler.go
+++ b/pkg/controller/state/reconciler.go
@@ -1,0 +1,369 @@
+// Copyright 2025 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package state
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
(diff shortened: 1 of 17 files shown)


pkg/controller/state/reconciler.go
```diff
@@ -0,0 +1,369 @@
...
+func (r *Reconciler[T]) Reconcile(ctx context.Context, req ctrl.Request) (reconcile.Result, error) {
+	logger := log.FromContext(ctx).WithName("state")
+	logger.Info("reconcile started", "req", req)
+
+	t := new(T)
+	obj := any(t).(StatusObject)
+	clientObj := any(t).(client.Object)
+	if u, ok := clientObj.(*unstructured.Unstructured); ok {
+		u.SetGroupVersionKind(r.unstructuredGVK)
+	}
```
helderjs, line 136: q: Why do you need this check?

josvazg: This came straight from the PoC using unstructured CRDs, I believe we do not need it for our controllers, but...
Qe might still test with unstructured data sometime?

I will let @s-urbaniak chime in to confirm whether or not we might need this going forward, even if not for production code.

s-urbaniak: this is needed if we are reconciling unstructured. For unstructured objects we must set the GVK otherwise the underlying shared index informer won't be able to find it in the cache, it won't know what type it is.

josvazg: Yet the question remains, are we going to use unstructured at all? when?

I suspect we might use it for some experiments on the side. This being expected to be a library I guess it merits the flexibility to support unstructured. Right?

s-urbaniak (approved):

pkg/controller/state/reconciler.go
s-urbaniak: i think it is fine to leave the support for unstructured, mostly for proof of concepts.

PR #2360, opened 2025-05-30
CLOUDP-221656: Added an openshift test cleaner to the regular cleaner job

igor-karpukhin:
# Summary

Added an openshift test cleaner to the regular cleaner job

## Checklist

### DIFF
--- a/.github/workflows/cleanup-test.yml
+++ b/.github/workflows/cleanup-test.yml
@@ -48,12 +48,6 @@ jobs:
 
           echo credentials=$(realpath ~/gcp_sa_cred.json) >> $GITHUB_OUTPUT
 
-      - name: Download tools for openshift test cleaner
-        run: |
-          wget https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest-4.6/openshift-client-linux.tar.gz -O openshift.tar.gz -q
-          tar xvf openshift.tar.gz
-          chmod +x oc && sudo mv oc /usr/local/bin/oc
-
       - name: Run cleaner
         env:
           MCLI_OPS_MANAGER_URL: 'https://cloud-qa.mongodb.com/'
--- a/.github/workflows/cleanup-test.yml
+++ b/.github/workflows/cleanup-test.yml
@@ -48,6 +48,12 @@ jobs:
 
           echo credentials=$(realpath ~/gcp_sa_cred.json) >> $GITHUB_OUTPUT
 
+      - name: Download tools for openshift test cleaner
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/cleanup-test.yml
```diff
@@ -48,6 +48,12 @@ jobs:
 
           echo credentials=$(realpath ~/gcp_sa_cred.json) >> $GITHUB_OUTPUT
 
+      - name: Download tools for openshift test cleaner
+        run: |
+          wget https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest-4.6/openshift-client-linux.tar.gz -O openshift.tar.gz -q
```
s-urbaniak, line 53: 4.6 is very old, also I am a bit concerned about keeping this up-to-date. Can't we inline this in the script itself?

scripts/openshift-upgrade-test-cleaner.sh
```diff
@@ -0,0 +1,51 @@
...
+  echo "AKO openshift upgrade test cleaner"
+  oc_login
+  echo "Cleaning up the test leftovers..."
+  namespaces=$(oc get namespaces | grep "ako-upgrade-test-" | awk '{print $1}')
+  if [ -z "$namespaces" ]; then
+    echo "No namespaces found for cleanup"
+  else
+    for ns in $namespaces; do
+      echo "Deleting namespace: $ns"
+      oc delete namespace --force "$ns" --wait=true --timeout=60s || true
```
s-urbaniak, line 45: i am thinking, why are we scripting this? can't we write a small go based program to delete the namespaces?

igor-karpukhin: Why? That's way easier with existing tools to just find namespace to delete and delete them, no?

.github/workflows/cleanup-test.yml
igor-karpukhin: This is from the upgrade test, which still works. We use a limited set of OC commands. But we can bump them to the latest one

josvazg: can't we devboxify it?
https://search.nixos.org/packages?channel=25.05&show=openshift&from=0&size=50&sort=relevance&type=packages&query=openshift

igor-karpukhin: I think we can, but this is only needed for this script. You can't launch it locally anyway

scripts/openshift-upgrade-test-cleaner.sh
igor-karpukhin: Ok, I see :) Then let's keep it as is

s-urbaniak (approved):

PR #2376, opened 2025-06-06
CLOUDP-321083: fix reconciliation loop on dbuser

tvdw-mongodb:
# Summary

In 140ba21f32fd5c77c2b070b8851b7a7b23a4b2f7 I added the ability to set a description on AtlasDatabaseUser. However, I overlooked that a description could already exist on the user in Atlas, in which case the operator will fetch it, realize it's different, and then... omit the change, because `omitempty` is set on the struct field.

With this patch we send an empty string instead.

## Checklist

### DIFF
--- a/internal/translation/dbuser/internal_test.go
+++ b/internal/translation/dbuser/internal_test.go
@@ -105,6 +105,34 @@ func TestToAndFromAtlas(t *testing.T) {
 	}
 }
 
+func TestToAtlas(t *testing.T) {
+	for _, tc := range []struct {
+		title     string
+		user      *User
+		atlasUser *admin.CloudDatabaseUser
+	}{
+		{
+			title: "Empty description should be converted as empty string, not nil",
+			user: func() *User {
+				user := defaultTestUser()
+				user.Description = ""
+				return user
+			}(),
+			atlasUser: func() *admin.CloudDatabaseUser {
+				inAtlas := defaultTestAtlasUser()
+				inAtlas.Description = pointer.MakePtr("")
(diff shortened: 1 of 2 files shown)


josvazg:
Thanks for the contribution! waiting on the testing bits to review

s-urbaniak (approved):
lgtm on green 👍

internal/translation/dbuser/internal_test.go
```diff
@@ -143,5 +171,6 @@ func defaultTestAtlasUser() *admin.CloudDatabaseUser {
 		GroupId:      testProjectID,
 		Password:     pointer.MakePtr(testPassword),
 		Username:     testUsername,
+		Scopes:       pointer.MakePtr([]admin.UserScope{}),
```
tvdw-mongodb, line 174: I added these empty scopes so that `toAtlas(defaultTestUser()) == defaultTestAtlasUser()`

internal/translation/dbuser/internal_test.go
```diff
@@ -143,5 +171,7 @@ func defaultTestAtlasUser() *admin.CloudDatabaseUser {
...
+		Description:  pointer.MakePtr(""),
```
tvdw-mongodb, line 175: This one is a bit ironic considering it's also literally the test case I'm adding, but like the comment above, it's so that the test may actually do what it says. The test still explicitly sets the two values it cares about so that if/when these values update, the test will still check the thing we care about

josvazg (approved):

PR #2440, opened 2025-07-10
CLOUDP-329966/promote-images

andrpac:
# Summary

In this step, we would ideally like to move the unofficial image built during the e2e helm-tests to official `prerelease` registries (docker.io, quay.io) if all the tests passed.

Jira-[CLOUDP-329966](https://jira.mongodb.org/browse/CLOUDP-329966)

## Proof of Work

1. Tested on forked AKO on schedule events on subset of e2e tests: https://github.com/andrpac/test-ako/actions/runs/16222019747
2. Tested on AKO by first waiting for test.yml to finish creating the image and then re-running the Promote Image script. 

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -61,7 +61,7 @@ jobs:
           repository: ${{ env.REPO }}
           version: ${{ steps.set_tag.outputs.tag }}
           tags: ${{ steps.set_image_url.outputs.image_url }}
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
           push_to_docker: false
 
   e2e:


.github/workflows/promote-image.yml
```diff
@@ -0,0 +1,83 @@
+name: Promote Image
+
+on:
+  workflow_run:
+    workflows: ["Test"]
+    types: [completed]
```
josvazg, line 6: **q**: Would this not run after any test run completed? It will not discriminate if it is the end of the nightly that run all the tests or any other run.

andrpac: There's an if that checks that the event has to be `schedule`. It is right here: 

`    if: |
      github.event.workflow_run.head_branch == 'main' &&
      github.event.workflow_run.conclusion == 'success' &&
      github.event.workflow_run.event == 'schedule'
`

Any other events that triggered `Test`, will skip the promote step.

josvazg: this means the manual trigger will not promote right?
I think it is fine for now

andrpac: Yes, the manual trigger won't promote for now. This would require a bit more effort since I would have to introduce the `release-nightly` argument on the workflow and do an extra if check. For now, I wanna get the simplest version of the entire pipeline ready.

josvazg: cool

PR #2524, opened 2025-07-16
CLOUDP-329235/releasing-images

andrpac:
# Summary

Current releases rely on cloud-qa e2e and integration tests which heavily depend on availability of resources in cloud-qa. Releases could be done in a more reliable way by using nightly images, which were previously tested and validated. 

Jira-[CLOUDP-329235](https://jira.mongodb.org/browse/CLOUDP-329235)

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/.github/actions/build-push-image/action.yaml
+++ b/.github/actions/build-push-image/action.yaml
@@ -99,3 +99,5 @@ runs:
         sbom: true
         push: true
         tags: ${{ inputs.tags }}
+        labels: |
+          org.opencontainers.image.revision=${{ github.sha }}
--- a/.github/actions/image2commit/action.yml
+++ b/.github/actions/image2commit/action.yml
@@ -2,7 +2,7 @@ name: image2commit
 description: Resolve full commit SHA from a promoted image tag.
 
 inputs:
-  register:
+  registry:
     description: "Registry (e.g., docker.io, quay.io)"
     required: true
   repo:
@@ -33,7 +33,7 @@ runs:
         full_sha=$(${{
           github.action_path
         }}/entrypoint.sh \
-          "${{ inputs.register }}" \
+          "${{ inputs.registry }}" \
           "${{ inputs.repo }}" \
           "${{ inputs.image_sha }}"
         )
--- a/.github/actions/image2commit/action.yml
+++ b/.github/actions/image2commit/action.yml
@@ -1,0 +1,42 @@
+name: image2commit
+description: Resolve full commit SHA from a promoted image tag.
+
+inputs:
+  register:
+    description: "Registry (e.g., docker.io, quay.io)"
+    required: true
+  repo:
+    description: "Repository path (e.g., andrpac/my-repo)"
+    required: true
+  image_sha:
+    description: "Short SHA or 'latest'"
+    required: true
+
+outputs:
+  commit_sha:
+    description: "Resolved full commit SHA"
+    value: ${{ steps.resolve.outputs.commit_sha }}
+runs:
+  using: "composite"
+  steps:
+    - name: Install skopeo and jq
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y skopeo jq
+
+    - name: Resolve commit SHA
+      id: resolve
+      shell: bash
+      run: |
+        chmod +x ${{ github.action_path }}/entrypoint.sh
+        full_sha=$(${{
+          github.action_path
+        }}/entrypoint.sh \
+          "${{ inputs.register }}" \
+          "${{ inputs.repo }}" \
+          "${{ inputs.image_sha }}"
+        )
+
+        echo "Raw full_sha: $full_sha"
+        echo "commit_sha=$full_sha" >> $GITHUB_OUTPUT
--- a/.github/workflows/promote-image.yml
+++ b/.github/workflows/promote-image.yml
@@ -21,19 +21,6 @@ jobs:
       - name: Checkout PR commit
         uses: actions/checkout@v4
       
-      # Note, we have to be careful how we retrive the image. The event that pushed 
-      # the image to the ghcr.io repo was mainly a push/schedule that passed all the
-      # tests. This event has access to github.ref_name. However, the workflow_run
-      # event does not have access github.ref_name set up. 
-      # 
-      # Therefore, we need to manually specify the branch as main
-      - name: Prepare image tag
-        id: set_tag
-        uses: ./.github/actions/set-tag
-        with:
-          branch_name: ${{ github.event.workflow_run.head_branch }}
-          commit_sha: ${{ github.event.workflow_run.head_sha }}
-
       - name: Log in to the GitHub Container Registry
         uses: docker/login-action@v3
         with:
@@ -41,23 +28,26 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Pull unofficial image from GitHub Container Registry
-        run: |
-          docker pull ${{ env.GHCR_REPO }}:${{ steps.set_tag.outputs.tag }}
-
       - name: Login to Docker registry
         uses: docker/login-action@v3
         with:
           registry: docker.io
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-            
+      
       - name: Log in to Quay registry
         uses: docker/login-action@v3
         with:
           registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Prepare image tag
+        id: set_tag
+        uses: ./.github/actions/set-tag
+        with:
+          branch_name: ${{ github.event.workflow_run.head_branch }}
+          commit_sha: ${{ github.event.workflow_run.head_sha }}
       
       - name: Prepare tag for promoted image
         id: promoted_tag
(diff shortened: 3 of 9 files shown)


.github/actions/build-push-image/action.yaml
```diff
@@ -99,3 +99,5 @@ runs:
         sbom: true
         push: true
         tags: ${{ inputs.tags }}
+        labels: |
+          org.opencontainers.image.revision=${{ github.sha }}
```
josvazg, line 103: **q**: Why is this needed now?

.github/actions/image2commit/action.yml
```diff
@@ -0,0 +1,42 @@
+name: image2commit
+description: Resolve full commit SHA from a promoted image tag.
+
+inputs:
+  register:
```
josvazg: **q**: Isn't registry more accurate?

```suggestion
  registry:
```

.github/actions/build-push-image/action.yaml
josvazg: ok I saw it used in the `entrypoint.sh`

.github/workflows/promote-image.yml
```diff
@@ -21,43 +21,33 @@ jobs:
...
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Pull unofficial image from GitHub Container Registry
-        run: |
-          docker pull ${{ env.GHCR_REPO }}:${{ steps.set_tag.outputs.tag }}
-
```
josvazg, line 47: **q**: why is this not needed now?

.github/workflows/release-image.yml
```diff
@@ -0,0 +1,270 @@
...
+        default: "latest"
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+     
+jobs:
+  # Image2commit: Creates a mapping between the image_sha given as input and the actual git commit
+  # This is necassary for the release-image step that requires checking out that exact git commit
```
josvazg: **nit**: typo

```suggestion
  # This is necessary for the release-image step that requires checking out that exact git commit
```

.github/workflows/release-image.yml
```diff
@@ -0,0 +1,270 @@
...
+          docker_image_url="${DOCKER_RELEASE_REPO}:${release_tag}"
+          quay_image_url="${QUAY_RELEASE_REPO}:${release_tag}"
+          quay_certified_image_url="${QUAY_RELEASE_REPO}:${certified_tag}"
+
+          echo "promoted_tag=$promoted_tag"                         >> $GITHUB_OUTPUT
+          echo "release_tag=$release_tag"                           >> $GITHUB_OUTPUT
+          echo "certified_tag=$certified_tag"                       >> $GITHUB_OUTPUT
+          echo "docker_image_url=$docker_image_url"                 >> $GITHUB_OUTPUT
+          echo "quay_image_url=$quay_image_url"                     >> $GITHUB_OUTPUT
+          echo "quay_certified_image_url=$quay_certified_image_url" >> $GITHUB_OUTPUT
```
josvazg: **q**: Why using outputs rather than simpler env vars?

All vars here are just text compositions, they do not depend on a command output.
Maybe I missed something.

.github/workflows/release-image.yml
```diff
@@ -0,0 +1,270 @@
...
+
+          git add -f "$RELEASE_DIR"
+          scripts/create-signed-commit.sh
+
+          gh pr create \
+            --draft \
+            --base main \
+            --head "$BRANCH" \
+            --title "$COMMIT_MESSAGE" \
+            --body "This is an autogenerated PR to prepare for the release"
```
josvazg: **follow-up**: this could be a makefile rule and a script instead. But it is fine to do as a follow up.

scripts/move-image.sh
```diff
@@ -13,25 +13,41 @@
...
+docker buildx imagetools create "${image_src_url}" --tag "${image_dest_url}"
+
+if [[ "${ALIAS_ENABLED}" == "true" && -n "${ALIAS_TAG}" ]]; then
+  echo "Aliasing ${image_src_url} as ${IMAGE_DEST_REPO}:${ALIAS_TAG}"
+  docker buildx imagetools create "${image_src_url}" --tag "${IMAGE_DEST_REPO}:${ALIAS_TAG}"
+  echo "Successfully aliased as ${IMAGE_DEST_REPO}:${ALIAS_TAG}"
+fi
+
+docker buildx rm "${BUILDER_NAME}" > /dev/null
+echo "Successfully moved ${image_src_url} -> ${image_dest_url}"
```
josvazg, line 53: **note**: TIL that multi-arch images could not be simply pull and push! I guess it is because by default you will only get the local platform, rather than the full multi-arch image when you pull.

Great job!

.github/workflows/release-image.yml
```diff
@@ -0,0 +1,270 @@
...
+          export RELEASE_DIR="releases/${VERSION}"
+
+          git config --global user.name "${{ steps.generate_token.outputs.user-name }}"
+          git config --global user.email "${{ steps.generate_token.outputs.user-email }}"
+
+          mkdir -p "$RELEASE_DIR"
+          cp -r deploy "$RELEASE_DIR/deploy"
+          cp -r bundle "$RELEASE_DIR/bundle"
+          cp -r helm-charts "$RELEASE_DIR/helm-charts"
+          cp bundle.Dockerfile "$RELEASE_DIR/bundle.Dockerfile"
```
josvazg, line 220: **q**: how does this work? I mean, in theory only the release directory should be changed, but AFAICS here we will change both the release dir and the legacy deploy bundle & helm-charts dirs right? am I missing something

josvazg (commented):
LGTM great job!

Before approving, I have some questions that I need help with to fully understand teh change.

Also the release docs need change and we should mark the old workflows as "deprecated" in the name right?

.github/actions/image2commit/action.yml
andrpac: Oh yeah, it is.

.github/workflows/promote-image.yml
andrpac: The move-image.sh already does that here. 

`docker buildx imagetools create "${image_src_url}" --tag "${image_dest_url}"`

Note, docker pull is problematic as this only pulls for the underlying architecture. So in our case most likely amd64. So the subsequent push will only have one architecture.

.github/workflows/release-image.yml
andrpac: Oh yes you're right. I did not even notice that. This is my mistake since before the logic used to be way more complex and a lot more things used to happen in here.

scripts/move-image.sh
andrpac: Exactly!!

.github/workflows/release-image.yml
andrpac: You're not really changing anything. Just to make sure, the RELEASE_DIR won't be there before, since this is a new release, right? So this directory is untracked by GitHub (so it stays there between git checkouts). 

Deploy dir will be the only `brand-new` thing we get since it has to use the new image and overwrite the other existing deploy dir. So we have to save this somewhere before we get the HEAD of main. So we put everything in the untracked RELEASE_DIR. 

Checkout -f will leave the files intact in the release commit (nothing was committed to this point). We only track (git add) and commit the new RELEASE_DIR in the new branch.

.github/workflows/promote-image.yml
josvazg: thx for the explanation. 

**q**: should we leave a comment as a reminder for future us somewhere about his?

andrpac: Yeah, I think it makes sense. I also wanted to have an explicit commit with that so that other people know this.

.github/workflows/release-image.yml
josvazg: I do not quite get it. There are 3 commits involved here:
- A: The released version commit.
- B: The latest in which we will merge into, which might or not might not be the same as the released version commit.
- C: The new commit that will include the new `RELEASE_DIR`

When A=B=`latest` all is good, because dirs `deploy`, `bundle` and `helm-charts` should already reflect the version we want to create the `RELEASE_DIR` from.

When A is older than B, I am not sure how this works and how we avoid issues. Note that C will go on top of B and it should ONLY include changes in `RELEASE_DIR`.

andrpac: So assume we have something like and we will release from A

`A -> .... -> B`. 

We will end up doing something like this right? 

`A -> .... -> B -> C` where, as you said, C will include the new dir RELEASE_DIR. 

Now, we sit at commit A since we do an explicit checkout at the very beginning at that exact commit. We run the script. Note, the only Modified file to git will be the deploy directory since that was changed with the new image. 

We now create this new directory RELEASE_DIR which is untracked by git. We copy all the files inside. These files in the new RELEASE_DIR will also be untracked. 

Now, we do a git checkout -f which will revert back the original release commit A since nothing was committed up to this point. We now create this new branch from the head of main, C. We only add this new directory in C so now it becomes tracked by git. Note, this will not exist in any of the older commits right? Since it was just added in this commit. 

So if you were to go back and do a git checkout at A, nothing should have changed. Maybe I am missing something, and I do not see it.

josvazg: thx for the explanation, I missed the `checkout -f` trick I believe.

.github/workflows/release-image.yml
andrpac: Yes, I agree. I will do this as a cleanup step. Since I'm also planning on cleaning up other things.

josvazg (approved):
LGTM

Do not forget the releases doc changes, but it is fine if they come on a follow up PR

andrpac:
> LGTM
> 
> Do not forget the releases doc changes, but it is fine if they come on a follow up PR



> LGTM
> 
> Do not forget the releases doc changes, but it is fine if they come on a follow up PR

Yes and it makes sense. I would love to do them now, but let's see the pipeline working first! Thank you for the feedback!

helderjs (approved):

PR #2528, opened 2025-07-18
CLOUDP-331104/manual-promotion

andrpac:
# Summary

This pr should allow promotions of images on merging into main or manually if all tests run and pass.

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/.github/workflows/promote-image.yml
+++ b/.github/workflows/promote-image.yml
@@ -6,13 +6,39 @@ on:
     types: [completed]
 
 jobs:
+  # Check-test-status: Uses the hook called test-status on the test.yml job in order to 
+  # see if any test failed or was skipped; promotion should only run when all tests suceeded!
+  check-test-status:
+    runs-on: ubuntu-latest
+    outputs:
+      test-status: ${{ steps.set-status.outputs.status }}
+    steps:
+      - name: Ensure test-status job completed
+        id: set-status
+        env:
+          REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/promote-image.yml
```diff
@@ -6,13 +6,38 @@ on:
     types: [completed]
 
 jobs:
+  # Check-test-status: Uses the hook called test-status on the test.yml job in order to 
+  # see if any test failed or was skipped; promotion should only run when all tests suceeded!
+  check-test-status:
```
josvazg, line 11: **q**: Why do we need this check? Can 't we just ensure the promotion can only be kicked off after the whole test suite succeeded? (regardless of the tests being initiated manually or by a schedule)

Context in my mind, but I might be wrong:
1. Promotion cannot be kicked directly neither manually not by schedule.
2. Promotion only happens as a result of a full test suite being successful with  promotion flag turned.
3. Test can be either scheduled as nightlies with promotion or kicked off manually with promotion, which in case of tests passing will trigger the promotion.

andrpac: So here's the annoying thing about GitHub Actions. It marks a a workflow as a success even if some of the jobs inside were skipped. 

Note that we need the cloud-test to explicitly run! Otherwise, we won't have the image stored in ghcr.io and the promote image workflow will then raise an error (complains about image missing from ghcr.io). Therefore, the check-test-status does just that. It's a hook that tells us if all the jobs ran (and succeeded) or not. 

Right now promotion happens in 3 different cases:
1) On schedule (which is nightly)
2) On merges when production code was changed (as we already run the whole test suite)
3) Manually with promotion flag enabled.

andrpac: Also note, we do not have an explicit check in the promotion image workflow that tells us whether we arrived from a manual, merge or schedule. The only thing we want is the check-test-status to succeed.

josvazg: ok, I see.

josvazg (approved):

PR #2537, opened 2025-07-22
Set flex 2 dedicated nightly tag to run after merge

josvazg:
# Summary

Ensure nightly and after merge test run the new e2e2 tests for flex to dedicated.

## Proof of Work

CI will run those tests **after** merge.

## Checklist

### DIFF
--- a/.github/workflows/tests-e2e2.yaml
+++ b/.github/workflows/tests-e2e2.yaml
@@ -74,7 +74,7 @@ jobs:
           E2E2_LABELS: ${{ env.e2e2_labels }}
           USE_JSON: true
           SKIP_PREFIXES: "[\"nightly\"]"
-          NIGHTLY_MATRIX: "[\"nightly-core\",\"nightly-integration\"]"
+          NIGHTLY_MATRIX: "[\"nightly-core\",\"nightly-integration\",\"nightly-flex2dedicated\"]"
         run: |
           # Nightly runs all tests, overriding PR labels as '["test/e2e2/*"]'
           if [ "${{ github.ref }}" == "refs/heads/main" ];then
--- a/test/e2e2/flex_to_dedicated_test.go
+++ b/test/e2e2/flex_to_dedicated_test.go
@@ -40,7 +40,7 @@ import (
 //go:embed flex2dedicated/*
 var flex2dedicated embed.FS
 
-var _ = Describe("Flex to Dedicated Upgrade", Ordered, Label("flex-to-dedicated"), func() {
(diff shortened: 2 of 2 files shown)


andrpac (approved):

.github/workflows/tests-e2e2.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
           E2E2_LABELS: ${{ env.e2e2_labels }}
           USE_JSON: true
           SKIP_PREFIXES: "[\"nightly\"]"
-          NIGHTLY_MATRIX: "[\"nightly-core\",\"nightly-integration\"]"
+          NIGHTLY_MATRIX: "[\"nightly-core\",\"nightly-integration\",\"nightly-flex2dedicated\"]"
```
josvazg, line 77: I am thinking we could probably automate this list. Using selectable tags this would be `test/e2e2/nightly-*` 
WDYT @helderjs ?

I would do it in a separate PR though

test/e2e2/flex_to_dedicated_test.go
```diff
@@ -40,7 +40,7 @@ import (
 //go:embed flex2dedicated/*
 var flex2dedicated embed.FS
 
-var _ = Describe("Flex to Dedicated Upgrade", Ordered, Label("flex-to-dedicated"), func() {
+var _ = Describe("Flex to Dedicated Upgrade", Ordered, Label("nightly-flex2dedicated", "flex-to-dedicated"), func() {
```
josvazg, line 43: Having to remember this is the biggest issue I have right now about this strategy.

The invariant is, every ´e2e2´ test must include at least one `nightly-` label. We could write a test to fail if that is not the case. I do not think it would be so complex. WDYT @helderjs ?

.github/workflows/tests-e2e2.yaml
helderjs: It will be good. Less one place to touch and less risk to forget.

test/e2e2/flex_to_dedicated_test.go
helderjs: Right test to test smells strange to me. By what we have discussed, needs vary from test from test, and the rule "must include at least one" sounds strict to me.

helderjs (approved):

test/e2e2/flex_to_dedicated_test.go
josvazg: Think of it as a linter that check test coverage: "are all e2e2 tests covered as nightlies?"

PR #2560, opened 2025-07-31
CLOUDP-329522: Fix Release Workflow

roothorp:
# Summary

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/scripts/move-image.sh
+++ b/scripts/move-image.sh
@@ -29,6 +29,12 @@ ALIAS_ENABLED="${ALIAS_ENABLED:-false}"
 image_src_url="${IMAGE_SRC_REPO}:${IMAGE_SRC_TAG}"
 image_dest_url="${IMAGE_DEST_REPO}:${IMAGE_DEST_TAG}"
 
+echo "Checking if ${image_dest_url} already exists remotely..."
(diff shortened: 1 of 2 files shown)


scripts/move-image.sh
```diff
@@ -29,12 +29,6 @@ ALIAS_ENABLED="${ALIAS_ENABLED:-false}"
 image_src_url="${IMAGE_SRC_REPO}:${IMAGE_SRC_TAG}"
 image_dest_url="${IMAGE_DEST_REPO}:${IMAGE_DEST_TAG}"
 
-echo "Checking if ${image_dest_url} already exists remotely..."
```
josvazg: **q**: Why is this not needed now?

roothorp: because we want the option to overwrite images in the event of a re-release. we don't technically need it here since the reason we are re-releasing is that the tag is incorrect, but I think it makes it easier should we need to in future.

roothorp: it admittedly does open us up to the risk of overwriting a tag if someone accidentally triggers a release for an already release version, however

josvazg: I see, ok

PR #2575, opened 2025-08-05
CLOUDP-334941: Connection Secret Controller

andrpac:
# Summary

This pr creates the ConnectionSecret controller, removing all the duplicated logic in the Deployment, DataFederation, and DatabaseUser controllers. 

Jira-https://jira.mongodb.org/browse/CLOUDP-334941

## General idea of the pipeline.

The connection secrets require a paired resource. The AtlasDatabaseUser is the main parent (since the connection secret will live in the same namespace as the User). The other parent has the general type of Endpoint. Endpoints could be either AtlasDeployment or AtlasDataFederation. The abstraction of Endpoint creates a clear separation from the logic of controller. That is to say, if someone wants to add a new type of Endpoint, the logic of the controller should remain the same.

Watchers primarily care about AtlasDatabaseUsers, AtlasDeployments and AtlasDataFederation. When a resource comes in, we create the request for reconciliation as follows: pick up the other counterpart of the pair and create all possible pairs using the following naming convention: `projectID$clusterName$DatabaseUsername`. The `$` is used in order to be able to parse the name in the reconciler. For example, if a AtlasDatabaseUser comes in with username `user`, we would create the following requests: `projectID$cluster1$user`, `projectID$cluster2$user`, .... , `projectID$df1$user`, `projectID$df2$user`, ...

In the reconciliation side, we first parse out the name (split across `$`) to extract identifiers of a connection secret. We then retrieve back the paired resource using indexers (they have a composite key of `projectID-clusterName` and `projectID-DatabaseUsername`). If any parent is missing, the secret should not exist (aka be deleted if it exists). We then check the scope and expiration date of user to see that they still match (if not, delete the secret if it exists). Finally, we check that the pair is ready before proceeding to upserting the connection secret. 

This only presents the general flow. For a more in-depth read, check out the attached TD to this ticket.

## Proof of Work

1. Implemented unit-tests that cover 85% of introduced code for Connection Secret controller
2. Implement integration test that triggers 1 Deployment, 1 DataFederation and 2 Users and checks for connection secrets as the spec and scopes of Users updates.
3. Made sure that all old integration tests still work.

## Checklist

  * **q**: Clarifying question.


internal/controller/atlasdatabaseuser/databaseuser.go
```diff
@@ -304,7 +274,7 @@ func (r *AtlasDatabaseUserReconciler) removeOldUser(ctx context.Context, dbUserS
 	return err
 }
 
-func isExpired(atlasDatabaseUser *akov2.AtlasDatabaseUser) (bool, error) {
+func IsExpired(atlasDatabaseUser *akov2.AtlasDatabaseUser) (bool, error) {
```
helderjs: As this is now shared between controller, move it to `timeutil` package and make it work with `time.Time` directly.
ex: `timeuitl.IsExpired(dbUser.Spec.DeleteAfterDate)`

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+				return false
+			}
+			oldObj, ok := e.ObjectOld.(T)
+			if !ok {
+				return false
+			}
+			return !ready(oldObj) && ready(newObj)
+		},
+	}
+}
```
helderjs: Move predicate related types and functions to `watch` package.

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+
+func (r *ConnectionSecretReconciler) For() (client.Object, builder.Predicates) {
+	// Filter out connection secrets based on the required labels
+	labelPredicates := predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		labels := obj.GetLabels()
+		_, hasType := labels[TypeLabelKey]
+		_, hasProject := labels[ProjectLabelKey]
+		_, hasCluster := labels[ClusterLabelKey]
+		return hasType && hasProject && hasCluster
+	})
```
helderjs: Same here

internal/controller/registry.go
```diff
@@ -128,6 +129,7 @@ func (r *Registry) registerControllers(c cluster.Cluster, ap atlas.Provider) {
 
 	integrationsReconciler := integrations.NewAtlasThirdPartyIntegrationsReconciler(c, ap, r.deletionProtection, r.logger, r.globalSecretRef, r.reapplySupport)
 	reconcilers = append(reconcilers, newCtrlStateReconciler(integrationsReconciler))
+	reconcilers = append(reconcilers, connectionsecret.NewConnectionSecretReconciler(c, r.deprecatedPredicates(), ap, r.logger, r.globalSecretRef))
```
helderjs: Why `deprecatedPredicates()` instead of `defaultPredicates()`?
Unless there's a good reason, deprecated code must be avoidded.

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+	deployments := &akov2.AtlasDeploymentList{}
+	if err := r.Client.List(ctx, deployments, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(indexer.AtlasDeploymentByProject, projectID),
+	}); err != nil {
+		r.Log.Errorf("failed to list AtlasDeployments: %v", err)
+		return nil
+	}
+
+	return r.generateConnectionSecretRequests(projectID, deployments.Items, []akov2.AtlasDatabaseUser{*user})
+}
```
helderjs: Resolve project id can be generic

```suggestion
func (r *ConnectionSecretReconciler) ResolveProjectId(ctx context.Context, ref akov2.ProjectDualReference, parentNamespace string) (string, error) {
	if ref.ExternalProjectRef != nil && ref.ExternalProjectRef.ID != "" {
		return ref.ExternalProjectRef.ID, nil
	}
	if ref.ProjectRef != nil && ref.ProjectRef.Name != "" {
		project := &akov2.AtlasProject{}
		if err := r.Client.Get(ctx, *ref.ProjectRef.GetObject(parentNamespace), project); err != nil {
			return "", fmt.Errorf("failed to resolve projectRef from deployment: %w", err)
		}
		return project.ID(), nil
	}
	return "", fmt.Errorf("missing both external and internal project references")
}

func (r *ConnectionSecretReconciler) newDeploymentMapFunc(ctx context.Context, obj client.Object) []reconcile.Request {
	deployment, ok := obj.(*akov2.AtlasDeployment)
	if !ok {
		r.Log.Warnf("watching AtlasDeployment but got %T", obj)
		return nil
	}

	projectID, err := r.ResolveProjectId(ctx, deployment.Spec.ProjectDualReference, deployment.GetNamespace())
	if err != nil {
		r.Log.Errorw("Unable to resolve projectID for deployment", "error", err)
		return nil
	}

	users := &akov2.AtlasDatabaseUserList{}
	if err := r.Client.List(ctx, users, &client.ListOptions{
		FieldSelector: fields.OneTermEqualSelector(indexer.AtlasDatabaseUserByProject, projectID),
	}); err != nil {
		r.Log.Errorf("failed to list AtlasDatabaseUsers: %v", err)
		return nil
	}

	return r.generateConnectionSecretRequests(projectID, []akov2.AtlasDeployment{*deployment}, users.Items)
}

func (r *ConnectionSecretReconciler) newDatabaseUserMapFunc(ctx context.Context, obj client.Object) []reconcile.Request {
	user, ok := obj.(*akov2.AtlasDatabaseUser)
	if !ok {
		r.Log.Warnf("watching AtlasDatabaseUser but got %T", obj)
		return nil
	}

	projectID, err := r.ResolveProjectId(ctx, user.Spec.ProjectDualReference, user.GetNamespace())
	if err != nil {
		r.Log.Errorw("Unable to resolve projectID for user", "error", err)
		return nil
	}

	deployments := &akov2.AtlasDeploymentList{}
	if err := r.Client.List(ctx, deployments, &client.ListOptions{
		FieldSelector: fields.OneTermEqualSelector(indexer.AtlasDeploymentByProject, projectID),
	}); err != nil {
		r.Log.Errorf("failed to list AtlasDeployments: %v", err)
		return nil
	}

	return r.generateConnectionSecretRequests(projectID, deployments.Items, []akov2.AtlasDatabaseUser{*user})
}
```

internal/controller/connectionsecret/resource_manager.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	if u.Spec.ProjectRef != nil && u.Spec.ProjectRef.Name != "" {
+		project := &akov2.AtlasProject{}
+		if err := c.Get(ctx, *u.Spec.ProjectRef.GetObject(u.Namespace), project); err != nil {
+			return "", fmt.Errorf("failed to resolve projectRef from user: %w", err)
+		}
+		return project.ID(), nil
+	}
+	return "", fmt.Errorf("missing both external and internal project references")
+}
```
helderjs: If using the generic method in the controller, there's no need of this function
```suggestion
```

internal/controller/connectionsecret/resource_manager.go
```diff
@@ -0,0 +1,71 @@
...
+	return false
+}
+
+func IsDeploymentReady(d *akov2.AtlasDeployment) bool {
+	return HasReadyCondition(d.Status.Conditions)
+}
+
+func IsDatabaseUserReady(u *akov2.AtlasDatabaseUser) bool {
+	return HasReadyCondition(u.Status.Conditions)
+}
```
helderjs: This can move along with predicates to the `watch` package

internal/controller/connectionsecret/requestname_extractor.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// LoadRequestIdentifiers determines whether the request name is internal or K8s format
+// and extracts ProjectID, ClusterName, and DatabaseUsername.
+func LoadRequestIdentifiers(ctx context.Context, c client.Client, req types.NamespacedName) (ConnSecretIdentifiers, error) {
+	var ids ConnSecretIdentifiers
+
+	// === Internal format: <ProjectID>$<ClusterName>$<DatabaseUserName>
+	if strings.Contains(req.Name, InternalSeparator) {
+		parts := strings.SplitN(req.Name, InternalSeparator, 3)
```
helderjs: If the number of part is not 3, something is wrong, not matter if less or more items.
```suggestion
		parts := strings.Split(req.Name, InternalSeparator)
```

internal/controller/connectionsecret/requestname_extractor.go
```diff
@@ -0,0 +1,292 @@
...
+		return ids, ErrK8sNameSplitEmpty
+	}
+
+	return ConnSecretIdentifiers{
+		ProjectID:        projectID,
+		ProjectName:      parts[0],
+		ClusterName:      clusterName,
+		DatabaseUsername: parts[1],
+	}, nil
+}
```
helderjs:
```suggestion
func (r *ConnectionSecretReconciler) loadRequestIdentifiers(ctx context.Context, req types.NamespacedName) (*ConnSecretIdentifiers, error) {
	if strings.Contains(req.Name, InternalSeparator) {
		return r.indetifiersFromInternalName(ctx, req)
	}
	
	return indentifiersFromSecret(ctx, req)
}

func (r *ConnectionSecretReconciler) indetifiersFromInternalName(ctx context.Context, req types.NamespacedName) (*ConnSecretIdentifiers, error) {
    	// === Internal format: <ProjectID>$<ClusterName>$<DatabaseUserName>
	parts := strings.Split(req.Name, InternalSeparator)
	if len(parts) != 3 {
		return nil, ErrInternalFormatPartsInvalid
	}
	if parts[0] == "" || parts[1] == "" || parts[2] == "" {
		return nil, ErrInternalFormatPartEmpty
	}
	return ConnSecretIdentifiers{
		ProjectID:        parts[0],
		ClusterName:      parts[1],
		DatabaseUsername: parts[2],
	}, nil
}

func (r *ConnectionSecretReconciler) indentifiersFromSecret(ctx context.Context, req types.NamespacedName) (*ConnSecretIdentifiers, error) {
	// === K8s format: <ProjectName>-<ClusterName>-<DatabaseUserName>
	var secret corev1.Secret
	if err := r.Client.Get(ctx, req, &secret); err != nil {
		return nil, err
	}

	labels := secret.GetLabels()
	projectID, hasProject := labels[ProjectLabelKey]
	clusterName, hasCluster := labels[ClusterLabelKey]

	// Missing labels or values
	if !hasProject || !hasCluster {
		return nil, ErrK8sLabelsMissing
	}
	if projectID == "" || clusterName == "" {
		return nil, ErrK8sLabelEmpty
	}

	sep := fmt.Sprintf("-%s-", clusterName)
	parts := strings.SplitN(req.Name, sep, 2)
	if len(parts) != 2 {
		return nil, ErrK8sNameSplitInvalid
	}
	if parts[0] == "" || parts[1] == "" {
		return nil, ErrK8sNameSplitEmpty
	}

	return ConnSecretIdentifiers{
		ProjectID:        projectID,
		ProjectName:      parts[0],
		ClusterName:      clusterName,
		DatabaseUsername: parts[1],
	}, nil
}
```

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+	GlobalPredicates []predicate.Predicate
+	EventRecorder    record.EventRecorder
+}
+
+func (r *ConnectionSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Parses the request name and fills up the identifiers: ProjectID, ClusterName, DatabaseUsername
+	log := r.Log.With("ns", req.Namespace, "name", req.Name)
+	log.Debugw("reconcile started")
+
+	ids, err := LoadRequestIdentifiers(ctx, r.Client, req.NamespacedName)
```
helderjs:
```suggestion
	ids, err := r.loadRequestIdentifiers(ctx, req.NamespacedName)
```

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			log.Debugw("connectionsecret not found; assuming deleted")
+			return workflow.TerminateSilently(nil).WithoutRetry().ReconcileResult()
+		}
```
helderjs: If it was deleted, accidentally, will it not be recreated?

internal/controller/connectionsecret/requestname_extractor.go
```diff
@@ -0,0 +1,292 @@
...
+// LoadPairedResources fetches the paired AtlasDeployment and AtlasDatabaseUser forming the ConnectionSecret
+// using the registered indexers
+func LoadPairedResources(ctx context.Context, c client.Client, ids ConnSecretIdentifiers, namespace string) (*ConnSecretPair, error) {
+	compositeDeploymentKey := ids.ProjectID + "-" + ids.ClusterName
+	deployments := &akov2.AtlasDeploymentList{}
+	if err := c.List(ctx, deployments, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(indexer.AtlasDeploymentBySpecNameAndProjectID, compositeDeploymentKey),
+	}); err != nil {
+		return nil, err
+	}
```
helderjs: This key pair is unique, why listing instead of getting?

internal/controller/connectionsecret/requestname_extractor.go
```diff
@@ -0,0 +1,292 @@
...
+
+	compositeUserKey := ids.ProjectID + "-" + ids.DatabaseUsername
+	users := &akov2.AtlasDatabaseUserList{}
+	if err := c.List(ctx, users, &client.ListOptions{
```
helderjs: Same here, why not get?

internal/controller/connectionsecret/requestname_extractor.go
```diff
@@ -0,0 +1,292 @@
...
+		return &ConnSecretPair{
+			ProjectID:  ids.ProjectID,
+			Deployment: &deployments.Items[0],
+			User:       nil,
+		}, ErrNoUserFound
+	case len(deployments.Items) > 1:
+		return nil, ErrManyDeployments
+	case len(users.Items) > 1:
+		return nil, ErrManyUsers
+	}
```
helderjs: Not need to split every single possible error when they basically mean the same: there is no deploy/user or more than one deploy/user
```suggestion
	if len(deployments.Items) == 0 || len(users.Items) == 0 {
	    return errors.New("deployment or user is missing") // just example of message, adjust it for best
	}
```

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+
+		case errors.Is(err, ErrManyDeployments), errors.Is(err, ErrManyUsers):
+			log.Errorw("ambiguous pairing; multiple matches", "reason", workflow.ConnSecretAmbiguousResources, "error", err)
+			return workflow.Terminate(workflow.ConnSecretAmbiguousResources, err).ReconcileResult()
+
+		default:
+			log.Errorw("failed to load paired resources", "reason", workflow.ConnSecretInvalidResources, "error", err)
+			return workflow.Terminate(workflow.ConnSecretInvalidResources, err).ReconcileResult()
+		}
+	}
```
helderjs: This will be simple if we also simplify inner error handling

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+		switch {
+		// This means there's no owner resources; the secret will be garbage collected
+		case errors.Is(err, ErrNoPairedResourcesFound):
+			log.Debugw("no paired resources found")
+			return workflow.TerminateSilently(nil).WithoutRetry().ReconcileResult()
+
+		// This means an owner from the pair was deleted; the secret will be forcefully removed
+		case errors.Is(err, ErrNoDeploymentFound), errors.Is(err, ErrNoUserFound):
+			log.Infow("paired resource missing; scheduling deletion", "reason", workflow.ConnSecretOwnerMissing)
+			return r.handleDelete(ctx, req, ids, pair)
```
helderjs: With the resource owner being the database user you don't neeed this.

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+	// If the user expired, delete connection secret
+	expired, err := atlasdatabaseuser.IsExpired(pair.User)
+	if err != nil {
+		log.Errorw("failed to check expiration date", "reason", workflow.ConnSecretCheckExpirationFailed, "error", err)
+		return workflow.Terminate(workflow.ConnSecretCheckExpirationFailed, err).ReconcileResult()
+	}
+	if expired {
+		log.Infow("user expired; scheduling deletion", "reason", workflow.ConnSecretUserExpired)
+		return r.handleDelete(ctx, req, ids, pair)
+	}
```
helderjs: This is arguable, but I think we should not delete the connection secret if the user is expired.
If the database user resource exists, its connection secret also exists. In that way we avoid break customer integration and they will rely on Atlas failures (no able to use an expired user)

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+
+	// If the scope became invalid, delete connection secret
+	if pair.InvalidScopes() {
+		log.Infow("invalid scope; scheduling deletion", "reason", workflow.ConnSecretInvalidScopes)
+		return r.handleDelete(ctx, req, ids, pair)
+	}
```
helderjs: Why do you need to this here if you generate requests only for valid scopes? lines 199:202

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,282 @@
...
+
+	// Checks that AtlasDeployment and AtlasDatabaseUser are ready before proceeding
+	if ready, notReady := pair.IsReady(); !ready {
+		log.Debugw("waiting for paired resources to become ready", "notReady", strings.Join(notReady, ","))
+		return workflow.InProgress(workflow.ConnSecretNotReady, fmt.Sprintf("Not ready: %s", strings.Join(notReady, ", "))).ReconcileResult()
+	}
```
helderjs: Why do you need this if you have readiness check as part of the predicate?

internal/controller/connectionsecret/connectionsecrets.go
```diff
@@ -17,198 +17,266 @@ package connectionsecret
...
+	log.Infow("secret deleted", "reason", workflow.ConnSecretDeleted)
+	r.EventRecorder.Event(secret, corev1.EventTypeNormal, "Deleted", "ConnectionSecret deleted")
+	return workflow.TerminateSilently(nil).WithoutRetry().ReconcileResult()
 }
 
-func cleanupStaleSecrets(ctx *workflow.Context, k8sClient client.Client, projectID string, user akov2.AtlasDatabaseUser) error {
-	if err := removeStaleByScope(ctx, k8sClient, projectID, user); err != nil {
-		return err
+// handleUpdate manages the case where we will create or update the connection secret
+func (r *ConnectionSecretReconciler) handleUpdate(
```
helderjs: If it creates or updates, let have a better name.
```suggestion
func (r *ConnectionSecretReconciler) handleUpsert(
```

internal/controller/connectionsecret/connectionsecrets.go
```diff
@@ -17,198 +17,266 @@ package connectionsecret
...
+	ctx context.Context, req ctrl.Request, ids ConnSecretIdentifiers, pair *ConnSecretPair) (ctrl.Result, error) {
+	log := r.Log.With("ns", req.Namespace, "name", req.Name)
+
+	// ProjectName is required for ConnectionSecret metadata.name to create or update
+	projectName, err := r.resolveProjectName(ctx, ids, pair)
```
helderjs: Could the project name be resolved earlier? Together with the others.

internal/controller/connectionsecret/connectionsecrets.go
```diff
@@ -17,198 +17,266 @@ package connectionsecret
...
+			Name:      name,
+			Namespace: req.Namespace,
+		},
+	}
+
+	// Populate Secret data/labels
+	if err := fillConnSecretData(secret, ids, data); err != nil {
+		log.Errorw("failed to fill secret data", "reason", workflow.ConnSecretFailedToFillData, "error", err)
+		return workflow.Terminate(workflow.ConnSecretFailedToFillData, err).ReconcileResult()
+	}
```
helderjs: This could be a separate function

internal/controller/connectionsecret/listsecrets.go
```diff
@@ -19,12 +19,39 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/kube"
 )
 
+func Ensure(ctx context.Context, client client.Client, namespace, projectName, projectID, clusterName string, data ConnSecretData) (string, error) {
```
helderjs: Where is this being used?

internal/controller/connectionsecret/requestname_extractor.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// ConnSecretPair represents the pairing of an AtlasDeployment and an AtlasDatabaseUser
+// required to construct a ConnectionSecret. It holds resolved identifiers and the corresponding resources.
+// NOTE: this struct intentionally stores only ProjectID (not all identifiers) to keep only necessary information.
+type ConnSecretPair struct {
+	ProjectID  string
+	Deployment *akov2.AtlasDeployment
+	User       *akov2.AtlasDatabaseUser
+}
```
helderjs: I have the impression, we don't need these 2 structs, your load/resolve functions should be one(instead of 2) and return a secret already with empty data if new, or the current one if got from the cluster.
Later you would just update the data and persist it.

internal/controller/connectionsecret/listsecrets.go
andrpac: In listsecrets_test.go.

internal/controller/connectionsecret/connectionsecret_controller.go
andrpac: This makes sense

internal/controller/connectionsecret/connectionsecret_controller.go
andrpac: I created a generic SecretLabelPredicate that filters secrets based on the presence of certain labels. This would make the most sense since we should not hardcode the labels in the watch package specifically for connection secrets.

internal/controller/registry.go
andrpac: Oh cool. I did not see this before.

internal/controller/connectionsecret/connectionsecret_controller.go
andrpac: Perfect. If that's the case, I will remove the resource manager file, and move all the generic logic into the controller.

internal/controller/connectionsecret/resource_manager.go
andrpac: I moved them to conditions.go and their respective types. I think it makes more sense than in the watch package. Let me know what you think.

internal/controller/connectionsecret/listsecrets.go
helderjs: So there's no reason to keep it

internal/controller/connectionsecret/connectionsecret_controller.go
andrpac: This is a very complicated case. If the user deleted it (accidentally or not), we won't be able to recreate it. Once it's deleted, any call to client.Get() will fail (secret is missing) and thus we won't be able to reconstruct it back. The user would have to wait for the sync to get it back.

What I could do is keep on reconciling, but those requests are pretty much a waste since the user most likely won't recreate the secret manually after he deleted it.

internal/controller/connectionsecret/requestname_extractor.go
andrpac: indexers don't work with Client.Get(). You would need to do List. Moreover, listing also allows us to do error handling if the user created the same deployment into 2 or more namespaces. This will keep on firing an `ambiguous pairing` error.

This is from the official docs: 

"Indexes may be added to caches using a FieldIndexer. This allows you to easily and efficiently look up objects with certain properties. You can then make use of the index by specifying a field selector on calls to List on the Reader corresponding to the given Cache."

internal/controller/connectionsecret/connectionsecret_controller.go
andrpac: The current implementation deletes the secrets once they're expired. One test case also checks for this. I think the behaviour should stay consistent.

internal/controller/connectionsecret/connectionsecret_controller.go
andrpac: Good point. because users can change the scopes later. 

Let me give you an example to clarify. suppose I have 1 User and 3 Deployments. Assume I did not write any scopes on that User. Therefore, we would have 3 connections-secreted created for every user-deployment pair right? 

Now, suppose I come in later and update the Scope field on that User to only Deployment1. This would create 2 Update requests right? One for the `oldObject` and one for `newObject`. The `oldObject` still has the old Scopes which means it will create again 3 reconciliation requests (for all user-deployment pair). However, in the Reconciler, when the request for user-deployment2 and user-deployment3 arrive, when we fetch the User from cache using the indexers, we will see that the scopes no longer align. Note, the cache is aligned with the `newObject`! 

So we will actively delete those secrets with invalid scopes. Hope this clarifies.

internal/controller/connectionsecret/connectionsecret_controller.go
andrpac: The predicate only checks that one resource is ready, not both.

internal/controller/connectionsecret/connectionsecrets.go
andrpac: I wouldn't do that. Resolving projectName is technically the most expensive part if the project is not in the cluster. I would only do it where necessary to avoid potential unnecessary SDK calls.

internal/controller/connectionsecret/requestname_extractor.go
andrpac: The reason I had two different structs it to indicate the division of responsibility. ConnIdentifiers is supposed to contain all the necessary information for identifying connection secrets. ConnPair contains the actual parent resources.

internal/controller/connectionsecret/listsecrets.go
andrpac: Yeah, I simplified the tests inside the listsecrets_test.go to no longer use it. And I removed it subsequently.

internal/controller/connectionsecret/connectionsecret_controller.go
andrpac: I thought about it more, and we still need that in there. You are right that if a user is deleted, it should be garbage collected. But at the same time, if I manually update the username field in the User and apply, I would also need to remove the old connection secret somehow, explicitly.

Makefile
```diff
@@ -573,7 +573,7 @@ install-credentials: set-namespace ## Install the Atlas credentials for the Oper
 
 .PHONY: prepare-run
 prepare-run: generate vet manifests run-kind install-crds install-credentials
-	rm bin/manager
+	rm -rf bin/manager
```
josvazg: q: why recursive? `bin/manager` is a file, not a directory.

api/condition.go
```diff
@@ -177,6 +177,15 @@ func HasConditionType(typ ConditionType, source []Condition) bool {
 	return false
 }
 
+func HasReadyCondition(conditions []Condition) bool {
```
josvazg, line 180: nit: I believe there is a generic helper in k8s libs for this, let me check...

josvazg: Here it is:
https://pkg.go.dev/k8s.io/apimachinery@v0.33.3/pkg/api/meta#IsStatusConditionTrue

but you woudl need a [metav1](https://pkg.go.dev/k8s.io/apimachinery@v0.33.3/pkg/apis/meta/v1).[Condition](https://pkg.go.dev/k8s.io/apimachinery@v0.33.3/pkg/apis/meta/v1#Condition) to use it, which is not what you have here. This is a custom `Condition` type.

Take this as an informational comment. Nothing we can do here while we are using custom Conditions.

Makefile
andrpac: If the user does not have bin/manager, `make run` will error out. But yes, you are right, we could just do `-f`, so it silence the error

api/condition.go
andrpac: oh cool, did not know this. I am not super familiar with all the k8s libraries

api/v1/atlasdatabaseuser_types.go
```diff
@@ -197,6 +197,11 @@ func (p *AtlasDatabaseUser) ProjectDualRef() *ProjectDualReference {
 	return &p.Spec.ProjectDualReference
 }
 
+// IsDatabaseUserReady checks if the Ready condition is available
```
josvazg, line 200: note for future shelves: this will not play nice with autogenerated CRDs

But those use `metav1.Condition` so we might have a generic way instead of calling one per type like here.

Nothing actionable for this PR though

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+	TypeLabelKey           = "atlas.mongodb.com/type"
+	CredLabelVal           = "credentials"
+
+	userNameKey     string = "username"
+	passwordKey     string = "password"
+	standardKey     string = "connectionStringStandard"
+	standardKeySrv  string = "connectionStringStandardSrv"
+	privateKey      string = "connectionStringPrivate"
+	privateSrvKey   string = "connectionStringPrivateSrv"
+	privateShardKey string = "connectionStringPrivateShard"
```
josvazg: **nit**: I believe the string type here is redundant:

```suggestion
	ProjectLabelKey = "atlas.mongodb.com/project-id"
	ClusterLabelKey = "atlas.mongodb.com/cluster-name"
	TypeLabelKey    = "atlas.mongodb.com/type"
	CredLabelVal    = "credentials"

	userNameKey     = "username"
	passwordKey     = "password"
	standardKey     = "connectionStringStandard"
	standardKeySrv  = "connectionStringStandardSrv"
	privateKey      = "connectionStringPrivate"
	privateSrvKey   = "connectionStringPrivateSrv"
	privateShardKey = "connectionStringPrivateShard"
```

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/controller/workflow"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/indexer"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/kube"
+)
+
+const (
+	InternalSeparator string = "$"
```
josvazg: string typing is redundant in a const right?

```suggestion
	InternalSeparator = "$"
```

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+	if strings.Contains(req.Name, InternalSeparator) {
+		return r.indetifiersFromInternalName(req)
+	}
+
+	return r.indentifiersFromK8s(ctx, req)
+}
+
+// indetifiersFromInternalName loads the identifiers for the internal format
+// === Internal format: <ProjectID>$<ClusterName>$<DatabaseUserName>
+func (r *ConnSecretReconciler) indetifiersFromInternalName(req types.NamespacedName) (*ConnSecretIdentifiers, error) {
```
josvazg: typo?

```suggestion
func (r *ConnSecretReconciler) identifiersFromInternalName(req types.NamespacedName) (*ConnSecretIdentifiers, error) {
```

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+		ProjectID:        parts[0],
+		ClusterName:      parts[1],
+		DatabaseUsername: parts[2],
+	}, nil
+}
+
+// indentifiersFromSecret loads the identifiers for the k8s format
+// === K8s format: <ProjectName>-<ClusterName>-<DatabaseUserName>
+// K8s secret must exists in the cluster
+func (r *ConnSecretReconciler) indentifiersFromK8s(ctx context.Context, req types.NamespacedName) (*ConnSecretIdentifiers, error) {
```
josvazg: typo?

```suggestion
func (r *ConnSecretReconciler) identifiersFromK8s(ctx context.Context, req types.NamespacedName) (*ConnSecretIdentifiers, error) {
```

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+// and extracts ProjectID, ClusterName, and DatabaseUsername.
+func (r *ConnSecretReconciler) loadIdentifiers(ctx context.Context, req types.NamespacedName) (*ConnSecretIdentifiers, error) {
+	if strings.Contains(req.Name, InternalSeparator) {
+		return r.identifiersFromInternalName(req)
+	}
+
+	return r.identifiersFromK8s(ctx, req)
+}
+
+// indetifiersFromInternalName loads the identifiers for the internal format
```
josvazg:
```suggestion
// identifiersFromInternalName loads the identifiers for the internal format
```

josvazg: nit: I have lots of typos myself, I find the IDE can help me with this with a spell checker. Setup depends on which IDE you maybe be using.

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+		return nil, ErrInternalFormatErr
+	}
+	return &ConnSecretIdentifiers{
+		ProjectID:        parts[0],
+		ClusterName:      parts[1],
+		DatabaseUsername: parts[2],
+	}, nil
+}
+
+// indentifiersFromSecret loads the identifiers for the k8s format
```
josvazg:
```suggestion
// identifiersFromSecret loads the identifiers for the k8s format
```

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+	return &ConnSecretIdentifiers{
+		ProjectID:        projectID,
+		ProjectName:      parts[0],
+		ClusterName:      clusterName,
+		DatabaseUsername: parts[1],
+	}, nil
+}
+
+// loadPair creates the paired resource that contains the parent AtlasDatabaseUser and the Endpoint.
+// Endpoint could be AtlasDeployment or AtlasDataFederation
```
josvazg, line 168: **comment**: I like the concept of an endpoint here, because we know connection secrets are always for a database user to access an resource/endpoint such as a deployment, etc. We will leverage this further with auto generation when we need to add yet more endpoints such as bare bones flex & clusters, which will be separate CRDs.

A bit more problematic will be when the atlas database user itself gets upgraded with a new autogenerated CRD. We will have 2 possible "shapes" for a user, but only one can be used at a time. I mean, you could have user "allen" defined in a new CRD, but then you should not let the old CRD about the same "allen" user in the same project lying around. That is a conflict even without connection secrets.

internal/controller/connectionsecret/connectionsecret.go
andrpac: Resolved! Thank you.

internal/controller/connectionsecret/connectionsecret.go
andrpac: That's exactly what I had in mind when I designed the Endpoint. I wanted to make your life easier when you move to autogenerated stuff (since I know probably CRDs will change)

But here's the thing. AtlasDatabaseUser could also be abstracted as an Endpoint type. And the nice part is that as long as you respect the interface, the reconciler logic should still work just fine. So all you have to do is implement the Endpoint interface for AtlaslDBUser (and sadly implement the logic for it). This will also give you a bit more flexibility in terms of what a Connection Secret is between.

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+		return "", ErrUnresolvedProjectName
+	}
+
+	var err error
+	var projectName string
+	// Try resolving from the Endpoint if present
+	if pair.Endpoint != nil {
+		projectName, err = pair.Endpoint.GetProjectName(ctx)
+		if projectName != "" {
+			return kube.NormalizeIdentifier(projectName), nil
```
josvazg, line 253: **q**: Why do we need to normalize here? Which "name" of the project are we using?

Atlas projects have a name, but the naming conventions are more relaxed than the Kubernetes name conventions AFAICR. You can create a Kubernetes compliant project name that works for Atlas but also have an Atlas name that does not work for Kubernetes. For instance, spaces and uppercase are allowed in Atlas, but not in Kubernetes, I believe.

But it in general the important bit here is that the Kubernetes project name is not the Atlas name and in general may not match it.

I am guessing we:
- DO need normalisation because we use an Atlas name that is made Kubernetes compatible via normalisation, so that it can be used as a prefix of the connection secret name.
- And the reason we take the Atlas name and not the Kubernetes name is because sometimes the project might not be defined in Kubernetes at all, as resources could refer to the project by Atlas ID instead. So the Kubernetes name might not always be available.

I am correct with this assumption here?

andrpac: You are perfectly correct. We are using Spec Name not Metadata name. And that is because, as you said, if your AtlasDeployment or AtlasDatabaseUser have an ExternalRef to a projectID, you do not get the metadata.name, but you can always get Spec.Name (with Atlas SDK).

josvazg: Make sense, thanks for explaining!

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+	log := r.Log.With("ns", req.Namespace, "name", req.Name)
+
+	// project name is required for metadata.name
+	projectName, err := r.resolveProjectName(ctx, ids, pair)
+	if err != nil {
+		log.Errorw("failed to resolve project name", "error", err)
+		return workflow.Terminate(workflow.ConnSecretUnresolvedProjectName, err).ReconcileResult()
+	}
+
+	name := CreateK8sFormat(projectName, ids.ClusterName, ids.DatabaseUsername)
```
josvazg, line 288: **q**: This name feels odd, how about this one instead?
```suggestion
	name := K8sConnectionSecretName(projectName, ids.ClusterName, ids.DatabaseUsername)
```

internal/controller/connectionsecret/connectionsecret.go
```diff
@@ -0,0 +1,458 @@
...
+			current := &corev1.Secret{}
+			if err := r.Client.Get(ctx, client.ObjectKeyFromObject(secret), current); err != nil {
+				log.Errorw("failed to fetch existing secret", "error", err)
+				return err
+			}
+			secret.ResourceVersion = current.ResourceVersion
+			if err := r.Client.Update(ctx, secret); err != nil {
+				log.Errorw("failed to update secret", "error", err)
+				return err
+			}
```
josvazg, line 386: **block**: Get + Update operations in Kubernetes are subject to optimistic locking conflicts, when it was too optimistic!

One way to avoid this is which works without retries is a **server side apply**:
https://kubernetes.io/docs/reference/using-api/server-side-apply/

There is an example of this being used in the e2e2 test for integrations:
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/test/e2e2/integration_test.go#L102

Note this usage is very recent. Most of our code, specially tests will probably follow a similar approach to this naive update. But going forward we can raise the reliability of our code by avoiding such error prone pattern and just use a tried and tested solution for it.

This change is not so trivial, so we can take this over next week.

andrpac: Thank you Jose for pointing this out. I had no idea this was even a thing (pretty cool ngl). I will let you guys take over from here! Best of luck!

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,330 @@
...
+		return r.handleDelete(ctx, req, ids, pair)
+	}
+
+	// Paired resource must be ready
+	if !(pair.User.IsDatabaseUserReady() && pair.Endpoint.IsReady()) {
+		log.Debugw("waiting on paired resource to be ready")
+		return workflow.InProgress(workflow.ConnSecretNotReady, "resources not ready").ReconcileResult()
+	}
+
+	return r.handleUpsert(ctx, req, ids, pair)
```
josvazg, line 135: **comment**: We might want to rewrite this part to make it more explicit this is the state machine bit of the reconciler.

internal/controller/connectionsecret/connectionsecret_controller.go
```diff
@@ -0,0 +1,330 @@
...
+
+func (r *ConnSecretReconciler) generateConnectionSecretRequests(projectID string, endpoints []Endpoint, users []akov2.AtlasDatabaseUser) []reconcile.Request {
+	var reqs []reconcile.Request
+	for _, ep := range endpoints {
+		for _, u := range users {
+			if !allowsByScopes(&u, ep.GetName(), ep.GetScopeType()) {
+				continue
+			}
+
+			name := CreateInternalFormat(projectID, ep.GetName(), u.Spec.Username)
```
josvazg, line 202: **q**: Anotehr odd name, what about this instead?

```suggestion
			name := NewConnectionSecretRequestName(projectID, ep.GetName(), u.Spec.Username)
```

internal/controller/connectionsecret/connectionsecret_controller_test.go
```diff
@@ -0,0 +1,580 @@
...
+					Username:        "admin",
+					DeleteAfterDate: time.Now().UTC().Add(-1 * time.Hour).Format(time.RFC3339),
+				},
+			},
+			expectedDeletion: true,
+			expectedResult: func() (ctrl.Result, error) {
+				return workflow.TerminateSilently(nil).WithoutRetry().ReconcileResult()
+			},
+		},
+		"requque: resources are not ready yet": {
```
josvazg, line 119:
```suggestion
		"re-enque: resources are not ready yet": {
```

test/int/databaseuser_unprotected_test.go
```diff
@@ -667,6 +668,122 @@ var _ = Describe("Atlas Database User", Label("int", "AtlasDatabaseUser", "prote
 			})
 		})
 
+		It("Adds connection secret for data federation", Label("datafederation-user-add-secret"), func() {
```
josvazg, line 671: **comment**: This is borderline to the policy of moving tests to e2e2 and being able to drop the old int and e2e test after we have migrated to autogenerated code.

I understand this is not a new new test, but a new case within an existing one. Also this is on the database user object that should be deprecated once replaced by the autogen version, so I guess it is fine like this.

internal/controller/connectionsecret/connectionsecret.go
s-urbaniak: filip: this is a good one to resolve as it is a blocker.

internal/controller/connectionsecret/connectionsecret.go
filipcirtog: This has been addressed in the other PR. Closing the conversation to highlight that the improvements have been implemented.

internal/controller/connectionsecret/connectionsecret.go
filipcirtog: This has been addressed in the other PR. Closing the conversation to highlight that the improvements have been implemented.

internal/controller/connectionsecret/connectionsecret_controller_test.go
filipcirtog: This has been addressed in the other PR. Closing the conversation to highlight that the improvements have been implemented.

internal/controller/connectionsecret/connectionsecret_controller.go
filipcirtog: This has been addressed in the latest PR. Closing the conversation to highlight that the improvements have been implemented.

PR #2580, opened 2025-08-12
CLOUDP-335193: Added organization settings CRD and controller

igor-karpukhin:
# Summary

This PR introduces a new CRD: AtlasOrgSettings and its controller.

## Checklist

  * **q**: Clarifying question.


api/v1/atlasorgsettings_types.go
```diff
@@ -0,0 +1,105 @@
...
+func init() {
+	SchemeBuilder.Register(&AtlasOrgSettings{})
+	SchemeBuilder.Register(&AtlasOrgSettingsList{})
+}
+
+type AtlasOrgSettingsSpec struct {
+	// OrgId Unique 24-hexadecimal digit string that identifies the organization that
+	// contains your projects
+	// +required
+	OrgID string `json:"orgID"`
```
andrpac, line 33: q: Isn't OrgID already part of ConnectionSecret. Is there a reason we make it explicit?

internal/controller/atlas/provider.go
```diff
@@ -47,6 +48,7 @@ type Provider interface {
 
 type ClientSet struct {
 	SdkClient20250312002 *admin.APIClient
+	SdkClient20250312006 *admin6.APIClient
```
andrpac: q: Is there a reason for still keeping the old client?

internal/indexer/atlasorgsettingssecrets.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (AtlasOrgSettingsByConnectionSecretIndexer) Object() client.Object {
+	return &akov2.AtlasProject{}
+}
+
+func (*AtlasOrgSettingsByConnectionSecretIndexer) Name() string {
+	return AtlasOrgSettingsBySecretsIndex
+}
+
```
andrpac, line 31: missing Keys() and indexer not registered with manager.

api/v1/status/atlasorgsettings.go
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package status
+
+type AtlasOrgSettingsStatus struct {
+	UnifiedStatus `json:",inline"`
```
andrpac, line 19: q: Do we update/change this anywhere?

api/v1/atlasorgsettings_types.go
igor-karpukhin: That's right. But this is done so that CRD contains explicit information about the OrgID it changes. Although it can be moved to status. We can discuss that.

internal/controller/atlas/provider.go
igor-karpukhin: Already removed the old one :)

api/v1/status/atlasorgsettings.go
igor-karpukhin: Yes, see the generic state machine

andrpac: Oh cool. Didn't see that.

internal/controller/atlas/provider.go
andrpac: Wow that's a lot of files :)

internal/controller/atlasorgsettings/handler.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+type reconcileContext struct {
+	svc atlasorgsettings.AtlasOrgSettingsService
+	aos *akov2.AtlasOrgSettings
+}
+
+func (h *AtlasOrgSettingsHandler) newReconcileContext(ctx context.Context, aos *akov2.AtlasOrgSettings) (*reconcileContext, error) {
+	var objKey *client.ObjectKey
+	if aos.Spec.ConnectionSecretRef != nil && aos.Spec.ConnectionSecretRef.Name != "" {
```
andrpac, line 38: Shouldn't we also check that the secret is ready? Like for example, if the user decides to create the OrgSettings and he mentions the ConnectionSecretRef. But he did not apply the secret yet. Should't we requeue then?

igor-karpukhin: In this case we use the global secret ref. This is a common practice in the operator. In case no secret is set - the AKO uses the global secret. If the secret is specified but it doesn't exist, the AKO will return a reconciliation error, that will automatically add this request back to queue.

internal/controller/atlas/provider.go
```diff
@@ -44,7 +45,8 @@ type Provider interface {
 }
 
 type ClientSet struct {
-	SdkClient20250312002 *admin.APIClient
+	SdkClient20250312002 *v20250312002.APIClient
+	SdkClient20250312006 *v20250312006.APIClient
```
josvazg, line 49: q: You mention there was something in ...006 not present in 002? also if moving everything to 006 it broke something else?

internal/controller/atlasorgsettings/handler.go
```diff
@@ -0,0 +1,107 @@
...
+
+	akov2 "github.com/mongodb/mongodb-atlas-kubernetes/v2/api/v1"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/controller/reconciler"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/internal/translation/atlasorgsettings"
+	ctrlstate "github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/controller/state"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/result"
+	"github.com/mongodb/mongodb-atlas-kubernetes/v2/pkg/state"
+)
+
+type reconcileContext struct {
```
josvazg: **nit**: Request I think is a better name for this, which does not clash with the go context:

```suggestion
type reconcileRequest struct {
```

internal/controller/atlasorgsettings/handler.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return &reconcileContext{
+		svc: h.serviceBuilder(atlasSdk),
+		aos: aos,
+	}, nil
+}
+
+func (h *AtlasOrgSettingsHandler) upsert(ctx context.Context, currentState, nextState state.ResourceState,
+	aos *akov2.AtlasOrgSettings) (ctrlstate.Result, error) {
+	reconcileCtx, err := h.newReconcileContext(ctx, aos)
```
josvazg: **nit**: this is was named a request we could have this as:
```suggestion
	req, err := h.newReconcileRequest(ctx, aos)
```

no confusion between `ctx` and `req`

internal/controller/atlasorgsettings/handler.go
```diff
@@ -0,0 +1,107 @@
...
+		resp, apiErr := reconcileCtx.svc.Update(ctx, aos.Spec.OrgID, desiredSettings)
+		if apiErr != nil {
+			return result.Error(currentState, apiErr)
+		}
+		if resp == nil {
+			return result.Error(currentState, fmt.Errorf("atlas returned OrgSettings which is nil after update"))
+		}
+
+		return result.NextState(nextState, "Updated")
+	}
```
josvazg, line 84: **q**: This method is called upsert, but it only does updates.

I am guessing the settings always pre-exist, as long the org exists, meaning there is never an insert to be done.

Should we not rename this method as simple `update`?

internal/controller/atlasorgsettings/handler.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return result.NextState(nextState, "Ready")
+}
+
+func (h *AtlasOrgSettingsHandler) unmanage(orgID string) (ctrlstate.Result, error) {
+	return result.NextState(state.StateDeleted, fmt.Sprintf("unmanaged is AtlasOrgSettings for orgID %s.", orgID))
```
josvazg:
```suggestion
	return result.NextState(state.StateDeleted, fmt.Sprintf("unmanaged AtlasOrgSettings for orgID %s.", orgID))
```

internal/controller/atlasorgsettings/handler_test.go
```diff
@@ -0,0 +1,711 @@
...
+						MultiFactorAuthRequired:                sampleAtlasOrgSettings.Spec.MultiFactorAuthRequired,
+						RestrictEmployeeAccess:                 sampleAtlasOrgSettings.Spec.RestrictEmployeeAccess,
+						SecurityContact:                        sampleAtlasOrgSettings.Spec.SecurityContact,
+						StreamsCrossGroupEnabled:               sampleAtlasOrgSettings.Spec.StreamsCrossGroupEnabled,
+					},
+				}, nil, nil, nil, false), // No update expected
+			input:   &sampleAtlasOrgSettings,
+			objects: []client.Object{&fakeAtlasSecret},
+			want: ctrlstate.Result{
+				NextState: "Created",
```
josvazg, line 381: I am not sure this resource should have the concept of **Created**, it seems to me it is more like an IP access list kind of thing:
https://docs.google.com/presentation/d/1j1A7uF72C672Ct766O-ozj_BAwBFpRazdtTxb5o6Oag/edit?slide=id.g35221247c10_0_81#slide=id.g35221247c10_0_81

Only relevant state machines states are:
**Initial**
**Updated**
**Import Requested**
**Imported**
**DeletionRequested** (which is only for the Kubernetes side of things, as it does not delete it from Atlas, but rather un manages it as the settings will exist so long as the org does)
**Deleted**

BTW this is an interesting corker case in which deletion protection is always ON, right?

internal/controller/atlasorgsettings/atlasorgsettings_controller.go
```diff
@@ -0,0 +1,112 @@
...
+// +kubebuilder:rbac:groups=atlas.mongodb.com,namespace=default,resources=atlasorgsettings,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=atlas.mongodb.com,namespace=default,resources=atlasorgsettings/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=atlas.mongodb.com,namespace=default,resources=atlasorgsettings/finalizers,verbs=update
+
+type serviceBuilderFunc func(*atlas.ClientSet) atlasorgsettings.AtlasOrgSettingsService
+
+type AtlasOrgSettingsHandler struct {
+	ctrlstate.StateHandler[akov2.AtlasOrgSettings]
+	reconciler.AtlasReconciler
+	deletionProtection bool
```
josvazg: **q**: I think for this resource deletion protection makes no real sense: there is no way to remove the settings, they exist so long as the org does. And whether the org exists or not is not for this controller to handle.

Shouldn't we just remove this then from here?

internal/indexer/atlasorgsettingssecrets.go
```diff
@@ -0,0 +1,38 @@
...
+const (
+	AtlasOrgSettingsBySecretsIndex = "atlasorgsettings.spec.connectionSecretRef"
+)
+
+func NewAtlasOrgSettingsByConnectionSecretIndexer(logger *zap.Logger) *LocalCredentialIndexer {
+	return NewLocalCredentialsIndexer(AtlasOrgSettingsBySecretsIndex, &akov2.AtlasOrgSettings{}, logger)
+}
+
+func AtlasOrgSettingsRequest(list *akov2.AtlasOrgSettingsList) []reconcile.Request {
```
josvazg, line 32: **comment**: Yesterday I learn from @andrpac that in kube libs have [ExtractList](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#ExtractList) which might be able to make this manual conversion on every indexer go away.

But that would be a separate piece of work.

internal/translation/atlasorgsettings/atlasorgsettings.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (a *AtlasOrgSettings) Equal(other *AtlasOrgSettings) bool {
+	if other == nil {
+		return false
+	}
+
+	return a.ApiAccessListRequired != nil && other.ApiAccessListRequired != nil &&
+		*a.ApiAccessListRequired == *other.ApiAccessListRequired &&
```
josvazg, line 39:
```suggestion
	return pointer.Equal(a.ApiAccessListRequired, other.ApiAccessListRequired) &&
```

Where in `pointer`:
```go
func Equal[T comparable](a, b *T) bool {
  return a != nil && b!= nil && *a == *b
}
```

But I even wonder why deep equal would not work here, or at least work after a normalization.

internal/translation/atlasorgsettings/service.go
```diff
@@ -0,0 +1,65 @@
...
+func NewAtlasOrgSettingsService(api admin.OrganizationsApi) AtlasOrgSettingsService {
+	return &AtlasOrgSettingsServiceImpl{
+		orgSettingsAPI: api,
+	}
+}
+
+func (a *AtlasOrgSettingsServiceImpl) Get(ctx context.Context, orgID string) (*AtlasOrgSettings, error) {
+	resp, httpResp, err := a.orgSettingsAPI.GetOrganizationSettings(ctx, orgID).Execute()
+	if err != nil {
+		return nil, err
```
josvazg: **block**: It is better to wrap all error returns:

```suggestion
		return nil, fmt.Errorf("failed to get org settings: %w", err)
```

internal/translation/atlasorgsettings/service.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	if httpResp.StatusCode != 200 {
+		return nil, err
```
josvazg: **block**: This will return a `nil, nil` which I believe is ambiguous in this context.

```suggestion
		return nil, fmt.Errorf("bad response status when getting org settings: expected 200 but got %d", httpResp.StatusCode)
```

Or we might want to use the status message text rather than the code value.

internal/translation/atlasorgsettings/service.go
```diff
@@ -0,0 +1,65 @@
...
+	if atlasOrgSettings == nil {
+		return nil, nil
+	}
+
+	resp, httpResp, err := a.orgSettingsAPI.UpdateOrganizationSettings(ctx, orgID, atlasOrgSettings).Execute()
+	if err != nil {
+		return nil, err
+	}
+	if httpResp.StatusCode != 201 && httpResp.StatusCode != 200 {
+		return nil, err
```
josvazg: **block**: same issues with these error returns, they should be wrapped and the second one would be `nil` otherwise.

internal/controller/atlasorgsettings/handler.go
igor-karpukhin: Yes, there is no "create" api. I called it `upsert` to stick to the way controllers are written (e.g. AtlasThirdPartyIntegrations)

internal/controller/atlas/provider.go
igor-karpukhin: Yes, having only 006 breaks other features. I rolled back my changes to have two clients. This is not critical, we can update SDK later.

internal/controller/atlasorgsettings/handler_test.go
igor-karpukhin: We do have this state: https://github.com/mongodb/mongodb-atlas-kubernetes/blob/4cdcc17fbb725bd3eab17bc5925ac91d3df9c94d/pkg/state/state.go#L38. But I'm just thinking, since OrgSettings are already there, they can only be "Updated".

internal/translation/atlasorgsettings/atlasorgsettings.go
igor-karpukhin: A `DeepEqual` can be used, but it would require to copy the structure and to cleanup fields that are in AKO spec and not in Atlas. I'd leave it as is.

internal/translation/atlasorgsettings/service.go
igor-karpukhin: Good point, fixed

internal/translation/atlasorgsettings/service.go
igor-karpukhin: Good catch! fixed

internal/translation/atlasorgsettings/service.go
igor-karpukhin: Fixed as well

internal/controller/atlasorgsettings/atlasorgsettings_controller.go
igor-karpukhin: We can keep it or leave it. I think we can drop it, since there is no way to delete the AtlasOrgSettings. Good point. Let me remove it

internal/controller/atlasorgsettings/handler.go
igor-karpukhin: Thanks, fixed

internal/controller/atlasorgsettings/handler.go
igor-karpukhin: Ok, renamed

internal/controller/atlasorgsettings/handler.go
igor-karpukhin: Ok, renamed

internal/translation/atlasorgsettings/atlasorgsettings.go
josvazg: This has an easy fix BTW, struct embedding, and it has been used before in other resources:

1) Define a struct with only the fields that match the ones in Atlas.
2) Embed that struct into the one that has the AKO spec (usually the AKO spec itself).

When comparing, compare only the inner embedded spec.

helderjs (approved):
LGMT once CI is green and remaining comments are resolved

internal/controller/atlasorgsettings/handler.go
josvazg: **block**: I see, but the issue is that there is no creation for this one, unlike with AtlasThirdPArtyIntegrations. As there is no creation, there should be no insert and there should not be even a `HandleCreate` at all, it should just be calling `update` from `HandleInitial` & `HandleUpdate`

As I mentioned elsewhere this is a case with only the following states:

- **Initial**
- **Updated**
- **Import Requested**
- **Imported**
- **Deletion requested**
- **Deleted**

Most similar sample I think is IP access list: https://docs.google.com/presentation/d/1j1A7uF72C672Ct766O-ozj_BAwBFpRazdtTxb5o6Oag/edit?slide=id.g35221247c10_0_81#slide=id.g35221247c10_0_81

(Given than the org id always exists and is known at credential setting time, for this one even the import cases might not even matter. Which is already reflected in the current code.)

josvazg: Talked offline and I was probably wrong here. The states are regarding the Kubernetes sided of things, so indeed we do create the kube handler of the resource, even if the resource already exists.

So I guess the current code is correct then.

josvazg (approved):

PR #2596, opened 2025-08-19
CLOUDP-334946: Allow contributions to run tests when authorized

josvazg:
# Summary

Changing the trigger event from  the event from `pull_request` to [pull_request_target](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) we grant contributors the ability to test using AKO credentials. This should be still safe as the job for this workflow only runs when:
- The PR comes from the AKO repo itself.
- Or the PR has been labeled in a certain way, which is limited to AKO repo owners.

The idea is that contributions should run no tests by default, but then will run all tests normally once we have reviewed the code and deemed it safe to test, adding the `safe-to-test` label.

The safer `pull_request` event triggering is removed to avoid this workflow being triggered twice with confusing results.
 
*UPDATE*
As pointed by @avivkeller, this PR also changes the check to ensure the event triggered was setting one of the expected labels. Otherwise, once a label was set, a malicious contributor could send an update that would run the tests before the change could be examined. The remove labels workflow seemed to exist only for this, so it was also removed as it is no longer needed after this check fix.

Note this means, every contributor PR update would need to be reviewed and labeled to run tests.

## Proof of Work

TBD: will test also on a fork repo.

## Checklist


avivkeller:
A safer check would be checking that `github.event.label.name` is the expected label, since, this will only trigger _when_ the label is added, not all events following

josvazg:
> A safer check would be checking that `github.event.label.name` is the expected label, since, this will only trigger _when_ the label is added, not all events following

Thanks a lot @avivkeller ! That is a very good hint!
I will update the code so that it only runs when the label is set, not just when it is present.

s-urbaniak (approved):

PR #2604, opened 2025-08-20
CLOUDP-339233: Fix v2.10.0 tag to update deploy and helm-charts

josvazg:
# Summary

Update the `deploy` and `helm-charts` in the v2.10.0 tag, as it was expected by the **helm chart automation** and the **cli plugin** to have been done in such release.

The **helm chart automation** expects AKO to be the source of truth for the helm charts code. Otherwise it will issue PRs against the helm-charts repo to update. This was already fixed in `main` but is also fixed in the tag for consistency.

The **cli plugin** would use the version tag to access the `deploy` directory and use its YAMLs to install the operator. The expectation is the version of the operator image that tag code refers to is that of the same tag, but right now it was stale at `2.9.1` instead.

Commands executed to set these changes where:
```shell
cp -r releases/v2.10.0/deploy .
cp -r releases/v2.10.0/helm-charts .
```

#### UPDATE

Not sure why the check license headers is failing here, but given that this is just to fix a tag and that this was not adding nor removing Go code, I suggest to ignore such issue for the review.

This must merge and replace the tag after merging #2605 

## Proof of Work

Install from the CLI plugin for 2.10.0 should work without further fixed once this replaces the current `v2.10.0` tag.

## Checklist

### DIFF
--- a/helm-charts/atlas-operator-crds/Chart.yaml
+++ b/helm-charts/atlas-operator-crds/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: mongodb-atlas-operator-crds
 description: MongoDB Atlas Operator CRDs - Helm chart for installing and upgrading Custom Resource Definitions (CRDs) for the Atlas Operator.
 type: application
-version: v2.10.0
-appVersion: v2.10.0
+version: 2.10.0
+appVersion: 2.10.0
 kubeVersion: ">=1.15.0-0"
 keywords:
   - mongodb
--- a/helm-charts/atlas-operator/Chart.yaml
+++ b/helm-charts/atlas-operator/Chart.yaml
@@ -3,8 +3,8 @@ name: mongodb-atlas-operator
 description: |-
   MongoDB Atlas Operator - a Helm chart for installing and upgrading Atlas Operator: the official Kubernetes operator allowing to manage MongoDB Atlas resources from Kubernetes
 type: application
-version: v2.10.0
-appVersion: v2.10.0
+version: 2.10.0
+appVersion: 2.10.0
 kubeVersion: ">=1.15.0-0"
 keywords:
   - mongodb
@@ -19,6 +19,6 @@ maintainers:
     email: support@mongodb.com
 dependencies:
   - name: mongodb-atlas-operator-crds
-    version: "v2.10.0"
+    version: "2.10.0"
     repository:  "file://../../helm-charts/atlas-operator-crds"
     condition: mongodb-atlas-operator-crds.enabled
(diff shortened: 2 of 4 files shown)


josvazg:
Not sure why the check license headers is failing here, but given that this is just to fix a tag and that this was not adding nor removing Go code, I suggest to ignore such issue for the review.

deploy/all-in-one.yaml
```diff
@@ -6874,7 +6886,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-        image: mongodb/mongodb-atlas-kubernetes-operator:2.9.1
+        image: docker.io/mongodb/mongodb-atlas-kubernetes-operator:v2.10.0
```
igor-karpukhin: **nit:** docker.io is a default one for using docker.

.gitignore
```diff
@@ -40,9 +40,6 @@ test/e2e/output/
 node_modules
 tmp/
 
-# ready to work samples
-deploy/
```
igor-karpukhin: Please make sure you add it back, otherwise there is a risk that someone may commit their local artefacts

deploy/clusterwide/clusterwide-config.yaml
```diff
@@ -228,7 +228,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-        image: mongodb/mongodb-atlas-kubernetes-operator:2.9.1
+        image: docker.io/mongodb/mongodb-atlas-kubernetes-operator:v2.10.0
```
igor-karpukhin: **blocking**: here as well, no `v` prefix

deploy/all-in-one.yaml
igor-karpukhin: **blocking**: no `v` prefix for image tags

josvazg: 🤦‍♂️ this means the release directory is wrong as well! and the code that generated it!

.gitignore
josvazg: This is a bit problematic, it also means we need to add `-f` to the automation when making changes.

But I see what you mean for the development use case.

deploy/all-in-one.yaml
josvazg: ok, seems it was fixed later on in main, I will just fix the occurrences here in the tag.

.gitignore
josvazg: will revert this change, it is not in main, so the force will be need to be done anyways

andrpac:
I think you might need a rebase to fix the missing license-header-check

deploy/all-in-one.yaml
josvazg: fixed now

.gitignore
josvazg: done

deploy/clusterwide/clusterwide-config.yaml
josvazg: fixed now

PR #2639, opened 2025-09-01
CLOUDP-342256: Add predictiveAutoscaling Field

roothorp:
# Summary

Add field to Deployment Autoscaling to enable/disable predictive autoscaling

## Checklist

  * **q**: Clarifying question.


roothorp:
> Also I wonder if we should get this PR already to have the bundles updated as well.

@josvazg I did `make bundle` but there were many changes outside the scope of this change, so didn't commit it. I can add it if you'd prefer, but it looks like this has been missed in previous commits.

s-urbaniak (approved):

internal/controller/atlasdeployment/advanced_deployment_test.go
```diff
@@ -949,6 +949,130 @@ func TestHandleAdvancedDeployment(t *testing.T) {
...
 			},
 		},
+		"enable autoscaling with predictive autoscaling": {
+			atlasDeployment: &akov2.AtlasDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "cluster0",
+					Namespace: "test",
+				},
+				Spec: akov2.AtlasDeploymentSpec{
+					UpgradeToDedicated: true,
```
josvazg: **q**: Are we testing the migration at the same time?
I know this is a unit test, but I'd rather not introduce the noise if the focus of the test is not on the migration.

roothorp: Good catch; fixed!

igor-karpukhin (approved):

PR #2640, opened 2025-09-01
CLOUDP-340201: Fixing escaped credentials

filipcirtog:
# Summary

Data Federation was failing to create connection secrets when the password contained special escaped characters (e.g., P@sword%): _error":"parse \"mongodb://example-pr-user:P@sword%@example-pr.mongodb.net?ssl=true\": invalid URL escape \"%\""_

## Root Cause
The issue originated from the following problematic code snippet
`connURLs = append(connURLs, fmt.Sprintf("mongodb://%s:%s@%s?ssl=true", dbUser.Spec.Username, password, host))` In this code, the password was injected directly into the connection string using fmt.Sprintf without sanitizing or encoding special characters. This resulted in an invalid URL when the password contained characters such as %.

## Proof of Work
Ensured consistency by using methods that are already used by other controllers who parse escaped password. 
The issue was resolved by introducing proper handling of credentials via a helper function AddCredentialsToConnectionURL. This function uses url.Parse to sanitize the connection URL and adds proper credential encoding with url.UserPassword.

## Evidence of Successful Secret Creation with Passwords Lacking Special Characters
Screenshot 2025-09-01 at 13 13 37

## Checklist

### DIFF
--- a/internal/controller/atlasdatafederation/connectionsecrets_test.go
+++ b/internal/controller/atlasdatafederation/connectionsecrets_test.go
@@ -21,7 +21,6 @@ import (
 )
 
 func TestAddCredentialsToConnectionURL(t *testing.T) {
-
 	testCases := []struct {
 		name          string
 		connURL       string


internal/controller/atlasdatafederation/connectionsecrets.go
```diff
@@ -76,7 +76,13 @@ func (r *AtlasDataFederationReconciler) ensureConnectionSecrets(ctx *workflow.Co
 
 		var connURLs []string
 		for _, host := range connectionHosts {
-			connURLs = append(connURLs, fmt.Sprintf("mongodb://%s:%s@%s?ssl=true", dbUser.Spec.Username, password, host))
+			connURL, err := connectionsecret.AddCredentialsToConnectionURL(fmt.Sprintf("mongodb://%s?ssl=true", host), dbUser.Spec.Username, password)
```
josvazg: **nit**: This line does too much to my taste, what would you think of this instead?
```suggestion
			baseURL := fmt.Sprintf("mongodb://%s?ssl=true", host)             
			connURL, err := connectionsecret.AddCredentialsToConnectionURL(baseURL, dbUser.Spec.Username, password)
```

This makes it clearer what is constructing the base URL before it is further processed by `AddCredentialsToConnectionURL`, and the args for the `Sprintf` and the `AddCredentialsToConnectionURL` don't get mixed up.

internal/controller/atlasdatafederation/connectionsecrets.go
```diff
@@ -116,3 +122,12 @@ func dbUserBelongsToProject(dbUser *akov2.AtlasDatabaseUser, project *akov2.Atla
 
 	return true
 }
+
+func AddCredentialsToConnectionURL(connURL, userName, password string) (string, error) {
```
josvazg, line 127: I would at least add a unit test for this new method. Testing the whole `ensureConnectionSecrets` would be an overkill for this fix, but testing just `AddCredentialsToConnectionURL` should be easy and what is not covered in that unit tests should be covered by existing e2e tests.

josvazg: BTW, for the unit test, you can use a single test with several input/outputs to try out. There are several of those in the codebase already. The idea would be to try normal and as well as problematic inputs generate the expected results.

internal/controller/atlasdatafederation/connectionsecrets.go
filipcirtog: Nice suggestion! Thanks.

internal/controller/atlasdatafederation/connectionsecrets.go
filipcirtog: I initially tried testing ensureConnectionSecrets, but it was too complex. As per your suggestion, I added a simple unit test for AddCredentialsToConnectionURL. Thanks!

internal/controller/atlasdatafederation/connectionsecrets_test.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+func TestAddCredentialsToConnectionURL(t *testing.T) {
+	testCases := []struct {
+		name      string
+		connURL   string
+		userName  string
+		password  string
+		expected  string
+		expectErr bool
```
josvazg: **nit**: instead of expecting true/false you could use a string and check the expected error contains some string, or when empty that no error was expected

josvazg (approved):
LGTM!

PR #2593, opened 2025-08-19
CLOUDP-338956: Update Helm Charts & Deploy on Release PR

josvazg:
# Summary

Make sure Operator and CRD Helm Charts image references in the AKO repository get updates to the image version being released.

## Proof of Work

✅ [Pre-release test generated the deploy and helm](https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2614) **(NOT MERGED)**

## Checklist
- [X] Have you linked a jira ticket and/or is the ticket in the title?
- [X] Have you checked whether your jira ticket required DOCSP changes?
- [X] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?

PR #2652, opened 2025-09-03
Make Release Image not update deploy or bundle in main

josvazg:
# Summary

This undoes fix from CLOUDP-338956. As the CLI is now fixed to use the release directory of each version, instead of the version tag, and it has been fixed retroactively fro `2.9` and `2.8` next release should happen after ir with the CLI fix in place and therefore we do not need to copy the deploy dir to `main`.

Note this PR changes nothing about `helm-charts`. That cannot be changed until we revisit the way the helm charts dir syncs with the helm charts, as automation is set so that this repository is the source of truth over Atlas helm charts.

## Proof of Work

✅ [Release image run properly for a pre-release dryrun](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/17435347186/job/49503557214)
✅ [PR #2653 is the pre-release PR](https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2653/files) Shows only the release dir and the `helm-charts`.

## Checklist

### DIFF
--- a/.github/workflows/release-image.yml
+++ b/.github/workflows/release-image.yml
@@ -257,11 +257,10 @@ jobs:
           git checkout -f -b "$BRANCH" origin/main
           git push -f origin "$BRANCH"
 
-          # Update Helm Charts & Deploy on main
+          # Update Helm Charts on main
           cp -r "$RELEASE_DIR/helm-charts" .
-          cp -r "$RELEASE_DIR/deploy" .
 
-          git add -f "$RELEASE_DIR" helm-charts deploy
+          git add -f "$RELEASE_DIR" helm-charts
           scripts/create-signed-commit.sh
 
           gh pr create \


s-urbaniak (approved):

.github/workflows/release-image.yml
```diff
@@ -257,11 +257,10 @@ jobs:
           git checkout -f -b "$BRANCH" origin/main
           git push -f origin "$BRANCH"
 
-          # Update Helm Charts & Deploy on main
+          # Update Helm Charts on main
           cp -r "$RELEASE_DIR/helm-charts" .
-          cp -r "$RELEASE_DIR/deploy" .
```
igor-karpukhin, line 262: Were do these generated now then? Previously they where copied from `./bundle` and `./deploy` folders that updated after `make manifests && make bundle`

josvazg: Let me check, but this was not changed by either this fix nor the previous fix at CLOUDP-338956

see #2593

igor-karpukhin (approved):

.github/workflows/release-image.yml
josvazg: For completeness, step "Generate deployment configurations" uses action `gen-install-scripts` to update the bundle and deploy dirs, just as always.

PR #2668, opened 2025-09-09
CLOUDP-338824: Generate licenses only for releases

igor-karpukhin:
# Summary

* Generate licenses only for releases
* For every PR validate that we don't violate licenses policy

## Checklist

  * **q**: Clarifying question.


.github/workflows/release-branch.yml
```diff
@@ -56,6 +56,12 @@ jobs:
       run: |
         devbox run -- 'VERSION=${{ env.VERSION }} ./scripts/bump-helm-chart-version.sh'
 
+    - name: Generate licenses.csv
+      run: |
+        devbox run -- 'make build-licenses.csv'
+        git add .licenses-gomod.sha256
+        git add licenses.csv
```
josvazg, line 63: why add them to git at all?
I would expect these might ONLY show in a `releases/vX.Y.Z`

josvazg: BTW I think we need to fix `release-image` instead, this workflow is deprecated.

josvazg (changes requested):
The workflow changed has to be `release-image.yml`, release branch is deprecated now.

Also we do not want the licenses in their original git position, but tied to a particular release under `release/vX.Y.Z`
 
Apart from that, we might need more cleanups, maybe, but this is the minimum we need for now.

josvazg:
> The workflow changed has to be `release-image.yml`, release branch is deprecated now.
> 
> Also we do not want the licenses in their original git position, but tied to a particular release under `release/vX.Y.Z`
> 
> Apart from that, we might need more cleanups, such as deactivating the **ako-release** bot fixing the dependabot updates. Probably removing `update-licenses.yml` is enough.

We should probably remove the deprecated workflows at some point, we could discuss about it.

.github/workflows/release-branch.yml
igor-karpukhin: Ok, thanks. See the updated "release-image" job.

.github/workflows/release-image.yml
```diff
@@ -252,6 +256,9 @@ jobs:
           cp -rf bundle "$RELEASE_DIR/"
           cp -rf helm-charts "$RELEASE_DIR/"
           cp bundle.Dockerfile "$RELEASE_DIR/bundle.Dockerfile"
+          cp .licenses-gomod.sha256 "$RELEASE_DIR/"
```
josvazg: this file is no longer useful, it was only used to notice that the licenses needed to be re-updated.
we can remove it with all the logic about it. And we do not need to copy it to the release dir.

```suggestion
```

Makefile
```diff
@@ -178,14 +178,9 @@ build-licenses.csv: go.mod ## Track licenses in a CSV file
 recompute-licenses: ## Recompute the licenses.csv only if needed (gomod was changed)
 	@[ "$(GOMOD_SHA)" == "$(GOMOD_LICENSES_SHA)" ] || $(MAKE) build-licenses.csv
```
josvazg: This can go away as well now.

I would remove all usages and mentions of `GOMOD_LICENSES_SHA`

josvazg (commented):
I little bit more cleanup would be better, specifically around `GOMOD_LICENSES_SHA` in the Makefile, that file is no longer useful

.github/workflows/release-image.yml
igor-karpukhin: Ok, removed

Makefile
igor-karpukhin: Ok, I removed this makefile target

josvazg (approved):

PR #2709, opened 2025-09-19
CLOUDP-345379: File-based release version control

helderjs:
# Summary

## Proof of Work

Release version is controlled by the file version.json.
Compiled binary should contain the correct semver version.

## Checklist

  * **q**: Clarifying question.


.github/actions/build-push-image/action.yaml
```diff
@@ -1,33 +1,22 @@
 name: Build Container Image
 description: Builds the operator container image for the given architecture
 inputs:
-  get-existing-cache:
-    description: 'Get cache from the previous build'
-    required: false
-    default: 'false'
```
josvazg, line 7: why are these and the below inputs not required now?

.github/workflows/openshift-upgrade-test.yaml
```diff
@@ -33,9 +33,6 @@ jobs:
         with:
           ref: ${{github.event.pull_request.head.sha}}
           fetch-depth: 0 # required for tags
-      - name: Prepare tag
-        id: prepare
-        uses: ./.github/actions/set-tag
```
josvazg, line 38: **q**: what does have to do with the fix?

.github/actions/build-push-image/action.yaml
helderjs: All the dropped inputs were not used by the action, they were just polluting and make hard to understand the flow.

josvazg (commented):
There are some changes that I do not understand how they are related.

Also I guess the PR is not crucial to ensure we are really releasing the version we expected to release, right? I mean it is not asked as an input now.

.github/workflows/openshift-upgrade-test.yaml
helderjs: Not used, it was just creating confusion. Like, if it was really using the action, it would be impacted by the change.

version.json
```diff
@@ -0,0 +1,4 @@
+{
+  "current": "2.10.0",
+  "next": "2.11.0"
+}
```
s-urbaniak, line 4: we have a `major-version` file as well https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/major-version.

how do we make sure those stay in sync?

josvazg: yes, with this new file the old `major-version` should die and instead take the major version by parsing this file.

Something like this would work:
```shell
cat version.json | jq .current | tr -d '"' |awk -F. '{print $1}'
```

We could also do it on a followup, but I would create such ticket right away and link it from this PR.

docs/dev/release.md
```diff
@@ -332,10 +339,12 @@ And use the URL there to set `OPENSHIFT_UPGRADE_SERVER_API` so that openshift up
...
+1.  **Prevent Mismatches**: It stops the workflow if the version being released (e.g., `v3.0.0`) has a different major version than what the codebase expects (e.g., `current` is `v2.11.0`).
+2.  **Avoid Incorrect Patching**: It stops you from accidentally trying to release a patch for an older major version (e.g., `v1.2.3`) when the codebase has already moved on to a new major version.
+3.  **Skip Tests**: Certain tests that are expected to fail, like Helm upgrades, they should be skipped.
 
-If the create release branch job fails due an error such as `Bad major version for X... expected Y..`, review whether or not the `major-version` file was updated as expected. Check as well you are not trying to release a patch for the older major version from the new major version codebase.
+If the "Create Release Branch" job fails with an error like `Bad major version for X... expected Y...`, you should review the `current` field in `version.json`. Ensure it correctly reflects the codebase's state and is compatible with the version you intend to release.
```
josvazg, line 350: I think this is stale, the create release branch is now an obsolete step. The only action to use now is `Release Image`.

I think we can still make the check before the actual release happens, but note the release image created PR is an **"after the fact" review** of the already happened release.

docs/dev/release.md
```diff
@@ -23,6 +23,16 @@ At least **one** (1) week before the release the Kubernetes Version testing matr
 
 Please refer to the [CI documentation](ci.md#kubernetes-version-matrix) and submit a pull request, example: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2161 or https://github.com/mongodb/mongodb-atlas-kubernetes/pull/2082.
 
+## A Note on Versioning with `version.json`
+
+The `version.json` file is the definitive **source of truth** for managing software versions in this project.
+
+* **Source of Truth**: The file contains two primary fields: **`current`**, which reflects the latest stable version that has been released, and **`next`**, which designates the version targeted for the upcoming release.
+
+* **Automatic Updates for next Releases**: After a successful release, the CI pipeline will **automatically** update `version.json`. The `current` field is set to the version that was just released, and the `next` field is incremented to the next minor version (e.g., `2.11.0` would become `2.12.0`).
```
josvazg: **nit**: I would also make an example in case latest release is a patch:
```suggestion
* **Automatic Updates for next Releases**: After a successful release, the CI pipeline will **automatically** update `version.json`. The `current` field is set to the version that was just released, and the `next` field is incremented to the next minor version (e.g., `2.11.0` would become `2.12.0`, or `2.11.4` would become `2.12.0`).
```

Makefile
```diff
@@ -648,4 +654,14 @@ tools/scandeprecation/scandeprecation: tools/scandeprecation/*.go
...
+
+.PHONY: bump-version-file
+bump-version-file:
+	@echo "Bumping version in $(VERSION_FILE)..."
+
+	jq ' (.next | split(".") | map(tonumber) | .[1] += 1 | .[2] = 0 | map(tostring) | join(".")) as $$new_next | .current = .next | .next = $$new_next \
+	' $(VERSION_FILE) > $(VERSION_FILE).tmp && mv $(VERSION_FILE).tmp $(VERSION_FILE)
+
+	@echo "Version updated successfully:"
+	@cat $(VERSION_FILE)
```
josvazg, line 666: **q**: is this working for you?

```shell
% make bump-version-file  
Bumping version in version.json...
jq ' (.next | split(".") | map(tonumber) | .[1] += 1 | .[2] = 0 | map(tostring) | join(".")) as $new_next | .current = .next | .next = $new_next \
' version.json > version.json.tmp && mv version.json.tmp version.json
jq: error: syntax error, unexpected INVALID_CHARACTER, expecting end of file at <top-level>, line 1, column 142:
     (.next | split(".") | map(tonumber) | .[1] += 1 | .[2] = 0 | map(tostring) | join(".")) as $new_next | .current = .next | .next = $new_next \
                                                                                                                                                 ^
jq: 1 compile error
```

helderjs: Good catch! I tried to break the lines, but seems it breaks execution. Fixed.

docs/dev/release.md
helderjs: Applied

josvazg:
LGTM, we can revisit the docs separately.

Good to merge

PR #2720, opened 2025-09-24
CLOUDP-335401: Checkout the repo to check the version file

josvazg:
# Summary

Add the checkout so it won't fail like this:
https://github.com/mongodb/mongodb-atlas-kubernetes/actions/runs/17976950312/job/51132688395

## Proof of Work

CI should work but test can only be done once merged in main.

## Checklist

### DIFF
--- a/.github/workflows/release-image.yml
+++ b/.github/workflows/release-image.yml
@@ -88,6 +88,11 @@ jobs:
           echo "Setting pre-release repo mongodb/mongodb-atlas-kubernetes-operator-prerelease"
           echo "repo=mongodb/mongodb-atlas-kubernetes-operator-prerelease" | tee -a $GITHUB_OUTPUT
         fi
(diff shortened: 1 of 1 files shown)


.github/workflows/release-image.yml
```diff
@@ -88,6 +88,11 @@ jobs:
           echo "Setting pre-release repo mongodb/mongodb-atlas-kubernetes-operator-prerelease"
           echo "repo=mongodb/mongodb-atlas-kubernetes-operator-prerelease" | tee -a $GITHUB_OUTPUT
         fi
+    - name: Checkout code
```
helderjs, line 91: Shouldn't it be the first step? just in case...

josvazg: the first step does not need the repo files, it is just an if

PR #2763, opened 2025-10-06
Simplify translate API

josvazg:
# Summary

This makes the translator easier to use and more reusable. It is immutable (readonly) and can be used from different go routines.

## Proof of Work

CI passes

## Checklist


internal/v3/translate/translator.go
```diff
@@ -31,29 +31,23 @@ type PtrClientObj[T any] interface {
 	client.Object
 }
 
-// Translator allows to translate back and forth between a CRD schema version
+// Translator allows to translate back and forth between a CRD schema
 // and SDK API structures of a certain version
```
s-urbaniak: describe whether this is goroutine safe

internal/v3/translate/translator.go
```diff
@@ -31,29 +31,23 @@ type PtrClientObj[T any] interface {
...
-	version string
+	crd          *CRDInfo
+	majorVersion string
 }
 
 // APIImporter can translate itself into Kubernetes Objects.
 // Use to customize or accelerate translations ad-hoc
 type APIImporter[T any, P PtrClientObj[T]] interface {
-	FromAPI(t *Translator, target P) ([]client.Object, error)
+	FromAPI(t Translator, target P) ([]client.Object, error)
```
s-urbaniak: why not struct? I'd prevent unnecessary shallow copying

internal/v3/translate/translator.go
```diff
@@ -69,16 +63,18 @@ type APIExporter[T any] interface {
 //	  v20250312:
 //
 // In the above case crdVersion is "v1" and majorVersion is "v20250312".
-func NewTranslator(crd *apiextensionsv1.CustomResourceDefinition, crdVersion string, majorVersion string, deps DependencyRepo) *Translator {
-	return &Translator{
-		crd:          CRDInfo{definition: crd, version: crdVersion},
-		majorVersion: MajorVersion{version: majorVersion},
-		deps:         deps,
+// A translator is an immutable configuration object, it can be safely shared
+// across threads
```
s-urbaniak: this description reads like a description of the struct not the method, hence it should be on the struct. The godoc here should describe what New... does

internal/v3/translate/translator.go
```diff
@@ -69,16 +63,18 @@ type APIExporter[T any] interface {
...
+func NewTranslator(crd *apiextensionsv1.CustomResourceDefinition, crdVersion string, majorVersion string) Translator {
```
s-urbaniak:
```suggestion
func NewTranslator(crd *apiextensionsv1.CustomResourceDefinition, crdVersion string, majorVersion string) *Translator {
```
Let's not stress memory allocation

internal/v3/translate/translator.go
```diff
@@ -69,16 +63,18 @@ type APIExporter[T any] interface {
...
+	return Translator{
+		crd:          &CRDInfo{definition: crd, version: crdVersion},
+		majorVersion: majorVersion,
 	}
 }
 
-// FromAPI translaters a source API structure into a Kubernetes object
-func FromAPI[S any, T any, P PtrClientObj[T]](t *Translator, target P, source *S) ([]client.Object, error) {
+// FromAPI translaters a source API structure into a Kubernetes object.
+// Thread safe: translator is read-only
```
s-urbaniak: general convention from the past: describe the overall thread-safety constraint on the struct level and describe only exceptions on a specific method.

s-urbaniak (changes requested):
just a couple of nuances else LGTM

internal/v3/translate/translator.go
josvazg: Will move the constructor comment here.

internal/v3/translate/translator.go
josvazg: Translator is now 24 bytes, the size of a slice. The idea was to enforce immutability, as the data should be immutable anyways.

I can go back to a pointer. I am going to need to add a randomisation seed.

internal/v3/translate/translator.go
josvazg: moved to the struct

internal/v3/translate/translator.go
josvazg: done

s-urbaniak (approved):

PR #2764, opened 2025-10-07
CLOUDP-349340: transient build for generated manifests

s-urbaniak:
# Summary

This removes presumably unnecessary manifest artifacts that are supposed to be generated during releases only.

## Proof of Work

Passing e2e tests.

## Checklist

  * **q**: Clarifying question.


Makefile
```diff
@@ -426,6 +424,14 @@ x509-cert: ## Create X.509 cert at path tmp/x509/ (see docs/x509-user.md)
 
 clean: ## Clean built binaries
 	rm -rf bin/*
+	rm -rf config/manifests/bases/
```
josvazg, line 431: **nit or follow up**: make rule `clear-e2e-leftovers` can also be removed now

Makefile
```diff
@@ -319,11 +321,7 @@ validate-crds-chart: ## Validate the CRDs in the Helm chart
 
 .PHONY: bundle
 bundle: manifests  ## Generate bundle manifests and metadata, then validate generated files.
-	@echo "Building bundle $(VERSION)"
-	operator-sdk generate $(KUSTOMIZE) manifests -q --apis-dir=api
-	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
-	$(KUSTOMIZE) build --load-restrictor LoadRestrictionsNone config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
-	operator-sdk bundle validate ./bundle
+	INPUT_VERSION=$(VERSION) INPUT_ENV=prod INPUT_IMAGE_URL=$(IMG) sh ./.github/actions/gen-install-scripts/entrypoint.sh
```
roothorp, line 328: Do we always want `INPUT_ENV` to be prod? `gen-install-scripts` has a check if this is dev. If we always want this to be prod then why set it at all?

josvazg: The script requires that arg. I would leave it as is for now, otherwise you also need to fix that script, and this change is big enough already.

PR #2783, opened 2025-10-10
CLOUDP-350358: Improve and increase crd2go test coverage

helderjs:
# Summary

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/tools/crd2go/internal/checkerr/checkerr_test.go
+++ b/tools/crd2go/internal/checkerr/checkerr_test.go
@@ -1,3 +1,18 @@
+// Copyright 2025 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
 package checkerr
 
 import (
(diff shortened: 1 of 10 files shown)


tools/crd2go/internal/gotype/typedict.go
```diff
@@ -149,7 +149,7 @@ func (td TypeDict) RenameType(parentNames []string, gt *GoType) error {
 	}
 	goType.Name = typeName
 	if goType.Name == "string" {
-		log.Printf("here")
+		log.Printf("here") //TODO handle this case better
 	}
```
josvazg: This was a trace I missed to remove. We can remove it completely.
(This trace allowed me to use debugging on a particular type even when the code was passing here for many other non interesting cases)

tools/crd2go/pkg/crd2go/crd2go_test.go
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 //
 
-package crd2go_test
+package crd2go
```
josvazg, line 16: **q**: Why was this changed?

It is usually preferable to test packages from the outside, to focus on public interfaces. When testing from the inside it might be a code smell, as tests might be checking implementation details rather than public interfaces behaviour, or it might be due to some lack in the design that makes the code difficult to test or use. In this case tests were able to run from the outside without issues AFAIK.

helderjs: Sure, I believe this holds true if we want exclusively test the public interface of the package, and might be valid here once this is going to be a public (open source) package. However, the public interfaces implementations are usually a combination of several internal units, and testing these granular units directly, you can cover a vast number of edge cases with simple, focused tests.
Regarding `testing from the inside it might be a code smell, as tests might be checking implementation details`. The test only breaks if the fundamental contract of that small unit changes, which is precisely what a unit test should detect. The "brittle test" problem arises when people test how a function works, not what it accomplishes. This is poor test design, not an inherent flaw of white-box testing.
As last, I wanted all the tests to be under the same coverage.

josvazg: But in this case the tests worked fine without being in the same package and there is nothing added that required insider access.

Also it is perfectly fine in Go to have separate test files for testing from the outside and others for testing within. Here I see no need to change access.

PR #2804, opened 2025-10-15
Improve local CI checks

josvazg:
# Summary

Add addlicense and shellchecks to makefile and allow to run all quic k "CI checks" with the new `ci` rule

## Proof of Work

```shell
$ make -j3 ci
...
✅ CI PASSED all checks
make -j3 ci  110.63s user 25.97s system 265% cpu 51.432 total
```

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,9 @@ HELM_AKO_NAMESPACE = $(OPERATOR_NAMESPACE)
 
 GH_RUN_ID=$(shell gh run list -w Test -b main -e schedule -s success --json databaseId | jq '.[0] | .databaseId')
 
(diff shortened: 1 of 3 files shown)


josvazg:
> Does this mean that these first stages of CI would not run in parallel?

TIL that make -j allows for concurrent rule runs. As the `ci` depends on several rules, make will run them concurrently.
https://explainshell.com/explain?cmd=make+-j

Only issue is that concurrent runs have messy output, so when it fails it is usually clearer to remove the -j option.

josvazg:
BTW I think -k does not make sense in this case
https://explainshell.com/explain?cmd=make+-j+-k

PR #2805, opened 2025-10-15
Simplify CI workflows

josvazg:
# Summary

Collapse all GitHub CI quick tests into 2 calls:

1. `make unit-test`
2. `make -j3 all-lints` 

`unit-test` + `all-lints` = `ci`

Initially we wanted to run everything in a single `make -j3 ci` call, but it run slower than before (12 minutes instead of 9min). After the split we get 8 minutes instead of 9.

Also separate per tool CIs to trigger only when their code changes. Plus dummy changes to see them run here.

## Proof of Work

New CI should work as expected.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   unit-test:
-    name: AKO CI tests 
+    name: AKO unit-tests 
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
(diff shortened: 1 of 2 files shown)


roothorp (commented):
Love it! Great concept and anything that removes logic exclusive to GHA is a win in my book. I was unaware of the -j flag for make, thats very cool.

josvazg:
> Love it! Great concept and anything that removes logic exclusive to GHA is a win in my book. I was unaware of the -j flag for make, thats very cool.

I learned about -j for this. I had not used it before.

helderjs (approved):

PR #2810, opened 2025-10-20
Ensure e2e local script installs CRDs properly

josvazg:
# Summary

Files in `deploy/crds` are not final CRD YAML files that Kubernetes can consume, instead they are templates taht need to be expanded by **kustomize** before they become CRDs.

This is a left over fix from the recent cleanup of intermediate files in the repo. And also not seem before as the CI e2e test code path and the local one differ.

## Proof of Work

✅ `time make e2e label=flex`

## Checklist

### DIFF
--- a/scripts/e2e_local.sh
+++ b/scripts/e2e_local.sh
@@ -48,7 +48,7 @@ if [[ "${build}" == "true" ]]; then
     docker push "${bundle_image}"
 fi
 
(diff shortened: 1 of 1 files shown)


josvazg:
> Good short term fix, but we should get rid of the e2e local and standardise how we run e2e tests, regardless of where.

Agreed

PR #2818, opened 2025-10-20
CLOUDP-346332: Bump Go to 1.25 solving all related issues

josvazg:
# Summary

Go 1.25 required a couple of fixes:
- The **mockery** dependency had to be compiled against Go 1.25. This was not feasible as a Nix package.
- MacOS tests must use Apple SDK 15.2 library under Nix to avoid link [issue reported upstream](https://github.com/NixOS/nixpkgs/issues/450645). Fix is [taken from here](https://github.com/linode/cluster-api-provider-linode/pull/916/files)

CLOUDP-346332

## Proof of Work

CI passes, locally unit, e2e, e2e2 and int tests could be run as well on Mac.

Quick local ginkgo test was: `make [e2e2|e2e|test/int] label=blah` as this was a link issue, success was just being to see gingko fail.

## Checklist


go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/mongodb/mongodb-atlas-kubernetes/v2
 
-go 1.24.0
+go 1.25.0
 
-toolchain go1.24.7
+toolchain go1.25.1
```
roothorp, line 5: Can we bump this to 1.25.3? I think it's security & bug fixes, better safe than sorry?

josvazg: yep, let me try that on a new commit here

josvazg: I had issues with this, but then I also discovered another issue with the fix locally

it might be problematic, but I will retry again, once I am sure the base fix is good

josvazg: I am going to have to pass on this for now, you get this error often.

```shell
compile: version "go1.25.3" does not match go tool version "go1.25.1"
```

I would rather put this on a separate effort and pR to bump to a newer go 1.25 minor version.

PR #3296, opened 2026-03-24
CLOUDP-382045: Added IPAccessListEntry resource

igor-karpukhin:
# Summary

Added IPAccessListEntry autogenerated resource

## Proof of Work

e2e tests pass


internal/generated/controller/ipaccesslist/handler_v20250312.go
```diff
@@ -0,0 +1,464 @@
...
+
+// HandleImportRequested imports an existing Atlas IP access list entry.
+// The annotation mongodb.com/external-id must be set to "groupId:entryValue".
+func (h *Handlerv20250312) HandleImportRequested(ctx context.Context, ipaccesslist *akov2generated.IPAccessList) (ctrlstate.Result, error) {
+	externalID, ok := ipaccesslist.GetAnnotations()["mongodb.com/external-id"]
+	if !ok {
+		return result.Error(state.StateImportRequested, errors.New("missing annotation mongodb.com/external-id"))
+	}
+
+	groupID, entryValue, err := parseExternalID(externalID)
```
helderjs: groupID must come from dependency (via groupRef) or the groupID itself

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
```diff
@@ -0,0 +1,464 @@
...
+	if strings.Contains(ip, "/") {
+		return normalizeCIDR(ip)
+	}
+
+	return ip + "/32"
+}
+
+func normalizeCIDR(cidr string) string {
+	_, network, err := net.ParseCIDR(cidr)
+	if err != nil {
```
helderjs, line 559: Why do you return the cidr string if parse returns an error?

internal/generated/controller/ipaccesslist/handler_v20250312.go
```diff
@@ -0,0 +1,464 @@
...
+	parts := strings.SplitN(externalID, ":", 2)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", "", fmt.Errorf("invalid external-id %q: expected format groupId:entryValue", externalID)
+	}
+
+	return parts[0], parts[1], nil
+}
+
+// normalizeIP converts a plain IP address to a /32 CIDR
+func normalizeIP(ip string) string {
```
helderjs: Why the normalize "converts" the IP in a CIDR? Is that what the API really expect?

internal/generated/controller/ipaccesslist/handler_v20250312.go
```diff
@@ -0,0 +1,464 @@
...
+}
+
+// HandleInitial creates a new IP access list entry in Atlas.
+func (h *Handlerv20250312) HandleInitial(ctx context.Context, ipaccesslist *akov2generated.IPAccessList) (ctrlstate.Result, error) {
+	groupID, entryValue, err := h.resolveIdentity(ctx, ipaccesslist)
+	if err != nil {
+		return result.Error(state.StateInitial, fmt.Errorf("failed to resolve IPAccessList identity: %w", err))
+	}
+
+	entry := buildNetworkPermissionEntry(ipaccesslist)
```
helderjs: I think this function is doing validation that is responsibility of Atlas API as well as doing the work of crapi. Shoould you somply use crapi instead?

internal/generated/controller/ipaccesslist/handler_v20250312.go
```diff
@@ -0,0 +1,464 @@
...
+	if atlasStatus != "ACTIVE" {
+		return result.NextState(state.StateUpdating, fmt.Sprintf("IP access list entry is %s. Waiting for activation.", atlasStatus))
+	}
+
+	return result.NextState(state.StateUpdated, "IP access list entry is up to date.")
+}
+
+// HandleUpdated handles the updated steady state.
+func (h *Handlerv20250312) HandleUpdated(ctx context.Context, ipaccesslist *akov2generated.IPAccessList) (ctrlstate.Result, error) {
+	if expired := checkExpiry(ipaccesslist); expired {
```
helderjs: We should not check that, we should delegate to Atlas.

internal/generated/controller/ipaccesslist/handler_v20250312.go
```diff
@@ -0,0 +1,464 @@
...
+	if atlasStatus != "ACTIVE" {
+		return result.NextState(state.StateCreating, fmt.Sprintf("IP access list entry is %s. Waiting for activation.", atlasStatus))
+	}
+
+	return result.NextState(state.StateCreated, "IP access list entry is active.")
+}
+
+// HandleCreated handles the created steady state.
+func (h *Handlerv20250312) HandleCreated(ctx context.Context, ipaccesslist *akov2generated.IPAccessList) (ctrlstate.Result, error) {
+	if expired := checkExpiry(ipaccesslist); expired {
```
helderjs: We should not check that, we should delegate to Atlas.

internal/generated/controller/ipaccesslist/handler_v20250312.go
```diff
@@ -0,0 +1,464 @@
...
+		return result.Error(state.StateImportRequested, err)
+	}
+
+	return result.NextState(state.StateImported, "IP access list entry imported.")
+}
+
+// HandleImported handles the imported state.
+func (h *Handlerv20250312) HandleImported(ctx context.Context, ipaccesslist *akov2generated.IPAccessList) (ctrlstate.Result, error) {
+	if expired := checkExpiry(ipaccesslist); expired {
+		return result.NextState(state.StateDeletionRequested, "IP access list entry has expired.")
```
helderjs: Why do you delete the user if it expired? I think you are taking a decision on behave of the user and deviating from what's defined in the spec

helderjs (changes requested):
Few commments. I'd also add tests covering IP and expiration.

internal/generated/controller/ipaccesslist/handler_v20250312.go
igor-karpukhin: This is for import requested case, not for regular handling. Same pattern as we have in DatabaseUser and Group

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
igor-karpukhin: This is the best-effort normalization. A small check on the AKO side, if the CIDR is correct. If not, Atlas would return an error anyway

internal/generated/controller/ipaccesslist/handler_v20250312.go
igor-karpukhin: The new controller for autogenerated IPAccessList partially replicates how the controller for legacy ipAccessList field worked for AtlasProject resource. The translation need to explicitly convert IP address into CIDR notation. E.g.: `192.168.0.1` has to be stored and sent like `192.168.0.1/32`. I can replace it with h.Translator.ToAPI, but I'll have to patch it anyway.

internal/generated/controller/ipaccesslist/handler_v20250312.go
igor-karpukhin: We have to check this unfortunately, as list entries do not have IDs. If the IP/CIDR is expired by atlas, the AKO will try to create it, but atlas will fail in this case. This is why the legacy controller marks the entry in the status as expired and didn't reconcile it anymore after that

internal/generated/controller/ipaccesslist/handler_v20250312.go
igor-karpukhin: See the comment above

internal/generated/controller/ipaccesslist/handler_v20250312.go
igor-karpukhin: This is how legacy controller worked. The IP address from Atlas is always returned with the mask

internal/generated/controller/ipaccesslist/handler_v20250312.go
igor-karpukhin: I can revert this one to replicated the legacy controller behaviour, that would just mark it as `Expired`, let's discuss this one

internal/generated/controller/ipaccesslist/handler_v20250312.go
igor-karpukhin: UPD: tested it with the newest version of the SDK, and the IP is returned as is. Removed this method.

internal/generated/controller/ipaccesslist/handler_v20250312.go
igor-karpukhin: Fixed it to store the "expired" in the status

go.mod
```diff
@@ -101,6 +101,7 @@ require (
 	github.com/x448/float16 v0.8.4 // indirect
 	github.com/yudai/golcs v0.0.0-20170316035057-ecda9a501e82 // indirect
 	github.com/yudai/pp v2.0.1+incompatible // indirect
+	go.mongodb.org/atlas-sdk/v20250312016 v20250312016.0.0 // indirect
```
helderjs, line 104: **q**: Why is `v20250312016` still here?

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
```diff
@@ -0,0 +1,564 @@
...
+
+// handleSteadyState checks whether the spec changed and recreates the entry if needed.
+func (h *Handlerv20250312) handleSteadyState(ctx context.Context, currentState state.ResourceState, ipaccesslistentry *akov2generated.IPAccessListEntry) (ctrlstate.Result, error) {
+	deps, err := h.getDependencies(ctx, ipaccesslistentry)
+	if err != nil {
+		return result.Error(currentState, fmt.Errorf("failed to resolve IPAccessListEntry dependencies: %w", err))
+	}
+
+	// First, check if the entry has expired
+	if hasDeleteAfterDate(ipaccesslistentry) {
```
helderjs, line 272: What is the point here? The resource will eventually reconcile according to the sync period.
If a user is created to with delete adter 6 months, and the resource reconcile once per day, the queue will have 182 request on it.

helderjs (changes requested):
Again, I think we should not do the expiration calculation. If a user expires (an Atlas deletes it) reconciliation will fail and we simply reflect in a clear way in the status.

crd2go/openapi2crd.yaml
```diff
@@ -224,3 +223,34 @@ spec:
...
                 - $.terminationProtectionEnabled
+                
+    - gvk:
+        version: v1
+        kind: IPAccessListEntry
+        group: atlas.generated.mongodb.com
+      categories:
+        - atlas
+      shortNames:
+        - aial
```
josvazg, line 234: **q**: why not this abbreviation instead?

```suggestion
        - ipal
```

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
```diff
@@ -0,0 +1,564 @@
...
+}
+
+// HandleInitial creates a new IP access list entry in Atlas.
+func (h *Handlerv20250312) HandleInitial(ctx context.Context, ipaccesslistentry *akov2generated.IPAccessListEntry) (ctrlstate.Result, error) {
+	groupID, entryValue, err := h.resolveIdentity(ctx, ipaccesslistentry)
+	if err != nil {
+		return result.Error(state.StateInitial, fmt.Errorf("failed to resolve IPAccessListEntry identity: %w", err))
+	}
+
+	entry := buildNetworkPermissionEntry(ipaccesslistentry)
```
josvazg, line 72: **q**: I see a lot of custom logic, including conversions. Why was crapi not used here?

internal/generated/indexers/ipaccesslistentrybygroup.go
```diff
@@ -0,0 +1,91 @@
...
+	client "sigs.k8s.io/controller-runtime/pkg/client"
+	handler "sigs.k8s.io/controller-runtime/pkg/handler"
+	log "sigs.k8s.io/controller-runtime/pkg/log"
+	reconcile "sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	v1 "github.com/mongodb/mongodb-atlas-kubernetes/v2/generated/v1"
+)
+
+// nolint:dupl
+const IPAccessListEntryByGroupIndex = "ipaccesslistentry.groupRef"
```
josvazg, line 32: is this one custom? not autogenerated?

josvazg (commented):
My feedback so far:
- Low unit test coverage, is that coming?
- I see no reason why we could not use crapi's **ToAPI** and **FromAPI**, our canonical way to translate things back and forth between the API structs and kube.

igor-karpukhin:
> Again, I think we should not do the expiration calculation. If a user expires (an Atlas deletes it) reconciliation will fail and we simply reflect in a clear way in the status.

The expiration calculation is "best-effort". Without this, if the entry is expired, users won't notice until the next reconciliation, which can be in 3h. The status is set based on the response from Atlas and the current state of the CR. Regarding the 182 requests, they won't be there, as work queue deduplicates requests based on the key, so they won't be stacked.

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
igor-karpukhin: Because IP access list is stored as CIDR with the mask. I explained that to @helderjs already.

crd2go/openapi2crd.yaml
igor-karpukhin: I took the first character from **A**tlas**I**p**A**ccess**L**ist. But we can rename it, if we all agree in the short name

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
igor-karpukhin: See the comment below

go.mod
igor-karpukhin: A left over after the rebase. Fixed.

crd2go/openapi2crd.yaml
igor-karpukhin: What about `ipale`? As it is the **I**P**A**ccess**L**ist**E**ntry

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
josvazg: If this type is special and cannot be covered by **crapi**, I am ok to merge for now, but I will also create an issue against **crapi** to support this use case eventually. We do not want to go back to snowflakes.

Can you elaborate exactly what broke **crapi**? is it because it is a list (array) instead of a struct?
how does the "list stored as CIDR" breaks it?
cc @s-urbaniak this might be a corner case we missed

internal/generated/indexers/ipaccesslistentrybygroup.go
igor-karpukhin: Why? This one is generated

helderjs (approved):

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
igor-karpukhin: I noticed two things:
* `translator.ToAPI` uses marshalling with `map[string]any`, so it copies string as is. While `net.ParseCIDR` correctly parses the CIDR string. E.g. `192.168.1.5/24` -> `192.168.1.0/24`. Without this, spec and atlas spec will always be different
* `deleteAfterDate` is `*string` in the CRD spec but `*time.Time` in the API payload. That's why I parse timestamp manually

One more thing, not so critical, but `groupID` would be in the payload, but I think Atlas drops anything that doesn't match the schema

crd2go/openapi2crd.yaml
josvazg: I still like ipal better, but I am not going to block the PR on that

internal/generated/indexers/ipaccesslistentrybygroup.go
josvazg: it lacks the auto gen comment, that is why I asked

internal/generated/controller/ipaccesslistentry/handler_v20250312.go
josvazg: Thanks for the breakdown!

 For `deleteAfterDate`: yes, this is a crapi gap worth tracking — crapi needs field-level type transformers to handle `*string` ↔ `*time.Time` without manual parsing in handlers. I will open an issue against crapi for that.

For CIDR normalization: this doesn't belong in crapi — it's domain logic, not a translation concern. We can keep calling normalizeCIDR even if we move to crapi later.

I'm OK merging while we track the crapi issue on the side.

josvazg (approved):
LGTM, tracking crapi gap separately here CLOUDP-398228

PR #3311, opened 2026-04-09
Try new go fix command

roothorp:
# Summary

Go 1.26 reworked the `go fix` command;
> Go fix uses a suite of algorithms to identify opportunities to improve your code, often by taking advantage of more modern features of the language and library.

This should mean just running `go fix ./...` should modernise and/or fix things around our codebase, without changing functionality. The vast majority of this seems to be using `new()` instead of `pointer.MakePtr()`.

## Proof of Work

CI should pass.

## Checklist

  * **q**: Clarifying question.


api/condition.go
```diff
@@ -139,7 +139,7 @@ type Condition struct {
 	// Last time the condition transitioned from one status to another.
 	// Represented in ISO 8601 format.
 	// +optional
-	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+	LastTransitionTime metav1.Time `json:"lastTransitionTime"`
```
josvazg: **q**: @roothorp was this change on purpose? 

This is part of a public API we should not change unless we have a good reason

josvazg (commented):
LGTM except for a question, but would hold until that is cleared

api/condition.go
roothorp: I didn't manually do this, I assume `fix` did it because it is redundant (perhaps because there is no "empty" value for this field?) I can manually re-add the field if you think it would be better

josvazg: I think is safer to manually undo this one, at least for this automated change. This changes the wire (JSON) format is some corner cases and this change is not supposed to have side effects like that.

roothorp: I am fairly sure that this has no effect, because you cannot omit the `Time` field (since it is not a pointer). If we manually change this back, we are just going to have this again every time we run `fix` in future.

josvazg: yes, it cannot be nil, but not sure how the zero value of the time will be encoded

roothorp: reverted this

josvazg (approved):

PR #3316, opened 2026-04-14
openapi2crd: make OpenAPI spec path configurable in config

s-urbaniak:
# Summary

Makes the OpenAPI spec file path configurable in the `openapi2crd` config instead of hardcoding `../openapi/atlas-api-transformed.yaml`.

Key changes:
- Add `path` field to the OpenAPI definition in config YAML (`openapi2crd.yaml`, `openapi2crd.experimental.yaml`)
- Introduce `PackageLoader` interface with `LoadFromPackage`/`LoadFlattenedFromPackage` that accept a relative path parameter
- Refactor `Atlas` loader to use the configurable relative path when resolving Go module directories
- Remove `normalizeScalarStyles` and YAML quoting helpers (no longer needed after flattening improvements)
- Add `inline_enum_oneof` test case for flattening and fix enum merging in `flattenInlineCompositions`

## Proof of Work

- `make unit-test` passes
- Updated tests in `atlas_test.go` and `generator_test.go` cover the new path parameter

## Checklist

### DIFF
--- a/tools/openapi2crd/pkg/config/openapi.go
+++ b/tools/openapi2crd/pkg/config/openapi.go
@@ -27,19 +27,13 @@ import (
 	"github.com/spf13/afero"
 	"golang.org/x/sync/singleflight"
 
+	"github.com/mongodb/mongodb-atlas-kubernetes/tools/openapi2crd/pkg/apis/config/v1alpha1"
 	"github.com/mongodb/mongodb-atlas-kubernetes/tools/openapi2crd/pkg/flatten"
 )
 
+// Loader loads an OpenAPI spec based on the provided definition.
 type Loader interface {
-	Load(ctx context.Context, path string) (*openapi3.T, error)
-	LoadFlattened(ctx context.Context, path string) (*openapi3.T, error)
-}
-
-// PackageLoader resolves a Go package to a directory and loads an OpenAPI spec
-// at a path relative to that directory.
-type PackageLoader interface {
-	LoadFromPackage(ctx context.Context, pkg, relPath string) (*openapi3.T, error)
-	LoadFlattenedFromPackage(ctx context.Context, pkg, relPath string) (*openapi3.T, error)
+	Load(ctx context.Context, def v1alpha1.OpenAPIDefinition) (*openapi3.T, error)
 }
 
 type KinOpenAPI struct {
@@ -56,7 +50,7 @@ func NewKinOpenAPI(fs afero.Fs) *KinOpenAPI {
 	}
 }
 
-func (a *KinOpenAPI) Load(_ context.Context, path string) (*openapi3.T, error) {
+func (a *KinOpenAPI) load(_ context.Context, path string) (*openapi3.T, error) {
 	// Fast path: return the cached spec without entering singleflight.
 	// The mutex-guarded cache avoids the overhead of singleflight.Do on
 	// every call after the spec has already been parsed.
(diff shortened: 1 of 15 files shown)


tools/openapi2crd/pkg/flatten/testdata/additional_properties/golden.yaml
```diff
@@ -2,11 +2,11 @@ paths:
   /config:
     get:
       responses:
-        "200":
+        '200':
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Config"
+                $ref: '#/components/schemas/Config'
```
josvazg: **q**: seems unrelated?

tools/openapi2crd/pkg/flatten/structural.go
```diff
@@ -74,6 +74,7 @@ func flattenInlineCompositions(root *yaml.Node, key string) {
 			}
 			children := resolveSequenceRefs(seq, root, key)
 			if len(children) > 0 && allHaveEnum(children) {
+				mergeEnums(m, children, key)
```
josvazg: **q**: how is this related to the new path setting?

tools/openapi2crd/pkg/flatten/testdata/additional_properties/golden.yaml
s-urbaniak: pretty much, it's cosmetic, the AI made this more idiomatic to YAML

josvazg (commented):
I see some changes in golden files and test data that do not seem related to the main purpose of the PR, are those intended? How are they related to the new path setting?

tools/openapi2crd/pkg/flatten/structural.go
s-urbaniak: let me check, i think this sneaked in via rebasing, might be at least a separate commit.

s-urbaniak: yeah, i wanted this to be a separate commit, let me actually remove it from this PR

s-urbaniak: i removed it @josvazg

tools/openapi2crd/pkg/flatten/testdata/additional_properties/golden.yaml
s-urbaniak: removed this one as well @josvazg

tools/openapi2crd/pkg/config/openapi.go
```diff
@@ -35,6 +35,13 @@ type Loader interface {
 	LoadFlattened(ctx context.Context, path string) (*openapi3.T, error)
 }
 
+// PackageLoader resolves a Go package to a directory and loads an OpenAPI spec
+// at a path relative to that directory.
+type PackageLoader interface {
```
helderjs: **s**: I think this new interface introduces a distinction between implementations that share the same goal: loading an OpenAPI spec from a source. While I see the valid intention of simplifying routing, I believe it would be better to keep both implementations under the same interface and introduce a separate concrete component responsible for routing based on the config. This would remove that responsibility from generator.go, which shouldn't need to know how specs are resolved.

s-urbaniak: sounds good, what would be the proposal, something like this?
```
LoadFromPackage(ctx context.Context, pkg, relPath string, flatten bool) (*openapi3.T, error)
```

helderjs: Something like this for a unified interface?
```
type Loader interface {
    Load(ctx context.Context, def v1alpha1.OpenAPIDefinition) (*openapi3.T, error)
}
```

s-urbaniak: i like this, let me feed this into the agent, i think we can come up with a nice refactoring 👍

s-urbaniak: done, PTAL

helderjs (approved):

josvazg (approved):

PR #3323, opened 2026-04-20
CLOUDP-373260 Atlas Service Accounts support - phase 1

MaciejKaras:
# Summary

Adds Service Account (OAuth 2.0) authentication support to the Atlas Kubernetes Operator. Users can supply `clientId` / `clientSecret` in a Connection Secret; a new `ServiceAccountToken` controller obtains short-lived bearer tokens and caches them for all reconcilers to consume. Existing API-key authentication is unchanged.

See the [Technical Design](https://docs.google.com/document/d/1yf5GlGkhLHt_-o5c9sX-clKGNwbaxPtbojRjAD1WZBk/edit?tab=t.0#heading=h.ifgqtcm3jkzh).

Related: #3329 — the Helm chart changes that expose these credentials via `helm install` / `--set`.

## Proof of Work

Passing all new unit and e2e tests.

## Checklist

  * **q**: Clarifying question.


internal/controller/reconciler/credentials.go
```diff
@@ -78,14 +85,30 @@ func GetConnectionConfig(ctx context.Context, k8sClient client.Client, secretRef
 		return nil, fmt.Errorf("failed to read Atlas API credentials from the secret %s: %w", secretRef.String(), err)
 	}
 
-	cfg := &atlas.ConnectionConfig{
-		OrgID: string(secret.Data[orgIDKey]),
-		Credentials: &atlas.Credentials{
-			APIKeys: &atlas.APIKeys{
-				PublicKey:  string(secret.Data[publicAPIKey]),
-				PrivateKey: string(secret.Data[privateAPIKey]),
+	hasAPIKeys := len(secret.Data[publicAPIKey]) > 0 || len(secret.Data[privateAPIKey]) > 0
```
MaciejKaras: move to separate methods

MaciejKaras: maybe add validation for the Connection secret?

internal/controller/reconciler/credentials.go
```diff
@@ -106,8 +158,22 @@ func validate(cfg *atlas.ConnectionConfig) ([]string, bool) {
 		missingFields = append(missingFields, orgIDKey)
 	}
 
-	if cfg.Credentials == nil || cfg.Credentials.APIKeys == nil {
-		return append(missingFields, []string{publicAPIKey, privateAPIKey}...), false
+	if cfg.Credentials == nil {
+		return append(missingFields, publicAPIKey, privateAPIKey), false
+	}
+
+	if cfg.Credentials.ServiceAccount != nil {
```
MaciejKaras: that is not true. Credentials secret do not contain accessToken, but rather clientIDKey and clientSecretKey

internal/controller/accesstoken/accesstoken_test.go
```diff
@@ -0,0 +1,111 @@
...
+	// Pin the exact output for a known input pair. Any change to the derivation
+	// scheme (algorithm, separator, prefix, encoding) flips this value and
+	// every existing Access Token Secret on a live cluster would be orphaned
+	// behind a new name. Treat this assertion as a compatibility contract.
+	got, err := accesstoken.DeriveSecretName("atlas-operator", "my-sa-creds")
+	require.NoError(t, err)
+	assert.Equal(t, "atlas-access-token-my-sa-creds-587997bcdf678bb69ff7", got)
+}
+
+func TestDeriveSecretName_NamespaceSensitive(t *testing.T) {
```
josvazg: nit: I suspect many of these tests could be a single table test with different test case inputs.

The project tends to favour table tests when possible/appropriate as it is usually easier to add cases later on and the code is more compact and regular.

internal/controller/atlas/provider_test.go
```diff
@@ -143,6 +148,51 @@ func TestProvider_IsResourceSupported(t *testing.T) {
 	}
 }
 
+func TestProvider_SdkClientSet_NilCredentials(t *testing.T) {
+	p := NewProductionProvider("https://cloud.mongodb.com", false, false)
```
josvazg: **nit**: I see tabulation also possible here, and preferable

josvazg:
Added label `test/e2e2/*` to ensure the included e2e2 test would run.

internal/controller/serviceaccounttoken/serviceaccounttoken_controller.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func (r *ServiceAccountTokenReconciler) SetupWithManager(mgr ctrl.Manager, skipNameValidation bool) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(r.For()).
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(MapAccessTokenSecretToOwner),
+			builder.WithPredicates(credentialsLabelPredicate()),
+		).
```
josvazg, line 275: **q**: don't we have for and watches firing both on the same secrets twice?

Not a big deal, but it would be more explicit to fire FOR only on the client provided secrets and WATCH only the token produced ones, which I think was the initial intention.

internal/controller/serviceaccounttoken/serviceaccounttoken_controller.go
```diff
@@ -0,0 +1,272 @@
...
+	log *zap.SugaredLogger,
+	tokenSecret *corev1.Secret,
+	clientID, clientSecretValue, credsHash string,
+) (ctrl.Result, error) {
+	token, expiry, err := r.TokenProvider.FetchToken(ctx, clientID, clientSecretValue)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to refresh access token: %w", err)
+	}
+
+	tokenSecret.Data[accesstoken.AccessTokenKey] = []byte(token)
```
josvazg, line 146: `corev1.Secret.Data` has `json:"data,omitempty"`, so a Secret with no data fields deserializes with Data == nil.

Then
  tokenSecret.Data[accesstoken.AccessTokenKey] = []byte(token)
a write to a nil map, will panic in Go.

Maybe we should  guard all this from nils prepending:
```go
  if tokenSecret.Data == nil {
      tokenSecret.Data = map[string][]byte{}
  }
```

josvazg (commented):
looking good, just a few minor issues and questions and waiting on CI to confirm approvals

josvazg:
Cloud QA is having some issues cleaning up flexclusters, which explains the CI issue. Basically tests are looking good

Still I would want to see one full run of all tests to merge this one, as it is a cross controller feature.

helderjs (approved):

josvazg (approved):

josvazg:
I kicked off all other tests, we are not in a hurry as Cloud QA is broken for us right now and releasing this would be blocked by that anyways

josvazg:
BTW great job in this first contribution to AKO @MaciejKaras !

internal/controller/serviceaccounttoken/serviceaccounttoken_controller.go
MaciejKaras: Good point! AccessToken secret data shouldn't be empty, because the Operator is managing it, but I can imagine user mistakenly deleting the data.

internal/controller/accesstoken/accesstoken_test.go
MaciejKaras: Done!

internal/controller/atlas/provider_test.go
MaciejKaras: Done!

internal/controller/serviceaccounttoken/serviceaccounttoken_controller.go
MaciejKaras: That's a valid comment, thanks.

What happens today in each scenario:

 | Event                                     | `For` enqueues                                                                | `Watches` enqueues                                             |
  | ----------------------------------------- | ----------------------------------------------------------------------------- | -------------------------------------------------------------- |
  | SA Connection Secret changes              | self → real work                                                              | nil (no Secret owner)                                          |
  | API-key-only Connection Secret changes    | self → **wasted** reconcile (early-returns at lines 86–88)                    | nil                                                            |
  | Access Token Secret changes               | self → **wasted** reconcile (early-returns at lines 86–88)                    | owner Connection Secret → expected hash-match reconcile        |

The `Watches` side is already correctly narrowed: `MapAccessTokenSecretToOwner` returns `nil` for Secrets without a Secret owner, so it effectively only fires for Access Token Secrets. The asymmetry is on `For`.

I was thinking about adding a predicate on `For` that matches only Secrets with non-empty `clientId` and `clientSecret`. That naturally excludes both Access Token Secrets (they carry `accessToken` / `expiry` / `credentialsHash`) and API-key-only Connection Secrets, eliminating the wasted reconciles. wdyt?

```go
 func serviceAccountCredentialsPredicate() predicate.Predicate {
     return predicate.NewPredicateFuncs(func(obj client.Object) bool {
         secret, ok := obj.(*corev1.Secret)
         if !ok {
             return false
         }
         return len(secret.Data[reconciler.ClientIDKey]) > 0 &&
             len(secret.Data[reconciler.ClientSecretKey]) > 0
     })
 }

 func (r *ServiceAccountTokenReconciler) For() (client.Object, builder.Predicates) {
     return &corev1.Secret{}, builder.WithPredicates(
         credentialsLabelPredicate(),
         serviceAccountCredentialsPredicate(), // new
         predicate.ResourceVersionChangedPredicate{},
     )
 }
```

josvazg: seems like the right move!

MaciejKaras: Done.

internal/controller/reconciler/credentials.go
```diff
@@ -78,49 +84,115 @@ func GetConnectionConfig(ctx context.Context, k8sClient client.Client, secretRef
...
+	// may not have caught up yet. Returning an error prompts the downstream
+	// reconciler to retry rather than hitting Atlas with revoked credentials.
+	currentHash, err := accesstoken.CredentialsHash(string(secret.Data[ClientIDKey]), string(secret.Data[ClientSecretKey]))
+	if err != nil {
+		return "", err
+	}
+	if string(tokenSecret.Data[accesstoken.CredentialsHashKey]) != currentHash {
+		return "", fmt.Errorf("access token secret %s is stale (credentials rotated); waiting for the service-account-token controller to refresh", tokenRef.String())
+	}
+
```
igor-karpukhin, line 139: **q**: shouldn't we also verify if token is expired?

internal/controller/serviceaccounttoken/serviceaccounttoken_controller_test.go
```diff
@@ -0,0 +1,452 @@
...
+	assert.Equal(t, types.UID("test-uid"), tokenSecret.OwnerReferences[0].UID)
+
+	updatedSecret := &corev1.Secret{}
+	require.NoError(t, k8sClient.Get(context.Background(),
+		types.NamespacedName{Name: "sa-creds", Namespace: "ns"}, updatedSecret))
+	assert.Empty(t, updatedSecret.Annotations,
+		"credential secret must not be mutated by the controller")
+}
+
+func TestReconcile_RefreshesExpiredToken(t *testing.T) {
```
igor-karpukhin, line 140: **q**: does this only verifies a path with incorrect hash? Do we cover the case when the token is expired (old/invalid `expiry`)?

internal/controller/reconciler/credentials.go
MaciejKaras: The `CredentialsHashKey` check purpose is to prevent using stale access token that should be replaced by the serviceaccounttoken_controller when credentials were updated by the user. Renewing access token happens right after the credentials have changed, but it can take couple seconds until the access token Secret is replaced. We don't want to use previous credentials anymore to not raise any errors that are not relevant to new credentials - to not mislead the user.

The expiration check on the other hand should not provide additional value. If the access token has expired, user will still see the 401 error with proper information. I'll make sure that the error message contains information expired token by testing the API. If not, I'll add the check you've proposed. Thanks!

MaciejKaras:
```
HTTP/2 401
  www-authenticate: Bearer error="invalid_token"
  date: Wed, 29 Apr 2026 09:15:08 GMT
  server: mdbws

  (empty body)
```
  That's the entire response — no JSON body, just the bare WWW-Authenticate header. Status 401, error code `invalid_token`. I'll add the expiry assertion in the code, good catch!

internal/controller/serviceaccounttoken/serviceaccounttoken_controller_test.go
MaciejKaras: You're right. In this test I forgot to add `credentialsHash` in the token and it gets refreshed because of missing field, not because the token was expired.

MaciejKaras: This is now fixed! Thanks!

internal/controller/reconciler/credentials.go
MaciejKaras: This is now fixed!

internal/controller/accesstoken/accesstoken.go
```diff
@@ -0,0 +1,80 @@
...
+	secretNamePrefix = "atlas-access-token-"
+)
+
+// DeriveSecretName returns the deterministic Access Token Secret name for a
+// given Connection Secret. The connection secret name is included literally
+// in the result for operator debuggability; it is truncated when the total
+// would exceed the Kubernetes 253-character DNS-subdomain limit. Uniqueness
+// is guaranteed by the hash suffix, which always uses the full (untruncated)
+// name as input.
+func DeriveSecretName(namespace, connectionSecretName string) (string, error) {
```
igor-karpukhin: **nit**: the `Write` method on `hasher.Write` never returns an error, so you can simplify the `DeriveSecretName` so it just returns a string. https://pkg.go.dev/hash#Hash.

MaciejKaras: Good catch, didn't notice this earlier.

MaciejKaras: The same for CredentialsHash can be applied.

MaciejKaras: Done!

igor-karpukhin (approved):
LGTM. Great work @MaciejKaras! I tested it in my local cluster as well, it works 🎉

josvazg:
to be merged soon

PR #3334, opened 2026-04-28
CLOUDP-400938: fix(atlasdeployment): stop perpetual reconcile loop on terminationProtection and processArgs

ashortt-mdb:
## Summary

Fixes #3142 — two distinct bugs in the AtlasDeployment translation/reconcile path that caused the operator to loop endlessly on `UpdateCluster` / `UpdateProcessArgs` without ever converging.

- **`terminationProtectionEnabled: false` was dropped from PATCH.** `clusterCreateToAtlas` / `clusterUpdateToAtlas` wrapped the `bool` field with `pointer.MakePtrOrNil`, which returns `nil` for the zero value. Combined with the SDK's `omitempty`, `false` was stripped from the body, so when Atlas had `true` (e.g. set via UI) the operator could never push `false` and reconcile looped. Switched to `pointer.MakePtr` so the value is always sent.
- **processArgs `reflect.DeepEqual` triggered perpetual `UpdateProcessArgs` calls.** Atlas returns server-populated defaults (e.g. `defaultWriteConcern: "majority"`, `oplogSizeMB: 990`) for fields the user left unset in the CR. `reflect.DeepEqual` saw those as a diff, but the PATCH body built by `processArgsToAtlas` omits the same zero values via `MakePtrOrNil` — so Atlas received an effectively empty PATCH, kept its defaults, and the next reconcile saw the same diff. Replaced the call site with `deployment.ProcessArgsEqual`, which mirrors the PATCH-body omitempty semantics — fields the user has not set are treated as "no opinion" and not diffed.

## Test Plan

- [x] `make fmt && make lint && make unit-test` (clean)
- [x] New regression tests in `conversion_test.go`:
  - `TestClusterUpdateToAtlas_TerminationProtectionFalseIsSent`
  - `TestClusterCreateToAtlas_TerminationProtectionFalseIsSent`
  - `TestProcessArgs_UnsetFieldShouldNotDivergeFromAtlasDefaults`
- [x] New table-driven `TestProcessArgsEqual` in `compare_test.go` covering both-nil, AKO-unset vs Atlas-default, and explicit-divergence cases for every field.

## Notes

- Other bool fields in the same builders (`BackupEnabled`, `Paused`, `PitEnabled`) are already `*bool` and pass through directly — no change needed.
- `processArgsToAtlas` is intentionally left alone: omitting unset fields from the wire is correct; the bug was only in the equality check.
- Other reconcilers using `reflect.DeepEqual(akoState, atlasState)` (search_nodes.go, backup.go, flex_deployment.go) may have the same anti-pattern but are out of scope for this issue.

Refs #3142

### DIFF
--- a/internal/translation/deployment/compare.go
+++ b/internal/translation/deployment/compare.go
@@ -404,6 +404,22 @@ func ProcessArgsEqual(ako, atlas *akov2.ProcessArgs) bool {
 	if atlas == nil {
 		atlas = &akov2.ProcessArgs{}
 	}
+
+	// Each check below mirrors how processArgsToAtlas serializes the field:
+	//
+	//   *bool / *int64 fields  → check `!= nil`  (a non-nil pointer is sent,
+	//                                             even if it points to the
+	//                                             zero value)
+	//   string fields          → check `!= ""`   (empty string is dropped via
+	//                                             MakePtrOrNil and not sent)
+	//
+	// Keep the per-field style aligned with the converter — collapsing the two
+	// styles will reintroduce issue #3142.
+	//
+	// Fields intentionally absent below: they are not part of the SDK PATCH
+	// body and therefore cannot drive an update:
+	//   - DefaultReadConcern
+	//   - FailIndexKeyTooLong (deprecated by Atlas; warned on at the call site)
 	if ako.DefaultWriteConcern != "" && ako.DefaultWriteConcern != atlas.DefaultWriteConcern {
 		return false
 	}
--- a/internal/translation/deployment/compare.go
+++ b/internal/translation/deployment/compare.go
@@ -31,6 +31,7 @@ func ComputeChanges(desired, current *Cluster) (*Cluster, bool) {
 				Name:   desired.Name,
 				Paused: new(pointer.GetOrDefault(desired.Paused, false)),
 			},
+			pauseOnly: true,
 		}, true
 	}
 
--- a/internal/translation/deployment/compare.go
+++ b/internal/translation/deployment/compare.go
@@ -388,3 +388,45 @@ func areEqual[T comparable](desired, current *T) bool {
 
 	return val1 == val2
 }
+
+// ProcessArgsEqual reports whether the AKO-side processArgs and the
+// Atlas-side processArgs are semantically equivalent for reconcile purposes.
+//
+// Fields that are the zero value on the AKO side are treated as "no opinion"
+// and not compared. This mirrors processArgsToAtlas, which omits zero values
+// from the PATCH body via MakePtrOrNil — if a field would not be sent over
+// the wire, it must not drive an update either, or the reconciler loops on
+// Atlas-populated server defaults (issue #3142).
+func ProcessArgsEqual(ako, atlas *akov2.ProcessArgs) bool {
+	if ako == nil {
+		return true
+	}
+	if atlas == nil {
+		atlas = &akov2.ProcessArgs{}
+	}
+	if ako.DefaultWriteConcern != "" && ako.DefaultWriteConcern != atlas.DefaultWriteConcern {
+		return false
+	}
+	if ako.MinimumEnabledTLSProtocol != "" && ako.MinimumEnabledTLSProtocol != atlas.MinimumEnabledTLSProtocol {
+		return false
+	}
+	if ako.OplogMinRetentionHours != "" && ako.OplogMinRetentionHours != atlas.OplogMinRetentionHours {
+		return false
+	}
+	if ako.JavascriptEnabled != nil && !areEqual(ako.JavascriptEnabled, atlas.JavascriptEnabled) {
+		return false
+	}
+	if ako.NoTableScan != nil && !areEqual(ako.NoTableScan, atlas.NoTableScan) {
+		return false
+	}
+	if ako.OplogSizeMB != nil && !areEqual(ako.OplogSizeMB, atlas.OplogSizeMB) {
+		return false
+	}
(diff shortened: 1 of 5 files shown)


internal/translation/deployment/conversion.go
josvazg: flagged by claude

> *int64 fields have the same MakePtrOrNil shape as terminationProtectionEnabled had. In conversion.go:969-971, OplogSizeMB/SampleSizeBIConnector/SampleRefreshIntervalBIConnector go through MakePtrOrNil(int(GetOrDefault(*int64, 0))). If a user explicitly sets OplogSizeMB: 0 on the CR, it gets stripped from the PATCH.

> ProcessArgsEqual would then see *int64(0) on AKO vs the Atlas server default and report a diff → same loop class. Probably an invalid value anyway (Atlas likely rejects oplog 0), so impact is small, but the bug pattern is unfixed for these. Worth a follow-up issue or a MakePtr swap here too while you're in the area.

suggested fix 

Today:
```go
  OplogSizeMB: pointer.MakePtrOrNil(int(pointer.GetOrDefault(config.OplogSizeMB, 0))),
```
  The double-funnel (*int64 → int via GetOrDefault(0) → *int via MakePtrOrNil) collapses both "unset" and "explicitly zero" into nil.
  Replace it with a nil-preserving conversion:

```go
  OplogSizeMB: int64PtrToIntPtr(config.OplogSizeMB),
```
  with a tiny helper next to the other pointer helpers:

```go
  // int64PtrToIntPtr converts *int64 to *int while preserving nil-ness.
  // Use this instead of MakePtrOrNil(int(...)) when 0 is a valid user value
  // that must be sent over the wire (issue #3142).
  func int64PtrToIntPtr(p *int64) *int {
      if p == nil {
          return nil
      }
      v := int(*p)
      return &v
  }
```
  Apply to all three: OplogSizeMB, SampleSizeBIConnector, SampleRefreshIntervalBIConnector.

josvazg: > ProcessArgsEqual doesn't need to change — it already keys on != nil, so once the converter stops dropping explicit zeros, the
  comparator behaves correctly for them too.

>A regression test mirroring TestProcessArgs_UnsetFieldShouldNotDivergeFromAtlasDefaults but for the explicit-zero case (OplogSizeMB:
  pointer.MakePtr[int64](0) in AKO, 990 in Atlas → must produce a non-empty PATCH and converge after one round) would lock the
  invariant in.

> If you'd rather keep this PR scoped, file it as a follow-up — same root cause, but the issue title (#3142) is specifically about the
  looping behavior the author already fixed, so it's defensible to defer.

internal/translation/deployment/compare.go
```diff
@@ -388,3 +388,45 @@ func areEqual[T comparable](desired, current *T) bool {
...
+// from the PATCH body via MakePtrOrNil — if a field would not be sent over
+// the wire, it must not drive an update either, or the reconciler loops on
+// Atlas-populated server defaults (issue #3142).
+func ProcessArgsEqual(ako, atlas *akov2.ProcessArgs) bool {
+	if ako == nil {
+		return true
+	}
+	if atlas == nil {
+		atlas = &akov2.ProcessArgs{}
+	}
```
josvazg, line 407: **nitpick**:
So that us in the future remember why some fields are not here or why some conversions are the way they are:

```suggestion
	}
	
	// Each check below mirrors how processArgsToAtlas serializes the field:
	//
	//   *bool / *int64 fields  → check `!= nil`  (a non-nil pointer is sent,
  //                                             even if it points to the zero
  //                                             value)
  //   string fields          → check `!= ""`   (empty string is dropped via
  //                                             MakePtrOrNil and not sent)
  //
  // Keep the per-field style aligned with the converter — collapsing the two
  // styles will reintroduce issue #3142.
  //
  // Fields intentionally absent below: they are not part of the SDK PATCH body
  // and therefore cannot drive an update:
  //   - DefaultReadConcern
  //   - FailIndexKeyTooLong (deprecated by Atlas; warned on at the call site)
```

internal/translation/deployment/conversion.go
josvazg, line 965: **nitpick**: remind the coupling with the new comparison function.

```suggestion
  // processArgsToAtlas builds the PATCH body. When adding or removing a field
  // here, mirror the change in deployment.ProcessArgsEqual — the comparator
  // must skip exactly the fields this function omits, or the reconciler loops
  // (issue #3142).
```

josvazg (approved):
LGTM, although we might want to include the fix that could located for those int64 conversions.

helderjs (approved):

internal/translation/deployment/conversion.go
ashortt-mdb: Fixed in 598b6748. Added an `int64PtrToIntPtr` helper next to `processArgsToAtlas` and switched all three fields. Pinned with `TestProcessArgsToAtlas_ExplicitZeroIntIsSent` (and a `NilIntStaysNil` companion to lock the no-opinion case).

internal/translation/deployment/compare.go
ashortt-mdb: Applied verbatim in 27de71fe. Thanks — the absent-fields callout is exactly the kind of thing that drifts.

internal/translation/deployment/conversion.go
ashortt-mdb: Applied (added alongside the helper in 598b6748).

josvazg (approved):
Great quick fix!

igor-karpukhin (approved):

PR #3365, opened 2026-05-21
Potential fix for code scanning alert no. 51: Database query built from user-controlled sources

josvazg:
Potential fix for [https://github.com/mongodb/mongodb-atlas-kubernetes/security/code-scanning/51](https://github.com/mongodb/mongodb-atlas-kubernetes/security/code-scanning/51)

To fix this without changing functionality, keep accepting the same route parameter but introduce a dedicated sanitization/canonicalization function and use its output for the MongoDB filter. This makes the trust boundary explicit and avoids building the query from raw request data.

Best change in `test/app/main.go`:
- Add a helper function (for example `sanitizeKey`) that:
  - validates with the existing `safeKey` regex,
  - returns `(string, error)` so invalid input is rejected early.
- In `getKeyValue`, replace direct use of `mux.Vars(r)["key"]` with:
  1. read raw key,
  2. sanitize/validate via helper,
  3. use sanitized key in `filter`.

No new imports or dependencies are required.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/test/app/main.go
+++ b/test/app/main.go
@@ -70,9 +70,17 @@ func handler(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprintf(w, "It is working")
 }
 
+func sanitizeKey(raw string) (string, error) {
+	if !safeKey.MatchString(raw) {
+		return "", fmt.Errorf("invalid key")
+	}
+	return raw, nil
+}
+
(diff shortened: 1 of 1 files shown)


josvazg:
will not fix, I dismissed the alert as this is test only code

PR #3378, opened 2026-06-03
CLOUDP-410427: Updated AtlasPrivateEndpoint CR to the latest Atlas Go SDK

igor-karpukhin:
# Summary

Updated AtlasPrivateEndpoint CR to the latest Atlas Go SDK. Added two missing fields: `portMappingEnabled` (GCP) and `supportedRegions` (AWS)

### DIFF
--- a/internal/translation/privateendpoint/conversion.go
+++ b/internal/translation/privateendpoint/conversion.go
@@ -79,7 +79,8 @@ func (s *CommonEndpointService) ErrorMessage() string {
 type AWSService struct {
 	CommonEndpointService
 
-	ServiceName string
+	ServiceName      string
+	SupportedRegions []string
 }
 
 func (s *AWSService) Provider() string {
@@ -98,7 +99,8 @@ func (s *AzureService) Provider() string {
 
 type GCPService struct {
 	CommonEndpointService
-	AttachmentNames []string
+	AttachmentNames    []string
+	PortMappingEnabled bool
 }
 
 func (s *GCPService) Provider() string {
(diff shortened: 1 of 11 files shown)


MaciejKaras (approved):

internal/translation/privateendpoint/conversion.go
```diff
@@ -249,6 +253,7 @@ func NewPrivateEndpointStatus(peService EndpointService) status.AtlasPrivateEndp
 		switch pe := peService.(type) {
 		case *AWSService:
 			s.ServiceName = pe.ServiceName
+			s.SupportedRegions = pe.SupportedRegions
```
josvazg, line 256: q: why is portMappingEnabled missing from status?

GCPService.PortMappingEnabled is read from Atlas in serviceFromAtlas but never written to the status. SupportedRegions gets status treatment; PortMappingEnabled doesn't.  why?

internal/translation/privateendpoint/conversion.go
```diff
@@ -378,18 +384,36 @@ func serviceFromAtlas(peService *admin.EndpointService, endpoints EndpointInterf
...
+	return req
+}
+
+func serviceUpdateToAtlas(peService EndpointService) *admin.ApiAtlasModifyEndpointServiceRequest {
+	req := &admin.ApiAtlasModifyEndpointServiceRequest{
+		CloudProvider: peService.Provider(),
+	}
+	if awsSvc, ok := peService.(*AWSService); ok {
+		req.SupportedRemoteRegions = &awsSvc.SupportedRegions
+	}
```
josvazg, line 420: q: why different guards for supported region in create vs update?

In serviceCreateToAtlas (used when first creating the endpoint service in Atlas):

  if awsSvc, ok := peService.(*AWSService); ok && len(awsSvc.SupportedRegions) > 0 {
      req.SupportedRemoteRegions = &awsSvc.SupportedRegions
  }
  If SupportedRegions is nil or empty, the field is not sent — Atlas keeps its default.

  In serviceUpdateToAtlas (used when reconciling an existing endpoint):

  if awsSvc, ok := peService.(*AWSService); ok {
      req.SupportedRemoteRegions = &awsSvc.SupportedRegions  // always set
  }
  If SupportedRegions is nil, this sends SupportedRemoteRegions: null to Atlas. Depending on how Atlas handles that, it may clear all
  previously configured cross-regions.

internal/translation/privateendpoint/conversion.go
igor-karpukhin: Good point, let me add it

internal/translation/privateendpoint/conversion.go
igor-karpukhin: Ok, makes sense. Fixed with explicitly setting then to `[]string{}`

PR #3379, opened 2026-06-03
CLOUDP-400926: AtlasDataFederation CRD updated to the latest atlas go sdk

igor-karpukhin:
# Summary

AtlasDataFederation CRD updated to the latest atlas go sdk. Added new fields: `clusterName` and `cloudProviderConfig`


api/v1/atlasdatafederation_types.go
```diff
@@ -60,6 +67,20 @@ type AWSProviderConfig struct {
...
+// GCPProviderConfig is the configuration for running Data Federation in GCP.
+type GCPProviderConfig struct {
+	// Unique identifier of the role that Data Federation can use to access the data stores.
+	// +kubebuilder:validation:Required
+	RoleID string `json:"roleId"`
+}
+
 type DataProcessRegion struct {
 	// Name of the cloud service that hosts the Federated Database Instance's infrastructure.
 	// +kubebuilder:validation:Enum:=AWS
```
josvazg: q: do we need to update that we now also accept GCP and Azure? or is this still only  AWS?

test/e2e/datafederation_pe_test.go
```diff
@@ -128,6 +115,13 @@ var _ = Describe("DataFederation Private Endpoint", Label("datafederation"), Fla
 					{
 						Name:     "http-test",
 						Provider: "http",
+						ReadPreference: &akov2.ReadPreference{
+							Mode:                "nearest",
+							MaxStalenessSeconds: 90,
+							TagSets: [][]akov2.ReadPreferenceTag{
+								{{Name: "region", Value: "us-east-1"}},
+							},
+						},
```
josvazg: And also lines 153-154
**q**: Why provider "http"?

Atlas only supports these fields on Provider: "atlas" (cluster stores). The e2e test is exercising the new CRD fields, but against a store type that won't use them — so the test doesn't actually verify the Atlas-side behavior.

The unit roundtrip test (TestAtlasClusterStoreRoundtrip) correctly uses atlas as the provider; the e2e should either use atlas too or add a comment explaining why HTTP was chosen.

igor-karpukhin: Fixed this, as it wasn't use these new fields

api/v1/atlasdatafederation_types.go
igor-karpukhin: No, data process is still on AWS

josvazg (approved):

PR #3390, opened 2026-06-09
Avoid the workflow dirs and associated permissions

josvazg:
# Summary

Fix Push to RedHat workflow by pulling from origin instead of upstream. That way the CI automation does not complain it lacks permissions to access workflows, which it does not want to touch anyways.

## Proof of Work

⚠️ Branch test

## Checklist


josvazg:
We actually need to fix the permissions directly, no way around that

PR #3398, opened 2026-06-16
HELP-95382: Let AKO continue to reconcile DBUser scopes if clusters aren't ready

igor-karpukhin:
# Summary

Let AKO continue to reconcile DBUser scopes if clusters aren't ready. When creating a new AtlasDeployment and AtlasDatabaseUser for the deployment the user will fail to create initially since the deployment is not yet ready and we will only retry after 3h which makes the user creation slow and misleading (if the user checks what’s going on after the deployment is up they will just see a message about scoped not being ready)


josvazg:
Can you elaborate a bit more in the description? why was this a problem and why the boolean change fixes it?

ashortt-mdb (approved):

josvazg (approved):

PR #3411, opened 2026-06-25
CLOUDP-416337: Added analyticsAutoScaling for AtlasDeployment CR

igor-karpukhin:
# Summary

Added `analyticsAutoScaling` for AtlasDeployment CR

## Proof of Work

See e2e test

### DIFF
--- a/internal/translation/deployment/compare.go
+++ b/internal/translation/deployment/compare.go
@@ -88,6 +88,10 @@ func ComputeChanges(desired, current *Cluster) (*Cluster, bool) {
 			if autoScalingChanges := getAutoScalingChanges(desiredRegionConfig.AutoScaling, currentRegionConfig); autoScalingChanges != nil {
 				regionConfig.AutoScaling = autoScalingChanges
 			}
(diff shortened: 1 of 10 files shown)


internal/translation/deployment/compare.go
```diff
@@ -155,6 +165,33 @@ func getAutoScalingChanges(desired *akov2.AdvancedAutoScalingSpec, current *akov
 	}
 }
 
+func getAnalyticsAutoScalingChanges(desired *akov2.AdvancedAutoScalingSpec, current *akov2.AdvancedRegionConfig) *akov2.AdvancedAutoScalingSpec {
```
josvazg, line 168: q: This is almost the same as `getAutoScalingChanges`.  Why not collapse both into one function that accepts the "current" autoscaling value as a parameter instead of the whole **AdvancedRegionConfig**?

josvazg (approved):

internal/translation/deployment/compare.go
igor-karpukhin: I'd just leave it as is. We can unify them if needed if follow up PRs

PR #3451, opened 2026-08-03
CLOUDP-430646: Support project tags on AtlasProject

tkaye407:
# Summary

Adds support for [Atlas project tags](https://www.mongodb.com/docs/atlas/tags/) to the legacy `AtlasProject` resource.

Today only the autogenerated `Group` resource (`atlas.generated.mongodb.com`, gated behind `EXPERIMENTAL=1`) supports project tags, and legacy and autogenerated resources cannot be mixed. Internal Atlas Engineering teams that manage projects through AKO and Helm charts need tags for cost attribution and ownership, which blocks them on the legacy resource.

Most of the plumbing already existed — the internal translation `Project` struct already had a `Tags` field and both `fromAtlas` and `toAtlas` already converted it. The gaps filled here are the CRD field, the spec→internal mapping, an update (PATCH) call, and drift detection.

### Changes

| File | Change |
|---|---|
| `api/v1/atlasproject_types.go` | New `spec.tags []*TagSpec`, reusing the existing `TagSpec` type (same key/value length + pattern validation as deployments), `MaxItems=50` |
| `internal/translation/project/conversion.go` | `NewProject` now maps `Spec.Tags`, so tags flow through project creation |
| `internal/translation/project/project.go` | New `UpdateProject` on `ProjectService`, PATCHing `admin.GroupUpdate{Tags: ...}` via `ProjectsAPI.UpdateGroup`. Only tags are updated — the project name is CEL-immutable on the CRD and `withDefaultAlertsSettings` has no effect on existing projects |
| `internal/controller/atlasproject/project.go` | `syncProjectTags` + order-insensitive `tagsInSync`, called inline in `handleProject` |
| `internal/controller/workflow/reason.go` | New `ProjectNotUpdatedInAtlas` condition reason |
| `api/v1/atlasdeployment_types.go` | `TagSpec` doc comment said "on this deployment"; now "on this resource", since projects share the type |
| generated | `zz_generated.deepcopy.go`, `internal/mocks/translation/project_service.go`, `docs/api-docs.md` |

### Semantics

- **`tags` omitted** → the operator does not manage tags at all; tags set outside Kubernetes are left untouched. This keeps existing projects safe on upgrade.
- **`tags: []`** → clears all tags on the project.
- **Value or key drift** → PATCHes the project on the next reconcile.
- **Reordering only** → no API call. Atlas does not guarantee the order tags are returned in, so the comparison is order-insensitive.

Tags reconciliation lives inline in `handleProject` rather than as an `ensureProjectResources` ensurer, because tags are a field of the project document itself rather than a sub-resource. No new condition type or status field is introduced.

### Example

```yaml
apiVersion: atlas.mongodb.com/v1
kind: AtlasProject
metadata:
  name: my-project
spec:
  name: my-project
  tags:
    - key: environment
      value: dev
    - key: team
      value: atlas-eng
```

## Proof of Work

New unit tests, no integration or e2e tests added — the behaviour is fully covered at the translation and controller level.

- `internal/translation/project/project_test.go` — `TestUpdateProject`: update tags, clear tags (asserts the exact `GroupUpdate` payload), and the error path. Existing `TestGetProjectByName` / `TestCreateProject` already assert tags round-trip.
- `internal/translation/project/conversion_test.go` (new) — `TestNewProjectTags`: nil stays nil, empty stays empty, populated maps through.
- `internal/controller/atlasproject/project_tags_test.go` (new) — `TestTagsInSync` (8 cases incl. reordering, renamed key, added/removed tag) and `TestSyncProjectTags` (unmanaged / in sync / reordered / value changed / cleared / update error).
- `internal/controller/atlasproject/project_test.go` — new `TestHandleProject` case asserting the `ProjectNotUpdatedInAtlas` terminate path when the PATCH fails.

```
$ make unit-test
... exit 0, no failures (full suite)

$ make vet && make fmt && make generate && make manifests && make api-docs
... all clean; generated CRD diff is the additive `tags` property only
```

> [!NOTE]
> `make lint` was **not** run locally — `devbox` is not installed on my machine and the host `golangci-lint` refuses the repo's Go 1.26 target (`the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version`). Relying on CI for the lint gate; happy to fix anything it flags.

## Checklist
- [x] Have you checked whether your jira ticket required DOCSP changes? (said yes) 


josvazg:
@tkaye407 thanks a lot for this contribution!
My first review looks good. Basic testing, including all CI tests passed, but being an external contribution from a fork we have some permission issues that prevent further testing.

Unit test coverage seems good, but I wonder if we should also extend some e2e tests to exercise the tag support.

The other issue worth thinking about is key duplication handling; do we allow the latest tag value set win? do we fail in case of key reuse in the same spec?

josvazg:
Cloned PR for testing at #3454

josvazg:
One clarification in the meantime
> (atlas.generated.mongodb.com, gated behind EXPERIMENTAL=1)

The `atlas.generated.mongodb.com` resources are **not** gated by the EXPERIMENTAL, they available.

There are currently 3 groups defined in the operator.
- **atlas.mongodb.com** is the legacy group for manually crafted controller.
- **atlas.generated.mongodb.com** is the new set of generated resources, _present in production by default_.
- **atlas.generated.experimental.mongodb.com** which are gated by the EXPERIMENTAL, and not present in production.

josvazg:
https://github.com/mongodb/mongodb-atlas-kubernetes/pull/3454 (same commit) passed all tests

josvazg (approved):

igor-karpukhin:
I'd update one E2E test with `AtlasProject` to use newly added `tags`. Just as a regression, in case something is changed on the Atlas side.

josvazg:
> I'd update one E2E test with `AtlasProject` to use newly added `tags`. Just as a regression, in case something is changed on the Atlas side.

I would follow up myself with such as change after merging this one.

igor-karpukhin (approved):

PR #3467, opened 2026-08-21
Fixed compare method to support _GEN_2 instances

igor-karpukhin:
# Summary

Fixed compare method to support _GEN_2 instances for `AtlasDeployment` resource

### DIFF
--- a/internal/controller/validate/deployment.go
+++ b/internal/controller/validate/deployment.go
@@ -214,6 +214,16 @@ func advancedInstanceSizeInRange(currentInstanceSize, minInstanceSize, maxInstan
 		return err
 	}
 
+	if !currentSize.Generation.CompatibleWith(minSize.Generation) {
+		return fmt.Errorf("min instance size %s has a hardware generation incompatible with instance size %s",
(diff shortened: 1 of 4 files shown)


josvazg (commented):
q: This is require because we still have left shifted validation of generation names on the operator right?

igor-karpukhin:
> q: This is require because we still have left shifted validation of generation names on the operator right?

Right. But I can't remove it completely right now, because that means fixing all e2e and int tests

josvazg:
> > q: This is require because we still have left shifted validation of generation names on the operator right?
> 
> Right. But I can't remove it completely right now, because that means fixing all e2e and int tests

We should kick all test on this one right?

igor-karpukhin:
> > > q: This is require because we still have left shifted validation of generation names on the operator right?
> > 
> > 
> > Right. But I can't remove it completely right now, because that means fixing all e2e and int tests
> 
> We should kick all test on this one right?

Yes, tests already running

josvazg (approved):

PR #3486, opened 2026-09-02
CLOUDP-433229: Set limit for numShards to max 1000 as in Atlas

igor-karpukhin:
# Summary

Set limit for `numShards` in AtlasDeployment to max `1000` as in Atlas

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/api/v1/atlasdeployment_types.go
+++ b/api/v1/atlasdeployment_types.go
@@ -312,6 +312,10 @@ type AdvancedReplicationSpec struct {
 	// If you set this value to 1 and clusterType is SHARDED, MongoDB Cloud deploys a single-shard sharded cluster.
 	// Don't create a sharded cluster with a single shard for production environments.
 	// Single-shard sharded clusters don't provide the same benefits as multi-shard configurations
+	// The upper bound is the hard limit Atlas enforces on shards per cluster. The limit
+	// effective for a project may be lower (80 by default) and is validated by Atlas.
+	// +kubebuilder:validation:Minimum=0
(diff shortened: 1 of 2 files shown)


api/v1/atlasdeployment_types.go
```diff
@@ -312,6 +312,10 @@ type AdvancedReplicationSpec struct {
 	// If you set this value to 1 and clusterType is SHARDED, MongoDB Cloud deploys a single-shard sharded cluster.
 	// Don't create a sharded cluster with a single shard for production environments.
 	// Single-shard sharded clusters don't provide the same benefits as multi-shard configurations
+	// The upper bound is the hard limit Atlas enforces on shards per cluster. The limit
+	// effective for a project may be lower (80 by default) and is validated by Atlas.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=1000
```
josvazg, line 318: **q**: This is left shift validation, if Atlas changes the values we are on the hook to update again. Do we really want to do this?

igor-karpukhin: Yes, but this has to be limited to some value, otherwise there is a risk of one CR to crash the operator that is installed cluster-wide for example. I found the 1000 is the absolute maximum allowed value in Atlas code. It has to be limited in the CR, otherwise the AKO can be killed with OOM. And it will be just killed without a reason, the cluster admin has to look for why it failed.

josvazg (approved):

PR #3487, opened 2026-09-02
CLOUDP-433215: Use name and namespace as a key of deployment ID

igor-karpukhin:
# Summary

Use name and namespace as a key of deployment ID for the `AtlasBackupSchedule` reference

## Proof of Work

See tests

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/internal/controller/atlasdeployment/backup.go
+++ b/internal/controller/atlasdeployment/backup.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"slices"
 
 	"go.mongodb.org/atlas-sdk/v20250312023/admin"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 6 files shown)


josvazg:
Would want a bit more on the description on why this change is needed.

internal/controller/atlasdeployment/backup.go
```diff
@@ -278,11 +296,13 @@ func (r *AtlasDeploymentReconciler) garbageCollectBackupResource(ctx context.Con
...
 			for _, id := range backupSchedule.Status.DeploymentIDs {
-				if id != clusterName {
+				// slices.Contains, rather than a plain comparison, is what accepts
+				// the legacy bare-name key alongside the current one.
+				if !slices.Contains(deploymentKeys, id) {
 					continue
 				}
 
-				backupSchedule.UpdateStatus([]api.Condition{}, status.AtlasBackupScheduleUnsetDeploymentID(clusterName))
+				backupSchedule.UpdateStatus([]api.Condition{}, status.AtlasBackupScheduleUnsetDeploymentID(deploymentKeys...))
```
josvazg: q: Aren't these Status().Update + Update calls redundant? The inner loop keeps iterating after AtlasBackupScheduleUnsetDeploymentID already removed all matching keys. It is harmless but might be wasteful.

internal/controller/atlasdeployment/backup.go
josvazg, line 292: q: This does a cluster-wide List of all schedules on every deployment deletion. Should we use a field index on spec.backupScheduleRef could scope the query?

This is prior technical debt, not a blocker and could be a follow up.

josvazg (approved):
LGTM with a couple of questions and a possible follow up.

internal/controller/atlasdeployment/backup.go
igor-karpukhin: Good point, let me fix that

internal/controller/atlasdeployment/backup.go
igor-karpukhin: Yes, ideally. I haven't added it as both `AtlasBackupSchedule` and `AtlasBackupPolicy` are legacy resources

PR #3498, opened 2026-09-14
CLOUDP-444754: Support maintenance wave assignment on AtlasProject

tkaye407:
## Summary

Adds support for [Atlas maintenance wave assignment](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-getgroupmaintenancewindow#operation-getgroupmaintenancewindow-200-body-application-vnd-atlas-2023-01-01-json-waveassignment) to the legacy `AtlasProject` resource's `maintenanceWindow`, following the same pattern as #3451 (project tags).

Depends on the atlas-sdk-go bump to `v20250312025` (first version exposing `WaveAssignment` on `GroupMaintenanceWindow`), which has already landed on `main` via #3497 / CLOUDP-444816.

### Changes

| File | Change |
|---|---|
| `api/v1/project/maintenancewindow.go` | New `spec.maintenanceWindow.waveAssignment *int`, `Minimum=1`/`Maximum=3`, optional |
| `internal/translation/maintenancewindow/conversion.go` | `toAtlas`/`fromAtlas` pass `WaveAssignment` through to/from the SDK type |
| `internal/controller/atlasproject/maintenancewindow.go` | `isEmptyWindow`/`windowSpecified` updated to support wave-only windows (no `dayOfWeek` required); new `waveIsDifferent` drift check; `validateMaintenanceWindow` rejects `autoDefer` + `waveAssignment` together |
| `api/v1/project/zz_generated.deepcopy.go` (new), `api/v1/zz_generated.deepcopy.go` | Deep copy support for the new pointer field |
| `docs/api-docs.md` | Regenerated, additive `waveAssignment` entry |

### Semantics

- **`waveAssignment` omitted** → unmanaged. AKO never sends a clearing PATCH, since Atlas has no way to distinguish "never set" from "explicitly cleared" for this field (unlike tags, there's no non-nil-empty sentinel for a scalar).
- **`waveAssignment` set (1-3)** → synced to Atlas; drift-corrected on every reconcile via the existing `Update` path.
- **`waveAssignment` + `autoDefer` together** → rejected client-side. Atlas itself either hard-rejects this combination (on a system-default window) or silently force-clears `autoDefer` (on a user-defined window) — surfacing it as a validation error avoids the operator fighting a server-side override every reconcile.
- Wave-only specs (no `dayOfWeek`) are valid and do not create a "user-defined" window in Atlas, matching Atlas's own wave-only PATCH semantics.
[...]

PR #3499, opened 2026-09-14
CLOUDP-444754: Support maintenance wave assignment on AtlasProject

igor-karpukhin:
Original PR by @tkaye407: https://github.com/mongodb/mongodb-atlas-kubernetes/pull/3498

Adds spec.maintenanceWindow.waveAssignment (1-3, optional) so projects can be assigned to an Atlas maintenance wave for org-wide sequencing, matching the waveAssignment field on GroupMaintenanceWindow.

A nil waveAssignment means unmanaged: AKO never sends a clearing PATCH, since Atlas has no way to distinguish never-set from explicitly-cleared for this field. Wave-only windows (no dayOfWeek) are now valid, and autoDefer + waveAssignment together is rejected client-side to match the Atlas-side conflict.

api/v1/project/zz_generated.deepcopy.go is hand-written to match controller-gen's standard output, since controller-gen's object generator could not run in this sandbox (nested go exec failure); CI's check-missing-files/make generate should verify it byte-for-byte.

# Summary


internal/controller/atlasproject/maintenancewindow.go
```diff
@@ -123,14 +124,27 @@ func daysOrHoursAreDifferent(inAtlas, inAKO maintenancewindow.MaintenanceWindow)
 	return inAtlas.DayOfWeek != inAKO.DayOfWeek || inAtlas.HourOfDay != inAKO.HourOfDay
 }
```
josvazg, line 136: q: Isn't this an infinite reconcile loop?

When a user switches a CR from {dayOfWeek: 2, hourOfDay: 14} to wave-only {waveAssignment: 2} on a project that already has a schedule in Atlas:
windowSpecified returns true (wave is set), so syncAtlasWithSpec runs.
daysOrHoursAreDifferent returns true because AKO's DayOfWeek=0 ≠ Atlas's DayOfWeek=2.
The Update fires, but toAtlas sends DayOfWeek: nil (via MakePtrOrNil(0) + omitempty).
Atlas keeps DayOfWeek=2. Next reconcile: same drift → update again → loop forever.

Root cause: daysOrHoursAreDifferent compares DayOfWeek unconditionally, even when AKO doesn't manage it (value is zero). waveIsDifferent correctly treats nil as "unmanaged" — daysOrHoursAreDifferent lacks the same gate.

Suggested fix at **line 68**:
```go
if (windowSpec.DayOfWeek != 0 && daysOrHoursAreDifferent(*windowInAtlas, *windowInAKO)) ||
    waveIsDifferent(*windowInAtlas, *windowInAKO) {
```

internal/controller/atlasproject/maintenancewindow_test.go
```diff
@@ -427,6 +457,73 @@ func TestEnsureMaintenanceWindow(t *testing.T) {
 			isOK:       false,
 			wantStatus: "False",
 		},
+		{
+			name: "wave-only window is created without a schedule",
+			maintenanceWindow: project.MaintenanceWindow{
+				WaveAssignment: pointer.MakePtr(2),
+			},
+			maintenanceService: func() maintenancewindow.MaintenanceWindowService {
+				service := translation.NewMaintenanceWindowServiceMock(t)
```
josvazg, line 466: q: are we not missing some coverage here?
The test "wave-only window is created without a schedule" only validates initial creation (empty Atlas window). No test covers the scenario above — a wave-only CR reconciling against an Atlas project that already has a DayOfWeek schedule. This gap would have caught the loop bug.

internal/controller/atlasproject/maintenancewindow.go
igor-karpukhin: Yes, this is fixed already

internal/controller/atlasproject/maintenancewindow_test.go
igor-karpukhin: And this one as well, the existing INT test for Maintenance window is updated

josvazg (approved):

